PR #1, opened 2017-08-02
Add operator scaffolding

rossf7:

This PR adds scaffolding for the azure-operator. The aim is to have an operator built to our current standards with stub logic for creating a resource in Azure.

A later PR will have

- creating the stub Azure resource
- local setup guide


README.md
```diff
@@ -1 +1,33 @@
+[![CircleCI](https://circleci.com/gh/giantswarm/azure-operator.svg?style=svg)](https://circleci.com/gh/giantswarm/azure-operator)
```
rossf7: Readme will link to the local setup guide.

glide.yaml
```diff
@@ -0,0 +1,19 @@
+package: github.com/giantswarm/azure-operator
```
rossf7: I think the deps are correct but let me know if you think external deps need pinning.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+		// Internals
+		bootOnce: sync.Once{},
+		tpr:      tpr,
+	}
+
+	return newService, nil
+}
+
+// Boot starts the service and implements the watch for azuretpr.
+func (s *Service) Boot() {
```
rossf7: Creates the TPR on boot.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+			NewObjectFunc:     func() runtime.Object { return &azuretpr.CustomObject{} },
+			NewObjectListFunc: func() runtime.Object { return &azuretpr.List{} },
+		}
+
+		s.tpr.NewInformer(newResourceEventHandler, newZeroObjectFactory).Run(nil)
+	})
+}
+
+func (s *Service) addFunc(obj interface{}) {
+	customObject := *obj.(*azuretpr.CustomObject)
```
rossf7: Placeholders for adding the Azure resource later.

service/service.go
```diff
@@ -0,0 +1,144 @@
...
+		operatorConfig.Logger = config.Logger
+		operatorConfig.Viper = config.Viper
+
+		operatorService, err = operator.New(operatorConfig)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var versionService *versionservice.Service
```
rossf7: Uses the new version service here.

glide.yaml
xh3b4sd: Looks right to me. Only make sure the microendpoint dep uses master.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Settings.
+	if config.Flag == nil {
+		return nil, microerror.Maskf(invalidConfigError, "flag must not be empty")
+	}
+	if config.Viper == nil {
+		return nil, microerror.Maskf(invalidConfigError, "viper must not be empty")
+	}
+
+	tprConfig := tpr.Config{
```
xh3b4sd: Please use `DefaultConfig`.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/giantswarm/azure-operator/flag"
+)
+
+const (
+	TPRVersion     = "v1"
+	TPRDescription = "Managed Kubernetes clusters in Giantnetes running on Azure"
```
xh3b4sd: This is already defined in the TPR repo and should be used from there. See https://github.com/giantswarm/azuretpr/blob/32a7ab664cbb9036f6e1707cbdb7ce1b2d1593b8/meta.go.

service/operator/service.go
xh3b4sd: When type asserting interfaces we should always do proper error handling in case we do not receive the expected type.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+
+		s.tpr.NewInformer(newResourceEventHandler, newZeroObjectFactory).Run(nil)
+	})
+}
+
+func (s *Service) addFunc(obj interface{}) {
+	customObject := *obj.(*azuretpr.CustomObject)
+	s.Config.Logger.Log("debug", fmt.Sprintf("creating cluster '%s'", customObject.Spec.Cluster.Cluster.ID))
+
+	// TODO Add stub code for creating an Azure Resource Group.
```
xh3b4sd: Good enough for this PR. IMO we should use the operator framework in the next steps. See for instance the ingress-operator: https://github.com/giantswarm/ingress-operator/blob/0eda040938a8a9f9b5a9df6e4b6840ae4c6daf23/service/operator/service.go#L151.

service/service.go
```diff
@@ -0,0 +1,144 @@
+package service
+
+import (
+	"fmt"
+	"sync"
+
+	versionservice "github.com/giantswarm/microendpoint/service/version"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	k8sutil "github.com/giantswarm/operatorkit/client/k8s"
```
xh3b4sd: I do not see a need to alias the imports. Also `util` is quite a bad name.

glide.yaml
```diff
@@ -0,0 +1,14 @@
...
+- package: github.com/giantswarm/azuretpr
+  version: master
+- package: github.com/giantswarm/operatorkit
+  version: master
+- package: github.com/giantswarm/microerror
+  version: master
+- package: github.com/giantswarm/micrologger
+  version: master
+- package: github.com/giantswarm/microkit
+  version: master
```
kopiczko: Why is microkit still required?

service/service.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if config.Viper == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.Viper must not be empty")
+	}
+
+	var err error
+
+	var k8sClient kubernetes.Interface
+	{
+		k8sConfig := k8sutil.Config{
```
xh3b4sd: Please use the `DefaultConfig`.

service/operator/service.go
kopiczko: This should be in `azuretpr` `meta.go`.

service/operator/service.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// deleteFunc TODO
+func (s *Service) deleteFunc(obj interface{}) {
+	customObject := *obj.(*azuretpr.CustomObject)
+	s.Config.Logger.Log("debug", fmt.Sprintf("deleting cluster '%s'", customObject.Spec.Cluster.Cluster.ID))
```
kopiczko: `Config` is embedded you can just call `s.Logger.Log`

When this is merged https://github.com/giantswarm/micrologger/commit/3fc80a5b178fd25a757c8bbe54599b21b11a1110 you can now do something like this:

```
logger := s.Logger.With("cluster", customObject.Spec.Cluster.Cluster.ID)

logger.Log("debug", "deleting cluster)

....

logger.Log(...)
```

glide.yaml
xh3b4sd: Because this is an operator and microkit provides the command/daemon/server primitives/structures.

README.md
rossf7: I just meant that the readme is not yet complete. Sorry for the confusion!

service/service.go
rossf7: Ack. Fixed

service/operator/service.go
rossf7: I like that! <3

rossf7:
@xh3b4sd @kopiczko PTAL comments are now addressed. Also microkit now uses micrologger which fixed compile errors in the server. 

Tim, on using the reconciler framework. We need to see what Azure resources we need to create and how we do that. I'm open to the idea and I'll look at ingress-operator as an example.

PR #2, opened 2017-08-04
Local setup guide

rossf7:

This PR adds a local setup guide as well as the Dockerfile and an example TPO. The local setup doesn't cover certs yet because we don't need them till we're ready to boot machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 azure-operator
 TODO
 !vendor/**
+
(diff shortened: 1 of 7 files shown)


rossf7:
@xh3b4sd it's based on the guide @kopiczko did for `aws-operator`. I think it helps a lot for developing locally with `minikube` or another local k8s.

PR #8, opened 2017-08-23
Add Azure API client

rossf7:

This PR adds an Azure API client and config. This finishes some WIP code I had from setting up the operator.

The config is based on the ARM examples for the Azure SDK.
https://github.com/Azure/azure-sdk-for-go/blob/master/arm/examples/check/check.go


client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
+
+// Config contains the common attributes to create an Azure client.
+type Config struct {
+	// Dependencies
+	Logger micrologger.Logger
+
+	// ClientID is the ID of the Active Directory Service Principal.
```
rossf7: This is the config for the Active Directory Service Principal we need to use to authenticate against the Azure APIs.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+		ClientSecret:   "",
+		SubscriptionID: "",
+		TenantID:       "",
+	}
+}
+
+// Clients is the collection of Azure API clients.
+type Clients struct {
+	// DeploymentsClient manages deployments of ARM templates.
+	DeploymentsClient *resources.DeploymentsClient
```
rossf7: So far we just need these 2 API clients. More clients e.g. storage will need to be added later.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+
+	clients := &Clients{
+		DeploymentsClient: deploymentsClient,
+		GroupClient:       groupClient,
+	}
+
+	return clients, nil
+}
+
+func newDeploymentsClient(config *Config) (*resources.DeploymentsClient, error) {
```
rossf7: Here the first API client is created.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	client := resources.NewGroupClient(config.SubscriptionID)
+	client.Authorizer = autorest.NewBearerAuthorizer(spt)
+
+	return &client, nil
+}
+
+func newServicePrincipalToken(config *Config, scope string) (*adal.ServicePrincipalToken, error) {
+	oauthConfig, err := adal.NewOAuthConfig(azure.PublicCloud.ActiveDirectoryEndpoint, config.TenantID)
```
rossf7: Here the auth setup is configured.

glide.yaml
```diff
@@ -12,3 +12,9 @@ import:
   version: master
 - package: github.com/giantswarm/microendpoint
   version: master
+- package: github.com/Azure/azure-sdk-for-go
```
rossf7: For now the Azure libraries are pinned to master.

service/operator/service.go
```diff
@@ -13,14 +13,16 @@ import (
...
+	"github.com/giantswarm/azure-operator/client/azure"
 	"github.com/giantswarm/azure-operator/flag"
 )
 
 // Config represents the configuration used to create an Operator service.
 type Config struct {
 	// Dependencies.
-	Logger    micrologger.Logger
-	K8sClient kubernetes.Interface
+	AzureConfig *azure.Config
```
rossf7: The Azure config is stored rather than the clients to avoid the auth tokens expiring.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
+
+// Config contains the common attributes to create an Azure client.
+type Config struct {
+	// Dependencies
```
kopiczko: smaller nitpick: fullstop missing
bigger nitpick: there should be a empty line between the comment and a field definition. Otherwise it's the field comment in godoc.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
...
+		// Settings.
+		ClientID:       "",
+		ClientSecret:   "",
+		SubscriptionID: "",
+		TenantID:       "",
+	}
+}
+
+// Clients is the collection of Azure API clients.
+type Clients struct {
```
kopiczko: nitpick: That could be called a `ClientSet` like it is in K8s API. `NewClients` looks to me like it should return a slice.

client/azure/azure.go
```diff
@@ -0,0 +1,132 @@
+package azure
```
kopiczko: I'd like to make this flattened. `client.NewAzure` or `azure.NewClient` both look good to me. I don't see a reason to nest that package. We tend to do that way too much IMO.

client/azure/azure.go
rossf7: Good catch, fixed.

service/operator/service.go
```diff
@@ -123,6 +126,11 @@ func (s *Service) addFunc(obj interface{}) {
 		s.Logger.Log("error", "could not convert to azuretpr.CustomObject")
 	}
 
+	_, err := azure.NewClients(s.AzureConfig)
+	if err != nil {
+		s.Logger.Log("error", "could not create azure api clients '%#v'")
+	}
```
kopiczko: Should the client be recreated on every addFunc call?

service/service.go
```diff
@@ -68,6 +70,15 @@ func New(config Config) (*Service, error) {
...
 	var err error
 
+	var azureConfig *azure.Config
+	{
+		azureConfig := azure.DefaultConfig()
+		azureConfig.ClientID = config.Viper.GetString(config.Flag.Service.Azure.ClientID)
+		azureConfig.ClientSecret = config.Viper.GetString(config.Flag.Service.Azure.ClientSecret)
+		azureConfig.SubscriptionID = config.Viper.GetString(config.Flag.Service.Azure.SubscriptionID)
+		azureConfig.TenantID = config.Viper.GetString(config.Flag.Service.Azure.TenantID)
+	}
```
kopiczko: `config.AzureConfig` is actual dependency. It shouldn't be created here.

client/azure/azure.go
xh3b4sd: IMO there is nothing like a client package in an operator. For me these are all service packages that provide business logic to the operator. The services might use clients to connect to external dependencies via some API. I saw this already in the AWS operator and I still do not like it.

client/azure/azure.go
xh3b4sd: FYI in microkit it is called `collection`.

service/operator/service.go
xh3b4sd: IMO no. As mentioned above IMO the operator should use services to provide business logic. This business logic may use some client internally. Such clients should not be exposed outside the service though.

client/azure/azure.go
rossf7: To me having a client package that sets up the API connection makes sense. For k8s we extracted this to operatorkit. I don't think it makes sense to extract the Azure client at this stage.

I agree the business logic should be in services but I think those services can use the Azure client.

client/azure/azure.go
rossf7: I like ClientSet as a name and that it matches the K8s API.

client/azure/azure.go
rossf7: I'm fine with flattening the structure. I think client.NewAzureClientSet is a good name for it.

service/operator/service.go
rossf7: The problem is the auth tokens expire. We can build renewing the token into the client but I think we can do that later.

client/azure/azure.go
xh3b4sd: I see this client stuff is just a factory to create new clients. Still fits into the service package for me. My two cents. Enough said. Do what you think is right.

service/operator/service.go
xh3b4sd: This is what I am saying. This is business logic. "renewing tokens" is a primitive a service implementation should provide. A clients job is to talk to some API. Not more.

client/azure/azure.go
rossf7: This is renamed to AzureClientSet and the package structure is flattened.

service/operator/service.go
rossf7: I agree that is business logic but I don't think we need auto renewing tokens yet. I've created #9 to track this.

service/service.go
rossf7: The dependencies were wrong the AzureConfig is needed for the operator service but not the main service. This is fixed now.

rossf7:
Comment has been added to the addFunc.

Issue #7, opened 2017-08-22
Create Cluster Setup ARM template


rossf7:
Towards giantswarm/sap-hybris#3

## Overview

This ARM template will create Azure resources that will be used by later ARM templates.

## Resources

So far the following resources have been identified.

### Storage Account

A [storage account](https://docs.microsoft.com/en-us/azure/storage/common/storage-create-storage-account) will be created per guest cluster. This will be used for Blob Storage and VM Disks.

### Key Vault

A Key Vault will be created per guest cluster for storing certificate files. On startup nodes will download the secrets from Key Vault via a systemd unit.

rossf7:
# Tasks

- [x] Storage Account
- [ ] Key Vault
- [ ] Add Key Vault secret per certificate file
- [ ] Create Service Principal per guest cluster and grant read access to Key Vault
- [ ] Add systemd unit for downloading cert files from Key Vault

rossf7:
Tested MSI (Managed Service Identities) on a CoreOS node in 'westeurope' region. It adds an Azure agent which has a local endpoint for retrieving OAuth tokens.

Docs for using MSI to access Key Vault on Linux.
https://docs.microsoft.com/en-us/azure/active-directory/msi-tutorial-linux-vm-access-nonaad

This example ARM template shows how to add the extension.
https://github.com/Azure/azure-quickstart-templates/blob/master/201-vmss-msi-linux/azuredeploy.json#L361

rossf7:
Changes to add secrets are done. Issue can be closed once #26 is merged.

PR #10, opened 2017-08-23
Add Cluster Setup ARM template.

rossf7:
Towards #7 

A first version of the Cluster Setup template. For now, it just creates the Storage Account and outputs the name.

The storage account name is very restrictive (3-24 chars only lowercase alphanumeric). So the uniquestring function is used to generate it. The storage account is tagged with the clusterID so we can query for it.

See https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions

### DIFF
--- a/service/arm_templates/cluster_setup.json
+++ b/service/arm_templates/cluster_setup.json
@@ -1,0 +1,49 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "clusterID": {
+      "type": "string",
+      "metadata": {
+        "description": "Unique ID of this guest cluster."
+      }
+    },
+    "storageAccountType": {
+      "type": "string",
+      "defaultValue": "Standard_LRS",
+      "allowedValues": [
+        "Standard_LRS",
+        "Standard_GRS",
+        "Standard_ZRS",
+        "Premium_LRS"
+      ],
+      "metadata": {
+        "description": "Storage Account type."
+      }
+    }
+  },
+  "variables": {
+    "storageAccountName": "[concat(uniquestring(resourceGroup().id), 'clustersa')]"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "name": "[variables('storageAccountName')]",
+      "apiVersion": "2016-01-01",
+      "location": "[resourceGroup().location]",
+      "sku": {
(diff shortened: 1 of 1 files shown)


service/arm_templates/cluster_setup.json
```diff
@@ -0,0 +1,49 @@
...
+        "Standard_ZRS",
+        "Premium_LRS"
+      ],
+      "metadata": {
+        "description": "Storage Account type."
+      }
+    }
+  },
+  "variables": {
+    "storageAccountName": "[concat(uniquestring(resourceGroup().id), 'clustersa')]"
```
rossf7: Here is where the name is generated with the clustersa suffix.

service/arm_templates/cluster_setup.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "name": "[variables('storageAccountName')]",
+      "apiVersion": "2016-01-01",
+      "location": "[resourceGroup().location]",
```
rossf7: Location comes from the resource group which I like.

service/arm_templates/cluster_setup.json
```diff
@@ -0,0 +1,49 @@
...
+      "sku": {
+        "name": "[parameters('storageAccountType')]"
+      },
+      "tags": {
+        "clusterID": "[parameters('clusterID')]"
```
rossf7: Here the resource is tagged with the clusterID.

service/arm_templates/cluster_setup.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "name": "[variables('storageAccountName')]",
+      "apiVersion": "2016-01-01",
+      "location": "[resourceGroup().location]",
+      "sku": {
+        "name": "[parameters('storageAccountType')]"
+      },
```
kopiczko: for my interest: what does sku stands for?

Cocotton: In warehouses, that would mean "stock keeping unit".

service/arm_templates/cluster_setup.json
Cocotton: Do you think it would be better to add a "-" between the resource group id and 'clustersa'?

rossf7: The storage account names don't allow hyphens. Only lowercase and numeric chars are allowed. It's a pain the format is so restrictive.

PR #11, opened 2017-08-25
Add basic network setup ARM template

Cocotton:
Basic template that creates 1 virtuals networks, 2 subnets and 2 security groups (one for each subnet)

Next step:
- Create the routeTables
- Add them to the resources.properties.subnets.properties object

### DIFF
--- a/service/arm_templates/network_setup.json
+++ b/service/arm_templates/network_setup.json
@@ -9,13 +9,6 @@
         "description": "API version used by the Microsoft.Network/virtualNetworks resource."
       }
     },
-    "networkSecurityGroupsAPIVersion": {
-      "type": "string",
-      "defaultValue": "2017-03-01",
-      "metadata": {
-        "description": "API version used by the Microsoft.Network/networkSecurityGroups resource."
-      }
-    },
     "addressPrefix": {
       "type": "string",
       "defaultValue": "10.0.0.0/16",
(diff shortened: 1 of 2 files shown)


service/arm_templates/network_setup.json
```diff
@@ -0,0 +1,82 @@
...
+    "apiVersion": {
+      "type": "string",
+      "defaultValue": "2017-03-01",
+      "metadata": {
+        "description": "API version used by the Microsoft.Network/virtualNetworks resources."
+      }
+    },
+    "addressPrefix": {
+      "type": "string",
+      "defaultValue": "10.0.0.0/16",
```
Cocotton: I set this as a string, but it can be changed to be an array if there's a need to support multiple CIDR blocks.

service/arm_templates/network_setup.json
```diff
@@ -0,0 +1,82 @@
...
+    "subnetWorkerCIDR": {
+      "type": "string",
+      "defaultValue": "10.0.2.0/24",
+      "metadata": {
+        "description": "The CIDR block for the Kubernetes workers' subnet"
+      }
+    }
+  },
+  "variables": {
+    "virtualNetworkName": "[concat(uniquestring(resourceGroup().id), '-clustervn')]",
```
Cocotton: Could there be multiple virtual network per resourceGroup? If so, we'll need to pass the clusterID to this template and use it in the virtualNetworkName instead.

service/arm_templates/network_setup.json
kopiczko: I wonder how we can put comments in the parameters. Resources has `comments` property [1].

1: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates#comments-container

service/arm_templates/network_setup.json
kopiczko: I think it's always good to include `clusterID`.

kopiczko: What is `uniquestring`?

kopiczko: > Could there be multiple virtual network per resourceGroup?

I'm afraid you'll have to check :)

antoineco: @kopiczko https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-string#uniquestring

Cocotton: @kopiczko On Azure level we can create multiple virtual networks per resourceGroup. What I'm note sure is, does the GiantSwarm design requires having multiple virtual networks per resourceGroup?

In any case, I think I'll just find a way to setup a unique virtualNetworkName. That way, that'll be future-proof ^^

Cocotton:
Added some security rules in the lot.

The only traffic allowed on both the masters and workers subnets are between each other and outgoing traffic.

We can update the rules as soon as the load balancers are defined.

service/arm_templates/network_setup.json
rossf7: I looked at the schema in that link but there doesn't seem to be a comments section for parameters. We could include comments in the metadata description field.

service/arm_templates/network_setup.json
```diff
@@ -0,0 +1,221 @@
...
+            "properties": {
+              "description": "Default rule that denies any inbound traffic.",
+              "protocol": "*",
+              "sourcePortRange": "*",
+              "destinationPortRange": "*",
+              "sourceAddressPrefix": "*",
+              "destinationAddressPrefix": "[parameters('subnetMasterCIDR')]",
+              "access": "Deny",
+              "direction": "Inbound",
+              "priority": 1000
```
rossf7: Do all security group rules need to have a priority? Is there any benefit in listing the rules by priority in the template?

Cocotton:
The security groups and rules are now defined and created from their own template.

Now, the first template that should run is the security_groups_setup.json one. Once it's done, the template will output the security groups ID that should be fed to the network_setup.json template.

In the end, I suppose we're going to have a "main" template that will call all of the children templates and manage receiving and passing the outputs.

service/arm_templates/network_setup.json
```diff
@@ -0,0 +1,100 @@
...
+    },
+    "mastersSecurityGroupID": {
+      "type": "string"
+    },
+    "workersSecurityGroupID": {
+      "type": "string"
+    }
+  },
+  "variables": {
+    "virtualNetworkName": "[concat(uniquestring(resourceGroup().id), '-VirtualNetwork')]",
```
rossf7: I think it would be better if we used the cluster ID here rather than `uniquestring`. The cluster IDs are unique within an installation and it looks like Virtual Network names don't have to be globally unique.

https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions

I'd like our resource names to be predictable and driven by the cluster TPO as much as possible. This should mean we need fewer output variables from the templates.

rossf7:
Apart from the naming issue this LGTM. For organising the templates I'd like us to try and use Linked Templates. https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-linked-templates

This was one of the suggestions from Stuart @ Microsoft. We can then have a parent template that is called by the operator but that has multiple child templates. Totally agree this is a better structure than huge monolithic templates.

Cocotton:
The templates are now using the provided clusterID to setup resources names :)

service/arm_templates/network_setup.json
rossf7: FYI the CIDRs will need to be added to `azuretpr`. This is where we set them in `awstpr`. In the operator it will get the values from the cluster TPO when it runs the template.

https://github.com/giantswarm/awstpr/blob/master/spec/aws/vpc.go

rossf7:
This LGTM but I agree with @kopiczko it would be good to see it integrated into the operator. This is blocked on me finishing the deployment resource.

I'd also like to use Linked Templates but there is an issue. The linked templates either have to be public (e.g on GitHub) or in a private blob storage. I think we want to use blob storage for versioning reasons. As I think the operator should manage creating and updating the templates.

Issue #9, opened 2017-08-23
Azure client should become a service and support renewing auth tokens


rossf7:
The current API client is a factory for creating Azure clients. This should become a service and support renewing auth tokens when they expire.

PR #14, opened 2017-08-31
Add Azure Resource Group resource

### LINKED ISSUE #5: Create ResourceGroup resource
* All resources in a guest cluster should be created in an Azure [Resource Group](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview#resource-groups).
* The Resource Group name should include the unique cluster ID we generate for each guest cluster.
* The Resource Group needs to be created as a standalone resource using the operatorkit framework. This is because subsequent resources will need to provide the resource group as a parameter.

See https://godoc.org/github.com/Azure/azure-sdk-for-go/arm/resources/resources#GroupsClient.CreateOrUpdate

rossf7:
Fixes #5 

This PR adds the Resource Group resource to the operator. It creates an Azure resource group based on the cluster TPO. It also adds the updateFunc although resource groups will only be created or deleted.

When a resource group is deleted all the resources it contains are also deleted. We should be able to use this to cleanly delete clusters without orphan resources.


client/azure.go
```diff
@@ -56,8 +58,8 @@ func DefaultAzureConfig() *AzureConfig {
 type AzureClientSet struct {
 	// DeploymentsClient manages deployments of ARM templates.
 	DeploymentsClient *resources.DeploymentsClient
-	// GroupClient manages ARM resource groups.
-	GroupClient *resources.GroupClient
+	// GroupsClient manages ARM resource groups.
+	GroupsClient *resources.GroupsClient
```
rossf7: I was using the wrong client here! The GroupClient is more generic and harder to work with. Whereas the GroupsClient is specifically for managing resource groups.

client/azure.go
```diff
@@ -86,19 +88,29 @@ func NewAzureClientSet(config *AzureConfig) (*AzureClientSet, error) {
...
-		GroupClient:       groupClient,
+		GroupsClient:      groupsClient,
 	}
 
 	return clientset, nil
 }
 
+// ResponseWasNotFound returns true if the response code from the Azure API
+// was a 404.
+func ResponseWasNotFound(resp autorest.Response) bool {
```
rossf7: This helper function checks if the response from the Azure REST API was a 404.

main.go
```diff
@@ -59,13 +62,34 @@ func main() {
...
 
+		var storage microstorage.Storage
+		{
+			storage, err = memory.New(memory.DefaultConfig())
+			if err != nil {
+				panic(err)
+			}
+		}
+
+		var transactionResponder transaction.Responder
```
rossf7: Here a transaction responder is configured which is needed by our microstorage framework.

service/operator/service.go
```diff
@@ -137,6 +137,7 @@ func (s *Service) Boot() {
 		newResourceEventHandler := &cache.ResourceEventHandlerFuncs{
 			AddFunc:    s.addFunc,
 			DeleteFunc: s.deleteFunc,
+			UpdateFunc: s.updateFunc,
```
rossf7: Here the updateFunc support is added.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+		logger: config.Logger.With(
+			"resource", Name,
+		),
+	}
+
+	return newService, nil
+}
+
+// GetCurrentState gets the resource group for this cluster from the Azure API.
+func (r *Resource) GetCurrentState(obj interface{}) (interface{}, error) {
```
rossf7: GetCurrentState calls the Azure API and gets the resource group if it exists.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	return &resourceGroup, nil
+}
+
+// GetDesiredState returns the desired resource group for this cluster.
+func (r *Resource) GetDesiredState(obj interface{}) (interface{}, error) {
```
rossf7: GetDesiredState returns the configured resource group.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "computing the new resource group")
+
+	tags := map[string]string{
+		clusterIDTag:  key.ClusterID(customObject),
+		customerIDTag: key.ClusterCustomer(customObject),
+	}
+	resourceGroup := &azureresource.Group{
+		Name:      to.StringPtr(key.ClusterID(customObject)),
```
rossf7: These `to.StringPtr` helpers are from the Azure/go-autorest library.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "checked if the resource group has to be deleted")
+
+	return resourceGroupToDelete, nil
+}
+
+// GetUpdateState returns nil for the create, delete and update states because
+// resource groups are not updated.
+func (r *Resource) GetUpdateState(obj, currentState, desiredState interface{}) (interface{}, interface{}, interface{}, error) {
```
rossf7: See comment we only create or delete resource groups.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+		groupsClient, err := r.getGroupsClient()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		// Delete the resource group which also deletes all resources it
+		// contains. We wait for the error channel while the deletion happens.
+		_, errchan := groupsClient.Delete(*resourceGroupToDelete.Name, nil)
+		err = <-errchan
+		if err != nil {
```
rossf7: Here we wait for the errchan to make sure there are no errors deleting the resource group. Deleting the group also deletes all resources it contains.

We should be able to use this to cleanly delete clusters without having orphan resources.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+	return nil
+}
+
+// Underlying returns the underlying resource.
+func (r *Resource) Underlying() framework.Resource {
+	return r
+}
+
+func (r *Resource) getGroupsClient() (*azureresource.GroupsClient, error) {
+	azureClients, err := client.NewAzureClientSet(r.azureConfig)
```
rossf7: Currently, we create new API clients per request which is wasteful. #9 tracks fixing this.

service/resource/resourcegroup/resource_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+func Test_Resource_ResourceGroup_GetDesiredState(t *testing.T) {
+	testCases := []struct {
+		Obj              interface{}
+		ExpectedName     string
+		ExpectedLocation string
+		ExpectedTags     map[string]string
+	}{
+		{
+			// Case 1. Standard cluster ID format.
```
rossf7: These tests are based on the Namespace resource in our kvm-operator.

service/service.go
```diff
@@ -95,6 +96,18 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	var resourceGroupResource framework.Resource
```
rossf7: Here we add the new resource to the framework.

service/operator/service.go
```diff
@@ -164,3 +165,12 @@ func (s *Service) deleteFunc(obj interface{}) {
 		s.logger.Log("error", fmt.Sprintf("%#v", err), "event", "delete")
 	}
 }
+
+func (s *Service) updateFunc(oldObj interface{}, newObj interface{}) {
+	s.logger.Log("debug", "executing the operator's updateFunc")
```
xh3b4sd: FYI we lock these functions in other operators with a mutex. See for instance https://github.com/giantswarm/draughtsman-operator/blob/f3e8eaf183176479fed6e5228b2611dec1b5ee9d/service/operator/service.go#L141.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+		groupsClient, err := r.getGroupsClient()
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		resourceGroup, err = groupsClient.Get(key.ClusterID(customObject))
+		if err != nil {
+			if client.ResponseWasNotFound(resourceGroup.Response) {
+				// Fall through.
+				return nil, nil
```
xh3b4sd: `toCustomObject` will actually fail when we return `nil` here. It should at least be an empty, but initiliazed pointer. The alternative would be to handle `nil` in `toCustomObject`. My experience so far is that it is a good thing to differenciate between nil and the empty pointer because of the `interface` interface. We then can be more certain to get the results and types we are expecting.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// GetDesiredState returns the desired resource group for this cluster.
+func (r *Resource) GetDesiredState(obj interface{}) (interface{}, error) {
+	customObject, err := toCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "computing the new resource group")
```
xh3b4sd: FYI I find these additional logs quite spammy and redundant when the log resource is used for wrapping. IN the draughtsman operator I removed this logging already and I would remove it in other operators as well.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,298 @@
...
+		return microerror.Mask(err)
+	}
+
+	resourceGroupToCreate, err := toResourceGroup(createState)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if resourceGroupToCreate != nil {
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "creating the resource group in the Azure API")
```
xh3b4sd: FYI I find this kind of logging useful.

service/resource/resourcegroup/resource.go
xh3b4sd: Can we be sure this will never block forever?

rossf7: Uncertain, there should be a timeout on the API call. I'll double check that closes the error channel.

service/operator/service.go
rossf7: Ack. I wasn't sure if the locking was needed because the state is in the Azure API. But I think it makes sense to lock. I'll fix.

service/resource/resourcegroup/resource.go
rossf7: Good catch, I'll fix.

service/resource/resourcegroup/resource.go
rossf7: Cool, I'll remove here too. Good to standardise across operators.

service/operator/service.go
rossf7: Locking with the mutex is added.

service/resource/resourcegroup/resource.go
rossf7: This is done now. An empty pointer to the Group struct is returned. I've added a new Group struct rather than passing the Azure SDK object around which I think is cleaner.

service/resource/resourcegroup/resource.go
rossf7: Done.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,289 @@
...
+
+		groupsClient, err := r.getGroupsClient()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		// Delete the resource group which also deletes all resources it
+		// contains. We wait for the error channel while the deletion happens.
+		_, errchan := groupsClient.Delete(resourceGroupToDelete.Name, nil)
+		select {
```
rossf7: Here we wait for the error channel or time out after 5 mins. Deleting the resource group is a long running operation if it has a lot of resources.

rossf7:
@kopiczko I've added you as a review. Tim's feedback is now addressed. Can you take a look when you get a chance.

service/operator/service.go
```diff
@@ -157,10 +169,37 @@ func (s *Service) addFunc(obj interface{}) {
...
+
 	s.logger.Log("debug", "executing the operator's deleteFunc")
 
 	err := s.operatorFramework.ProcessDelete(obj, s.resources)
 	if err != nil {
 		s.logger.Log("error", fmt.Sprintf("%#v", err), "event", "delete")
 	}
 }
+
+func (s *Service) updateFunc(oldObj interface{}, newObj interface{}) {
```
kopiczko: I think body of this function should be `addFunc(newObj)`. I strongly disagree we should handle updates separately. It's unreliable, and should be used only as an optimization. I'd even remove `ProcessUpdate` from the framework as I said in another review.

If we call `addFunc(newObj)` here we'll force ourselves to create an idempotent operator from the very begging which is a very good thing.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,289 @@
+package resourcegroup
```
kopiczko: Again me being boring about packages flattening. `service/resource/resourcegroup/resource.go` it doesn't look good when using the code. Should be `service/resource/resource_group.go`. Then `resource.NewResourceGroup(config)`. We tend to make a lot of small packages and copy/paste code.

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,289 @@
...
+		groupClient, err := r.getGroupsClient()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		resourceGroup := azureresource.Group{
+			Name:      to.StringPtr(resourceGroupToCreate.Name),
+			Location:  to.StringPtr(resourceGroupToCreate.Location),
+			ManagedBy: to.StringPtr(managedBy),
+			Tags:      to.StringMapPtr(resourceGroupToCreate.Tags),
```
kopiczko: Why do we pass around custom `Group` object instead of using `azureresource.Group`?

service/resource/resourcegroup/resource.go
```diff
@@ -0,0 +1,289 @@
...
+
+	tags := map[string]string{
+		clusterIDTag:  key.ClusterID(customObject),
+		customerIDTag: key.ClusterCustomer(customObject),
+	}
+	resourceGroup := &Group{
+		Name:     key.ClusterID(customObject),
+		Location: key.Location(customObject),
+		Tags:     tags,
+	}
```
kopiczko: Why we pass group as a pointer if we still check if `Name` is empty?

service/operator/service.go
rossf7: I get the issue with `ProcessUpdate` and I think we should change the updateFunc to call `ProcessCreate`. At the moment updates are not required we just need to be able to create / delete resource groups and deployments.

We need to establish consensus on how we handle updates and if we're not going to use it we should remove `ProcessUpdate` from the framework.

From my POV I think there is benefit from detecting if a deployment has changed and only then launching a new deployment. But maybe we can just always launch a new deployment by calling `CreateOrUpdate` in the Azure SDK.

service/resource/resourcegroup/resource.go
rossf7: I don't think flattening works in this case because each resource needs to implement the Resource interface. Also I think there will only be a small number of resources (Resource Group, Deployment and CloudConfig). For me having them in separate packages makes the code easier to understand.

service/resource/resourcegroup/resource.go
rossf7: We return nil on error or an empty group if there is no resource group. Using the pointer means we can differentiate between these.

service/resource/resourcegroup/resource.go
rossf7: I defined a separate struct because we only care about a subset of the Azure SDK object.

To me this felt cleaner but it does add some overhead as we need to copy the values between the SDK object and our struct.

service/operator/service.go
rossf7: The updateFunc is updated.

rossf7:
@kopiczko the updateFunc is updated and I've replied to your feedback. PTAL.

service/operator/service.go
kopiczko: > From my POV I think there is benefit from detecting if a deployment has changed and only then launching a new deployment.

I think this detection should be the operator responsibility. We have no guarantee that we receive every object update. And in many cases object will arrive to the `addFunc` despite fact it already exists, i.e. after `ResyncPeriod` and when the operator is restarted.

service/resource/resourcegroup/resource.go
kopiczko: We should detect if we have error by checking if err is nil. I don't see problem with returning zero value when error is not nil.

service/operator/service.go
rossf7: Yes, I think that can and should be the operator's responsibility. I think your point on the `addFunc` being called out of sequence is key. Let's continue this as part of the `operatorkit` discussions.

service/resource/resourcegroup/resource.go
rossf7: Cool, I'll get rid of the pointer.

service/operator/service.go
kopiczko: Moreover updating functionality comes from the `cache` (client) package. This is not K8s API feature.

Issue #6, opened 2017-08-22
Create ARM template resource


rossf7:
* We want to use Azure [Resource Manager](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview#template-deployment) (ARM) templates for managing Azure resources.
* We should add an ARM template resource using the operatorkit framework that allows multiple templates to be run.
* To run the templates we need to create a deployment using the Go SDK.

See https://godoc.org/github.com/Azure/azure-sdk-for-go/arm/resources/resources#DeploymentsClient.CreateOrUpdate

kopiczko:
I didn't manage to do anything useful this week. We can't use Azure API w/o more privileges. Details: https://gigantic.slack.com/archives/C5WMGMRL7/p1503507738000658

kopiczko:
This looks like a good sample: https://github.com/Azure-Samples/resource-manager-go-resources-and-groups

rossf7:
The deployment resource is implemented and calls the main template. This uses linked templates to call the child templates. For now, we're using GitHub for hosting the templates so we can iterate on the template dev.

PR #15, opened 2017-09-05
Add deployment resource

rossf7:
Towards #6 

This PR adds the Deployment resource which deploys ARM templates. This PR implements getting the current deployments for the resource group. Creating deployments will come in a later PR.

### DIFF
--- a/service/resource/deployment/deployment.go
+++ b/service/resource/deployment/deployment.go
@@ -1,0 +1,11 @@
+package deployment
+
+const (
+	clusterSetupDeploymentName = "cluster-setup"
+)
+
+func getDeploymentNames() []string {
+	return []string{
+		clusterSetupDeploymentName,
+	}
+}
--- a/service/resource/deployment/deployment.go
+++ b/service/resource/deployment/deployment.go
@@ -1,11 +1,44 @@
 package deployment
 
+import (
+	"fmt"
+
+	"github.com/giantswarm/azuretpr"
+
+	"github.com/giantswarm/azure-operator/service/key"
+)
+
 const (
+	clusterIDParameterName     = "clusterID"
 	clusterSetupDeploymentName = "cluster-setup"
+	clusterSetupTemplate       = "cluster_setup.json"
+	templateBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/master/service/arm_templates"
+	templateVersion            = "1.0.0.0"
 )
 
 func getDeploymentNames() []string {
 	return []string{
 		clusterSetupDeploymentName,
 	}
+}
+
+func newClusterSetupDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
+	var deployment Deployment
+	{
+		deployment = Deployment{
+			Name: clusterSetupDeploymentName,
+			Parameters: map[string]interface{}{
+				clusterIDParameterName: key.ClusterID(cluster),
+			},
+			ResourceGroup:   key.ClusterID(cluster),
+			TemplateURI:     templateURI(clusterSetupTemplate),
+			TemplateVersion: templateVersion,
+		}
+	}
+
+	return deployment, nil
+}
+
+func templateURI(templateFileName string) string {
+	return fmt.Sprintf("%s/%s", templateBaseURI, templateFileName)
 }
--- a/service/resource/deployment/deployment.go
+++ b/service/resource/deployment/deployment.go
@@ -23,17 +23,14 @@ func getDeploymentNames() []string {
 }
 
 func newClusterSetupDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
-	var deployment Deployment
-	{
-		deployment = Deployment{
-			Name: clusterSetupDeploymentName,
-			Parameters: map[string]interface{}{
-				clusterIDParameterName: key.ClusterID(cluster),
-			},
-			ResourceGroup:   key.ClusterID(cluster),
-			TemplateURI:     templateURI(clusterSetupTemplate),
-			TemplateVersion: templateVersion,
-		}
+	deployment := Deployment{
+		Name: clusterSetupDeploymentName,
+		Parameters: map[string]interface{}{
+			clusterIDParameterName: key.ClusterID(cluster),
+		},
+		ResourceGroup:   key.ClusterID(cluster),
+		TemplateURI:     templateURI(clusterSetupTemplate),
+		TemplateVersion: templateVersion,
 	}
 
 	return deployment, nil
--- a/service/resource/deployment/resource.go
+++ b/service/resource/deployment/resource.go
@@ -95,10 +95,11 @@ func (r *Resource) GetCurrentState(obj interface{}) (interface{}, error) {
 			}
 
 			deployment := Deployment{
-				Name:          *deploymentExtended.Name,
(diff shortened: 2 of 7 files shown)


service/resource/deployment/deployment.go
```diff
@@ -0,0 +1,11 @@
+package deployment
+
+const (
+	clusterSetupDeploymentName = "cluster-setup"
+)
+
```
rossf7: Here we return the list of deployments created by the operator. For now only cluster setup is added.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	resourceGroupName := key.ClusterID(customObject)
+	deploymentClient, err := r.getDeploymentsClient()
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	var deployments []Deployment
+	{
```
rossf7: Here we get which deployments exist for the resource group.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,199 @@
...
+	return []Deployment{}, nil
+}
+
+// GetCreateState is not yet implemented.
+func (r *Resource) GetCreateState(obj, currentState, desiredState interface{}) (interface{}, error) {
+	return []Deployment{}, nil
+}
+
+// GetDeleteState returns an empty deployments collection. Deployments and the
+// resources they manage are deleted when the Resource Group is deleted.
```
rossf7: As per this comment deleting deployments will not be implemented. Instead, we'll delete the resource group to ensure all resources are deleted.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (r *Resource) getDeploymentsClient() (*azureresource.DeploymentsClient, error) {
+	azureClients, err := client.NewAzureClientSet(r.azureConfig)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return azureClients.DeploymentsClient, nil
+}
+
```
rossf7: These helper functions will be used in the create deployment PR.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,199 @@
...
+			deployments = append(deployments, deployment)
+		}
+	}
+
+	return deployments, nil
+}
+
+// GetDesiredState is not yet implemented.
+func (r *Resource) GetDesiredState(obj interface{}) (interface{}, error) {
+	return []Deployment{}, nil
```
kopiczko: Is it WIP?

rossf7: I'm going to do GetDesiredState in the follow-up PR with the create logic. I thought it would be good to split the changes into smaller PRs.

kopiczko: If it's easier for you I can review this, but I think it's good to have it as a whole in a single PR. Does it make a difference for you?

rossf7: Cool :) I'll do it as a single PR

service/resource/deployment/deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+
+	"github.com/giantswarm/azuretpr"
+
+	"github.com/giantswarm/azure-operator/service/key"
+)
+
+const (
+	clusterIDParameterName     = "clusterID"
+	clusterSetupDeploymentName = "cluster-setup"
```
rossf7: For now we use the cluster-setup template. This will be replaced by the master template.

service/resource/deployment/deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	clusterSetupTemplate       = "cluster_setup.json"
+	templateBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/master/service/arm_templates"
```
rossf7: For now we use the template from master on GitHub. We can iterate on this and decide if we use the commit SHA or blob storage.

service/resource/deployment/deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	templateVersion            = "1.0.0.0"
+)
+
+func getDeploymentNames() []string {
+	return []string{
+		clusterSetupDeploymentName,
+	}
+}
+
+func newClusterSetupDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
```
rossf7: Here the deployment settings are created.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	deployments := []Deployment{
+		clusterSetup,
+	}
+	return deployments, nil
+}
+
+// GetCreateState returns the deployments that should be created for this
+// cluster.
+func (r *Resource) GetCreateState(obj, currentState, desiredState interface{}) (interface{}, error) {
```
rossf7: Here we check if deployments already exist for this resource group.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,294 @@
...
+			deploymentsToCreate = append(deploymentsToCreate, desiredDeployment)
+		}
+	}
+
+	return deploymentsToCreate, nil
+}
+
+// GetDeleteState returns an empty deployments collection. Deployments and the
+// resources they manage are deleted when the Resource Group is deleted.
+func (r *Resource) GetDeleteState(obj, currentState, desiredState interface{}) (interface{}, error) {
```
rossf7: See the comment. We delete the resource group to ensure all resources are deleted.

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,294 @@
...
+	return []Deployment{}, []Deployment{}, []Deployment{}, nil
+}
+
+// Name returns the resource name.
+func (r *Resource) Name() string {
+	return Name
+}
+
+// ProcessCreateState creates the deployments via the Azure API.
+func (r *Resource) ProcessCreateState(obj, createState interface{}) error {
```
rossf7: Here is where the deployments are created.

rossf7:
@Cocotton The PR is updated so it now creates the deployments.

@fgimenez Adding you as a reviewer since Pawel is AFK.

service/resource/deployment/deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	var deployment Deployment
+	{
```
fgimenez: Do we get any benefit in this case, given the function length, by defining an inner block?

service/resource/deployment/resource.go
```diff
@@ -0,0 +1,294 @@
...
+		if d.Name == name {
+			return d, nil
+		}
+	}
+
+	return Deployment{}, microerror.Maskf(notFoundError, name)
+}
+
+func toCustomObject(v interface{}) (azuretpr.CustomObject, error) {
+	customObjectPointer, ok := v.(*azuretpr.CustomObject)
```
fgimenez: Shouldn't we check for `v` being nil here?

service/resource/deployment/deployment.go
rossf7: Good point, removed!

service/resource/deployment/resource.go
rossf7: Also done.

PR #16, opened 2017-09-07
Main arm template

Cocotton:
It is now possible to call the template main.json to deploy the cluster_setup.json, security_groups_setup.json and virtual_network_setup.json templates.

If you want to test it with the azure-cli, you'll only have to provide the clusterID, masterLoadBalancerCidr and workerLoadBalancerCidr parameters since they don't have defaults. If you have an idea of what a good default address could be for the 2 loadbalancers, let me know.

I also tried matching the templates' parameters to the azuretpr [structures elements](https://github.com/giantswarm/azuretpr/tree/39b23dfc8bebc3a706b3b8dd286a3612e0fe35b2/spec/azure). 

I think an important part to review is the [parameters block](https://github.com/giantswarm/azure-operator/pull/16/files#diff-bdfeb20a499ae8d7abc02b96a7ec37a5R4) in the main template to define good parameters names that will be used in the templates as well as in the tpr.

### DIFF
--- a/service/arm_templates/main.json
+++ b/service/arm_templates/main.json
@@ -73,6 +73,18 @@
     "kubernetesKubeletPort": {
       "type": "string",
       "defaultValue": "10250"
+    },
+    "clusterSetupTemplate": {
+      "type": "string",
+      "defaultValue": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/cluster_setup.json"
+    },
+    "securityGroupsSetupTemplate": {
+      "type": "string",
+      "defaultValue": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/security_groups_setup.json"
+    },
+    "virtualNetworkSetupTemplate": {
+      "type": "string",
+      "defaultValue": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/virtual_network_setup.json"
     }
   },
   "resources": [
@@ -83,7 +95,7 @@
       "properties": {
         "mode": "incremental",
         "templateLink": {
-          "uri": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/cluster_setup.json",
+          "uri": "[parameters('clusterSetupTemplate')]",
           "contentVersion": "1.0.0.0"
         },
         "parameters": {
@@ -99,7 +111,7 @@
       "properties": {
         "mode": "incremental",
         "templateLink": {
-          "uri": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/security_groups_setup.json",
+          "uri": "[parameters('securityGroupsSetupTemplate')]",
           "contentVersion": "1.0.0.0"
         },
         "parameters": {
@@ -123,7 +135,7 @@
       "properties": {
         "mode": "incremental",
         "templateLink": {
-          "uri": "https://raw.githubusercontent.com/giantswarm/azure-operator/main-arm-template/service/arm_templates/virtual_network_setup.json",
+          "uri": "[parameters('virtualNetworkSetupTemplate')]",
           "contentVersion": "1.0.0.0"
         },
         "parameters": {
--- a/service/arm_templates/main.json
+++ b/service/arm_templates/main.json
@@ -42,10 +42,16 @@
         "description": "The masters subnet's CIDR."
       }
     },
-    "masterLoadBalancerCidr": {
+    "apiLoadBalancerCidr": {
       "type": "string",
       "metadata": {
-        "description": "The masters load balancer's CIDR."
+        "description": "The Kubernetes API load balancer's CIDR."
+      }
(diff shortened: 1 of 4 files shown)


service/arm_templates/main.json
```diff
@@ -0,0 +1,152 @@
...
+      }
+    },
+    "workerSubnetCidr": {
+      "type": "string",
+      "defaultValue": "10.0.2.0/24",
+      "metadata": {
+        "description": "The masters subnet's CIDR."
+      }
+    },
+    "masterLoadBalancerCidr": {
```
rossf7: For AWS we have separate ELBs for `etcd` and `apiserver`. As etcd is internal but apiserver is also used by users. Since the Azure approach is to have separate subnets I think we should split this.

Long term we want to remove the etcd loadbalancer. It's only needed for Calico and they are working on using the k8s API for storage.

rossf7:
The master template LGTM and its good to see how they integrate. I have a worry about having a single master template due to the number of parameters involved.

The operator should be driven by the values in the cluster TPO. So that means for things like the apiserver port even though there is a default we should provide the value from TPO. For now I'm OK with having a single master template to get us going. But we may need a way of generating the parameters in Go from the TPO.

On the load balancer CIDRs I think they also need to be driven by the TPO. For the vnet peering I think we'll need to assign unique values per guest cluster. This is what we do for AWS and the IP ranges are assigned by our `kubernetesd`API.

service/arm_templates/main.json
```diff
@@ -0,0 +1,152 @@
...
+      "type": "string",
+      "metadata": {
+        "description": "The masters load balancer's CIDR."
+      }
+    },
+    "workerLoadBalancerCidr": {
```
rossf7: I think this should be `ingressLoadBalancerCidr` as that is what the LB is exposing. This matches what we do in `aws-operator`.

service/arm_templates/security_groups_setup.json
```diff
@@ -193,68 +180,68 @@
...
+              "sourceAddressPrefix": "[parameters('workerLoadBalancerCidr')]",
+              "destinationAddressPrefix": "[parameters('workerSubnetCidr')]",
               "access": "Allow",
               "direction": "Inbound",
               "priority": "3800"
             }
           },
           {
-            "name": "workersLBToKubelet",
+            "name": "workerLBToKubelet",
```
rossf7: I don't think this rule is needed. The kubelet port is only used by k8s and shouldn't be exposed by the LB. The master already has access because traffic is allowed between the master and worker nodes.

service/arm_templates/main.json
Cocotton: What I could do is split the masterLoadBalancerCidr into 2 load balancers, masterAPILoadBalancerCidr(public ip) and masterEtcdLoadBalancerCidr (private ip). Does that sound good to you?

service/arm_templates/main.json
```diff
@@ -0,0 +1,125 @@
...
+        "description": "The main CIDR block reserved for this virtual network."
+      }
+    },
+    "masterSubnetCidr": {
+      "type": "string"
+    },
+    "workerSubnetCidr": {
+      "type": "string"
+    },
+    "apiLoadBalancerCidr": {
```
rossf7: I like the new names but I have a question on the CIDR's. Do we need to provide these or will they be outputs when the load balancer resources are created?

If we need to provide them then they need to be added to `azuretpr`.

service/arm_templates/main.json
```diff
@@ -0,0 +1,125 @@
...
+    "etcdPort": {
+      "type": "string"
+    },
+    "kubernetesIngressSecurePort": {
+      "type": "string"
+    },
+    "kubernetesIngressInsecurePort": {
+      "type": "string"
+    },
+    "clusterSetupTemplate": {
```
rossf7: Could we have a single base URI parameter for all the templates? If you can use the `concat` function to build the template URIs that will reduce the number of params.

service/arm_templates/main.json
```diff
@@ -0,0 +1,125 @@
...
+      "type": "string"
+    },
+    "etcdLoadBalancerCidr": {
+      "type": "string"
+    },
+    "ingressLoadBalancerCidr": {
+      "type": "string"
+    },
+    "kubernetesAPISecurePort": {
+      "type": "string"
```
rossf7: I've got another query. In `clustertpr` we store the ports as ints. So we need to convert to a string in the operator. WDYT about making these int parameters?

service/arm_templates/main.json
Cocotton: Both can work, but I think ultimately they will be provided by the azuretpr.

service/arm_templates/main.json
Cocotton: Yes, I'll update the PR in a couple of minute to reflect that.

service/arm_templates/main.json
rossf7: Cool having them in the TPR is fine.

Cocotton:
@rossf7 I updated the main template as well as the security_groups_setup one to receive the port parameters as integers rather than string.

PR #17, opened 2017-09-08
Launch main template

rossf7:
Towards #6 

This PR updates the operator to deploy the main ARM template that was merged in #16. It also adds a template branch flag to override the GitHub branch when creating template links.

For now we're using a single main template that takes in parameters from the cluster TPO. We may need to split this later if there are too many parameters, not sure yet.


examples/cluster.yaml
```diff
@@ -35,10 +35,10 @@ spec:
       domain: "cluster.local"
       hyperkube:
         docker:
-          image: "quay.io/giantswarm/hyperkube:v1.6.4_coreos.0"
+          image: "quay.io/giantswarm/hyperkube:v1.7.5_coreos.0"
```
rossf7: Some boyscouting to update values in the cluster TPO.

examples/cluster.yaml
```diff
@@ -57,3 +57,13 @@ spec:
...
+    storage:
+      accountType: "Standard_LRS"
+    virtualNetwork:
+      cidr: "10.0.0.0/16"
+      masterSubnetCidr: "10.0.1.0/24"
+      workerSubnetCidr: "10.0.2.0/24"      
+      loadBalancer:
+        apiCidr: "10.0.3.0/25"
+        etcdCidr: "10.0.3.128/25"
+        ingressCidr: "10.0.4.0/25"
```
rossf7: New config that has been added to `azuretpr`.

flag/service/azure/template/template.go
```diff
@@ -0,0 +1,5 @@
+package template
+
+type Template struct {
+	Branch string
+}
```
rossf7: Adds flag to override the GitHub branch used for deploying templates. Defaults to master if not set.

service/resource/deployment/deployment.go
```diff
@@ -4,38 +4,75 @@ import (
...
 
 func getDeploymentNames() []string {
 	return []string{
-		clusterSetupDeploymentName,
+		mainDeploymentName,
 	}
 }
 
-func newClusterSetupDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
+func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
```
rossf7: Here the main template deployment is created. The child templates are called from this template via template links.

service/resource/deployment/resource.go
```diff
@@ -34,13 +39,17 @@ type Config struct {
 // best effort.
 func DefaultConfig() Config {
 	return Config{
+		TemplateBranch: masterBranch,
```
rossf7: The DefaultConfig defaults the template branch to master.

examples/local/cluster.tmpl.yaml
```diff
@@ -35,10 +35,10 @@ spec:
       domain: "cluster.local"
       hyperkube:
         docker:
-          image: "quay.io/giantswarm/hyperkube:v1.6.4_coreos.0"
+          image: "quay.io/giantswarm/hyperkube:v1.7.5_coreos.0
```
xh3b4sd: The closing double quote is missing.

service/resource/deployment/deployment.go
```diff
@@ -4,38 +4,75 @@ import (
...
 )
 
 const (
-	clusterIDParameterName     = "clusterID"
-	clusterSetupDeploymentName = "cluster-setup"
-	clusterSetupTemplate       = "cluster_setup.json"
-	templateBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/master/service/arm_templates"
-	templateVersion            = "1.0.0.0"
+	mainDeploymentName          = "cluster-main-template"
+	templatesBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/%s/service/arm_templates/"
```
xh3b4sd: This looks like we are fetching the template at runtime from the Github API. I noticed recently that especially the raw content API is not very reliable. I am against this. We should rather have a templating library like k8scloudconfig. I seem to not understand how this here works. I checked the deployment resource and noted that there is logic in `ProcessCreateState` which actually belongs to `GetCreateState`. The process methods should only take state and act on it via some client implementation. All state creation and manipulation does not belong there. See https://github.com/giantswarm/azure-operator/blob/746539df00e3d47f5842f6f2f48d57dba44978be/service/resource/deployment/resource.go#L178.

rossf7: For hosting linked templates the standard approaches are GitHub or private blob storage (Azure equivalent of S3). I see the reliability problem with using GitHub. We decided to go with it for now so we can iterate on the template deployment.

We could add a template resource to the operator that uploads the ARM templates to blob storage. That's what we'll be doing for the cloudconfig.

Ack, I'll remove the state manipulation from `ProcessCreateState`.

examples/local/cluster.tmpl.yaml
rossf7: Fixed

flag/service/azure/template/template.go
kopiczko: Maybe we should call it `Version` or something more generic, to hide that detail. My thinking here is if we decide to switch from Github to Azure storage or anything else, we probably don't want to change flag names.

service/resource/deployment/deployment.go
xh3b4sd: I see. I missed how Azure works. So this just is what it is.

rossf7:
@xh3b4sd Logic for transforming the param structure is moved from ProcessCreate to where the deployment resource is created.

@kopiczko Good point on the flag name. I've renamed to URI Version as the templates also have a version.

PTAL

service/resource/deployment/resource.go
```diff
@@ -21,9 +21,14 @@ const (
...
 	createTimeout  = 5 * time.Minute
 	deploymentMode = "Incremental"
+	masterBranch   = "master"
 )
 
 type Config struct {
+	// URIVersion is used when creating template links for ARM templates.
+	// Defaults to master for deploying templates hosted on GitHub.
+	URIVersion string
+
```
xh3b4sd: We usually structure configurations like this. 

* dependencies
* internals
* settings

Lets please keep this order and comment structure. Right now it is super easy to scan config, service, initialization and validation structures because we are almost consistent with this. See some random microservice: https://github.com/giantswarm/kubernetesd/blob/bea1a75406b6442b59ff04de51009b57462462d2/service/creator/service.go#L33.

kopiczko: I think configuration shouldn't have `internals`. They should emerge in `New` based on the configuration IMO.

service/resource/deployment/resource.go
```diff
@@ -188,18 +198,10 @@ func (r *Resource) ProcessCreateState(obj, createState interface{}) error {
...
 
-			params := make(map[string]interface{}, len(deploy.Parameters))
-			for key, val := range deploy.Parameters {
-				params[key] = struct {
-					Value interface{}
-				}{
-					Value: val,
-				}
-			}
 			deployment := azureresource.Deployment{
```
xh3b4sd: IMO the deployment is the resource state. So all this should just be injected in here and the process method transforms the interface to the proper type and executes client logic against it. I also assumed we fix this separately.

service/resource/deployment/resource.go
xh3b4sd: I thought this is common sense. Thanks for proper clarification.

rossf7: Ack I missed that. The ordering is updated.

service/resource/deployment/resource.go
rossf7: The problem is the Azure API returns a DeploymentExtended object from the Get API request but takes in a Deployment object. So I went with implementing an internal structure in https://github.com/giantswarm/azure-operator/blob/master/service/resource/deployment/spec.go

https://godoc.org/github.com/Azure/azure-sdk-for-go/arm/resources/resources#DeploymentsClient.Get
https://godoc.org/github.com/Azure/azure-sdk-for-go/arm/resources/resources#DeploymentsClient.CreateOrUpdate

I can switch to using the Deployment object as the resource but if so I'd like to do later in a separate issue.

xh3b4sd: IMO the state/structure used to act against some client is the one a resource should dispatch and manage. When this is your custom Deployment defined in the spec this is ok. Then this should also be the current, desired, create and delete state. Fine with fixing this separately.

rossf7: Ack but I think this is done. All the methods pass a []Deployment so we can deploy multiple templates later if needed. The logic that processed the parameters is now called from `GetDesiredState` and not in `ProcessCreateState`.

The delete state is not implemented because deleting the resource group deletes all resources including the deployments.

PR #20, opened 2017-09-18
Vendor updates

rossf7:
Updates vendor and to use latest operatorkit changes.


service/operator/service.go
```diff
@@ -126,83 +129,46 @@ func New(config Config) (*Service, error) {
...
-		err := s.tpr.CreateAndWait()
-		if tpr.IsAlreadyExists(err) {
-			s.Logger.Log("debug", "third party resource already exists")
-		} else if err != nil {
-			s.Logger.Log("error", fmt.Sprintf("%#v", err))
-			return
-		}
-
-		s.Logger.Log("debug", "starting list/watch")
+		o := func() error {
```
rossf7: I added the retry support for booting the operator as I think it's useful in general and not just for KVM.

service/operator/service.go
```diff
@@ -126,83 +129,46 @@ func New(config Config) (*Service, error) {
...
-
-	err := s.operatorFramework.ProcessCreate(obj, s.resources)
-	if err != nil {
-		s.logger.Log("error", fmt.Sprintf("%#v", err), "event", "create")
+func (s *Service) bootWithError() error {
+	err := s.tpr.CreateAndWait()
+	if tpr.IsAlreadyExists(err) {
+		s.Logger.Log("debug", "third party resource already exists")
+	} else if err != nil {
+		s.Logger.Log("error", fmt.Sprintf("%#v", err))
```
xh3b4sd: When we handle the error here we do not need to log it. We would now log it twice. Lets get rid of this log though.

rossf7: Good point, done.

PR #21, opened 2017-09-19
Control plane

Cocotton:
This PR add the master node(s) setup template. It first creates a network interfaces for every required master node and then creates the virtual machines attached with their own NIC.

The templates allows to create 1 or more master nodes. Simply set the _masterCount_ and _nasterIPAddresses_ accordingly.

## Questions/Notes

- The template creates a vhds container in the storage account to store the os/data disks attached to the VMs.
- We could set a list of "allowedValues" to the _masterVMSize_ parameter depending on the ones you think could be used. I did not want to list them all since there's a lot.
- The _masterAdminSSHKeyPath_ parameter could be built automatically using the _masterAdminUsername_ parameter, if you think this could save some time (e.g. /home/_masterAdminUsername_/.ssh/authorized_keys
- We'll have to update the new parameters to match those that will be setup in the tpr.

## Test

I tested the template using the azure-cli. If you want to test it too, first create a _main.parameters.json_ file with the following content

```
{
  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "clusterID": {
      "value": "TESTID"
    },
    "storageAccountType": {
      "value": "Standard_LRS"
    },
    "virtualNetworkCidr": {
      "value": "10.0.0.0/16"
    },
    "masterSubnetCidr": {
      "value": "10.0.1.0/24"
    },
    "workerSubnetCidr": {
      "value": "10.0.2.0/24"
    },
    "apiLoadBalancerCidr": {
      "value": "10.0.1.1/32"
    },
    "etcdLoadBalancerCidr": {
      "value": "10.0.1.2/32"
    },
    "ingressLoadBalancerCidr": {
      "value": "10.0.2.1/32"
    },
    "kubernetesAPISecurePort": {
      "value": 443
    },
    "etcdPort": {
      "value": 2379
    },
    "kubernetesIngressSecurePort": {
      "value": 30011
    },
      "value": 30010
    },
    "templatesBaseURI": {
      "value": "https://raw.githubusercontent.com/giantswarm/azure-operator/control_plane/service/arm_templates/"
    },
    "masterCount": {
      "value": 1
    },

    "masterVMSize": {
      "value": "Standard_A1"
    },
    "masterAdminUsername": {
      "value": "core"
    },
    "masterAdminSSHKeyData": {
      "value": "ssh-rsa YOUR_PUBLIC_KEY"
    },
    "masterOSImagePublisher": {
      "value": "CoreOS"
    },
    "masterOSImageOffer": {
      "value": "CoreOS"
    },
    "masterOSImageSKU": {
      "value": "Stable"
    },
    "masterOSImageVersion": {
      "value": "1465.7.0"
    },
    "masterDataDiskSizeGB": {
      "value": 50
    },
    "masterCloudConfigData": {
      "value": "SOME_BASE64_ENCODED_CLOUD-CONFIG_DATA"
    }
  }
}

```
Make sure to update the _masterAdminSSHKeyData_ and _masterCloudConfigData_ parameters.

Then, run `az group deployment create -g YOUR_RESOURCE_GROUP --template-file main.json --parameters @main.parameters.json`


service/arm_templates/main.json
```diff
@@ -50,28 +50,93 @@
     "kubernetesIngressInsecurePort": {
       "type": "int"
     },
+    "masterCount": {
+      "type": "int",
+      "defaultValue": 1
+    },
+    "masterIPAddresses": {
```
rossf7: Do we have to assign the IP address? Or can Azure assign one from the masters CIDR?

In our `kubernetesd` API we have IPAM logic for assigning the subnet network ranges. But having to assign IP addresses for each node would be more difficult.

service/arm_templates/main.json
```diff
@@ -50,28 +50,93 @@
...
+      "type": "array",
+      "metadata": {
+        "description": "An array containing all the master nodes' ip addresses."
+      }
+    },
+    "masterVMSize": {
```
rossf7: We should provide the VM size via the `azuretpr` custom object. I don't think we need to have the possible values in the template.

For AWS we have configuration in our APIs for the allowed EC2 instance types per installation. We can do the same for Azure.

service/arm_templates/main.json
```diff
@@ -50,28 +50,93 @@
...
+      "type": "string"
+    },
+    "masterDataDiskSizeGB": {
+      "type": "int"
+    },
+    "masterAdminUsername": {
+      "type": "string"
+    },
+    "masterAdminPassword": {
```
rossf7: I don't think password auth should be enabled. Can we remove the parameter and force the use of SSH keys?

service/arm_templates/main.json
```diff
@@ -50,28 +50,93 @@
...
+      "type": "string"
+    },
+    "masterDisablePasswordAuth": {
+      "type": "bool",
+      "defaultValue": true
+    },
+    "masterAdminSSHKeyPath": {
+      "type": "string"
+    },
+    "masterAdminSSHKeyData": {
```
rossf7: Can we just use `masterAdminSSHKeyData` and remove `masterAdminSSHKeyPath`?

While we're developing I think it's good provide a master SSH key via the operator. It's also possible to create individual user accounts with SSH keys via `k8scloudconfig`. But that can come later.

https://github.com/giantswarm/clustertpr/blob/master/spec/kubernetes/ssh/user.go

service/arm_templates/main.json
```diff
@@ -50,28 +50,93 @@
...
+      "type": "string"
+    },
+    "masterOSImagePublisher": {
+      "type": "string"
+    },
+    "masterOSImageOffer": {
```
rossf7: What is the difference between Publisher and Offer? We'll need to add the values for the CoreOS image to `azuretpr`.

service/arm_templates/main.json
Cocotton: We can dynamically assign an IP address, so I'll make it so. Was not sure if you wanted to have full control on that.

rossf7: Thanks dynamically assigning is great. In this case less control is better as otherwise the IPAM logic gets too complex in kubernetesd.

service/arm_templates/main.json
Cocotton: I tried removing it from the template, but it's not working and the cli throws an error telling the value of _linuxConfiguration.ssh.publicKeys.path_ is invalid.

I'll build it in the variable (/home/_masterAdminUsername_/.ssh/authorized_keys, that way the tpr won't need to provide it.

service/arm_templates/main.json
Cocotton: If GiantSwarm built their own CoreOS image, we could have Publisher:GiantSwarm Offer:CoreOS

service/arm_templates/main.json
```diff
@@ -50,28 +50,81 @@
...
+      "defaultValue": "master_nodes_setup.json"
+    },
     "securityGroupsSetupTemplateFile": {
       "type": "string",
       "defaultValue": "security_groups_setup.json"
     },
+    "virtualMachinesSetupTemplateFile": {
+      "type": "string",
+      "defaultValue": "virtual_machines_setup.json"
+    },
```
kopiczko: I don't see this file. Is it ninja fix?

service/arm_templates/master_nodes_setup.json
```diff
@@ -0,0 +1,169 @@
...
+                "id": "[parameters('masterSubnetID')]"
+              }
+            }
+          }
+        ]
+      }
+    },
+    {
+      "apiVersion": "[parameters('virtualMachinesAPIVersion')]",
+      "type": "Microsoft.Compute/virtualMachines",
```
kopiczko: Shouldn't we use some kind of autoscaling here?

service/arm_templates/main.json
kopiczko: This is how it's defined on Azure:

```
 +          "imageReference": {
 +            "publisher": "[parameters('masterOSImagePublisher')]",
 +            "offer": "[parameters('masterOSImageOffer')]",
 +            "sku": "[parameters('masterOSImageSKU')]",
 +            "version": "[parameters('masterOSImageVersion')]"
 +          },
```

service/arm_templates/main.json
kopiczko: I don't see this file. Is it ninja fix?

service/arm_templates/master_nodes_setup.json
Cocotton: I had a talk with @rossf7 a couple of days ago and the conclusion was that for now, we would only run with 1 master.

Although, this PR allows the create of more than one master is desired (masterCount and masterIPAddresses parameters).

I propose we attack the autoscaling into a separate PR/issue. What do you say?

service/arm_templates/main.json
Cocotton: Good catch, forgot about that parameter that is unused. Just removed it from the file :)

service/arm_templates/master_nodes_setup.json
kopiczko: Cool

Cocotton:
@kopiczko @rossf7, I renamed the master_nodes_setup.json template to nodes_setup.json and also updated it to make it agnostic from the concept of masters/workers. That way, we'll only need one template to create both the masters and workers. I also updated to main template to reflect those changes.

The change does not require anything new from the azuretpr :)

If everything still feel good with you, I'll go ahead and update the [deployment structure parameters](https://github.com/giantswarm/azure-operator/blob/master/service/resource/deployment/deployment.go#L60) as well as the tests.

service/arm_templates/main.json
```diff
@@ -131,9 +180,54 @@
...
       }
     },
+    {
+      "apiVersion": "2016-09-01",
+      "name": "availability_sets_setup",
+      "type": "Microsoft.Resources/deployments",
+      "properties": {
+        "mode": "incremental",
+        "templateLink": {
+          "uri": "[variables('availabilitySetsTemplateURI')]",
```
kopiczko: I think we can make availabilitySetsTemplate generic. Like we have in nodesSetupTemplate. We would create one AS for masters and one AS work workers here. Does that makes sense?

service/resource/deployment/deployment.go
```diff
@@ -66,13 +66,23 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment,
...
 		"etcdLoadBalancerCidr":          cluster.Spec.Azure.VirtualNetwork.LoadBalancer.EtcdCIDR,
 		"ingressLoadBalancerCidr":       cluster.Spec.Azure.VirtualNetwork.LoadBalancer.IngressCIDR,
+		"masterVMSize":                  cluster.Spec.Azure.Masters[0].VMSize,
+		"masterDataDiskSizeGB":          cluster.Spec.Azure.Masters[0].DataDiskSizeGB,
+		"masterAdminUsername":           cluster.Spec.Azure.Masters[0].AdminUsername,
+		"masterAdminSSHKeyData":         cluster.Spec.Azure.Masters[0].AdminSSHKeyData,
+		"masterOSImagePublisher":        cluster.Spec.Azure.Masters[0].OSImage.Publisher,
+		"masterOSImageOffer":            cluster.Spec.Azure.Masters[0].OSImage.Offer,
+		"masterOSImageSKU":              cluster.Spec.Azure.Masters[0].OSImage.SKU,
+		"masterOSImageVersion":          cluster.Spec.Azure.Masters[0].OSImage.Version,
```
kopiczko: Are we going to support only a single master? I think we have to find a way to handle arrays in ARM templates.

Cocotton: In the end, I think it would make more sens to update the [Masters and Workers fields](https://github.com/giantswarm/azuretpr/blob/master/spec/azure.go#L13) in the azuretpr to be of type azure.Node instead of []azure.Node.

In the end, I would assume that every masters will have the same specs, as well as the workers.

We would still be able to create multiple masters/workers by passing the [nodeCount parameter](https://github.com/giantswarm/azure-operator/pull/21/files#diff-bdfeb20a499ae8d7abc02b96a7ec37a5R262)

kopiczko: For master nodes this is true. Not necessary for worker nodes. We have each node in the cluster configured separately in other operators, and it will make it easier to keep it that way.

@xh3b4sd you may want to provide more input here.

xh3b4sd: +1 it should be possible to have workers with different capabilities running in your cluster. You may want to have a cluster with 6 nodes for workloads where 3 have a lot of computing power and 3 have a lot of memory or disk space.

Cocotton: Makes sense to me, I'll start looking at how we could provide an arrow of masters/workers as parameters

Cocotton:
@kopiczko @xh3b4sd as per our discussion earlier, I updated the nodes_setup template as well as the main template to receive an array of azure.Node object from the azuretpr.

Everything seems to be working, except for the cloud-config, which I fetch from [here](https://github.com/giantswarm/azure-operator/pull/21/files#diff-c1e703348c0627cd9d47637a92477b83R49). It appears the result is not Base64 encoded.

I'll have a look into that later today(getting late here) unless anyone has an idea of what's going on.

FYI, I also had to vendor update the azuretpr for all the tests to pass.

service/arm_templates/nodes_setup.json
```diff
@@ -0,0 +1,141 @@
+{
```
kopiczko: I think this file should be called `node_setup.json` asa it creates  a single node.

service/arm_templates/main.json
```diff
@@ -163,5 +188,63 @@
...
+          "contentVersion": "1.0.0.0"
+        },
+        "parameters": {
+          "clusterID": {"value": "[parameters('clusterID')]"},
+          "storageAccountName": {"value": "[reference('cluster_setup').outputs.storageAccountName.value]"},
+          "prefix": {"value": "[concat(variables('masterPrefix'), '-', copyIndex(1))]"},
+          "subnetID": {"value": "[reference('virtual_network_setup').outputs.masterSubnetID.value]"},
+          "availabilitySetID": {"value": "[reference('master_availability_sets_setup').outputs.availabilitySetID.value]"},
+          "disablePasswordAuth": {"value": "[parameters('masterDisablePasswordAuth')]"},
+          "nodeCustomConfig": {"value": "[parameters('mastersCustomConfig')[copyIndex()]]"},
```
kopiczko: I think it would be great to pass all the parameters from `nodeCustomConfig` explicitly here, and take them as separate variables in `nodes_setup.json`. 

Benefits:

* We can describe separate variables in `nodes_setup.json`, and add comments.
* `nodeCustomConfig` properties are accessed in a single place - then it's easy to discover what parameters it ha.
*  The object parameter stays in scope of a single file/template - again, it's easier to discover the object definition.
* `nodes_setup.json` becomes easier to understand, better documented on its own.

service/resource/deployment/deployment.go
```diff
@@ -46,7 +46,7 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment,
 	certSecrets := convertCertsToSecrets(certs)
 
 	// TODO Master CloudConfig will be passed in as a template parameter.
```
kopiczko: This TODO is done :)

service/resource/deployment/deployment.go
```diff
@@ -66,10 +66,12 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment,
 		"apiLoadBalancerCidr":           cluster.Spec.Azure.VirtualNetwork.LoadBalancer.APICIDR,
 		"etcdLoadBalancerCidr":          cluster.Spec.Azure.VirtualNetwork.LoadBalancer.EtcdCIDR,
 		"ingressLoadBalancerCidr":       cluster.Spec.Azure.VirtualNetwork.LoadBalancer.IngressCIDR,
+		"mastersCustomConfig":           cluster.Spec.Azure.Masters,
```
kopiczko: As you can see `params` is passed to `convertParams` below, which isn't quite ready yet for receiving nested object types. I'll try to update it today.

For the reference current implementation:

```go
// convertParameters converts the map into the structure used by the Azure API.
func convertParameters(inputs map[string]interface{}) map[string]interface{} {
	params := make(map[string]interface{}, len(inputs))
	for key, val := range inputs {
		params[key] = struct {
			Value interface{}
		}{
			Value: val,
		}
	}

	return params
}
```

xh3b4sd:
@Cocotton the cloud config template is a base64 encoded and gzip compressed string. See

- https://github.com/giantswarm/azure-operator/blob/ac5b64ab79bf519bac5580484e393fe5f4205dbf/service/cloudconfig/service.go#L107
- https://github.com/giantswarm/k8scloudconfig/blob/05b5ac2a4ed0d145efb1ce8002f26b228ae3cc1b/cloudconfig.go#L51

service/arm_templates/main.json
kopiczko: Note to self: this is done. Commend didn't disappear.

service/arm_templates/main.json
```diff
@@ -163,5 +188,70 @@
...
+          "availabilitySetID": {"value": "[reference('master_availability_sets_setup').outputs.availabilitySetID.value]"},
+          "disablePasswordAuth": {"value": "[parameters('masterDisablePasswordAuth')]"},
+          "vmSize": {"value": "[parameters('mastersCustomConfig')[copyIndex()].vmSize]"},
+          "adminUsername": {"value": "[parameters('mastersCustomConfig')[copyIndex()].adminUsername]"},
+          "adminSSHKeyData": {"value": "[parameters('mastersCustomConfig')[copyIndex()].adminSSHKeyData]"},
+          "dataDiskSizeGB": {"value": "[parameters('mastersCustomConfig')[copyIndex()].dataDiskSizeGB]"},
+          "osImagePublisher": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.publisher]"},
+          "osImageOffer": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.offer]"},
+          "osImageSKU": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.sku]"},
+          "osImageVersion": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.version]"},
```
kopiczko: Cool. I like the object doesn't leak outside.

service/resource/deployment/deployment.go
kopiczko: I created issue for that: https://github.com/giantswarm/azure-operator/issues/31

Cocotton:
@xh3b4sd it looks like Azure does not accept encrypted cloud-config files :( 

I tested this by base64decoding, uncompressing and re-base64encoding the cloud-config file, fetched [here](https://github.com/giantswarm/azure-operator/pull/21/files#diff-c1e703348c0627cd9d47637a92477b83R32), right before passing the `masterCloudConfig` string to the `params` map. The deployment was successful with the unencrypted cloud-config.

I think you guys will agree that it's not the best idea to pass the full base64 encoded file directly to the arm template, since we dangerously approaching the file size limit (65K on Azure if I recall correctly).

I guess we could maybe do some inline compression/encoding like seen [here](https://coreos.com/os/docs/latest/cloud-config.html#write_files)?

Or maybe have a smaller cloud-config calling a bigger one?

What do you guys think?

Cocotton:
@kopiczko I managed to deploy a cluster of 2 masters (using the azure-operator) without any issues from the `convertParameters` function.

I guess this is good news, but I'm not sure to understand why it's not failing. I'll dig a bit more on that :)

xh3b4sd:
@Cocotton we could just not compress and encode the cloud config. Calling `String` instead of `Base64` would be good enough. See

- https://github.com/giantswarm/azure-operator/blob/ac5b64ab79bf519bac5580484e393fe5f4205dbf/service/cloudconfig/service.go#L107
- https://github.com/giantswarm/k8scloudconfig/blob/05b5ac2a4ed0d145efb1ce8002f26b228ae3cc1b/cloudconfig.go#L62

When we have the size limitation anyway we probably need to do the trick with the small CC calls the bigger one. We are already doing the same thing in AWS. See 

- https://github.com/giantswarm/aws-operator/blob/855f90b848b4b7b1355770878c184262a300eb41/service/create/templates.go#L136

Cocotton:
@xh3b4sd see my last commit for the modifications.

Is this the right way to do it? I was thinking the correct way would be to open a PR in the k8scloudconfig repo to add a new function returning the base64 encoded string. Would you rather have me do that?

xh3b4sd:
@Cocotton IMO this is good enough for now when it works like that. We have plans to use ignition at some point and want to use a dedicate CloudConfig/Ignition operator for such things in the future anyway. So there will be movement based on the past experiences anyway. So lets keep it like this if it does the trick for azure.

Cocotton:
Sounds good to me! :)

examples/cluster.yaml
```diff
@@ -59,12 +59,22 @@ spec:
...
+        offer: "CoreOS"
+        sku: "Stable"
+        version: "1465.7.0"
     storage:
       accountType: "Standard_LRS"
     virtualNetwork:
       cidr: "10.0.0.0/16"
       masterSubnetCidr: "10.0.1.0/24"
-      workerSubnetCidr: "10.0.2.0/24"      
+      workerSubnetCidr: "10.0.2.0/24"
```
kopiczko: I tried to create a cluster according to the local guide and failed with:

```
{"caller":"github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/framework/framework.go:102","error":"[{/Users/pawel/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/framework/framework.go:222: }
 {/Users/pawel/go/src/github.com/giantswarm/azure-operator/service/resource/deployment/resource.go:242: } {resources.DeploymentsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code=\"InvalidTemplate\" Message=\"Deployment template validation failed: 'The template parameters 'masterCloudConfigData, mastersCustomConfig' in the parameters file are not valid; they are not present in the original template and can therefore not be provided at deployment time. The only supported parameters for this template are 'clusterID, storageAccountType, virtualNetworkCidr, masterSubnetCidr, workerSubnetCidr, apiLoadBalancerCidr, etcdLoadBalancerCidr, ingressLoadBalancerCidr, kubernetesAPISecurePort, etcdPort, kubernetesIngressSecurePort, kubernetesIngressInsecurePort, keyVaultName, keyVaultSecretsObject, templatesBaseURI, certificatesTemplateFile, clusterSetupTemplateFile, securityGroupsSetupTemplateFile, virtualNetworkSetupTemplateFile'. Please see https://aka.ms/arm-deploy/#parameter-file for usage details.'.\"}]","event":"create","time":"17-10-05 14:28:30.563"}
```

service/cloudconfig/service.go
```diff
@@ -104,5 +106,5 @@ func newCloudConfig(template string, params k8scloudconfig.Params) (string, erro
 		return "", microerror.Mask(err)
 	}
 
-	return cloudConfig.Base64(), nil
+	return base64.StdEncoding.EncodeToString([]byte(cloudConfig.String())), nil
```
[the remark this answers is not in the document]
Cocotton: The .Base64() method returns a gzip compressed cloud-config, but Azure does not seem to allow that.

I has to be an uncompressed base64 encoded cloud-config :(

examples/cluster.yaml
Cocotton: Make sure to `export AZURE_TEMPLATE_URI_VERSION="control_plane"`

service/cloudconfig/service.go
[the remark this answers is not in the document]
rossf7: It's a shame gzip compression isn't supported as I know we're close to the 64kb limit. Unless we can reduce it we may need to download the cloudconfig from blob storage but for now lets go with this.

service/resource/deployment/deployment.go
rossf7: Issue was closed as `convertParams` is working with nested objects.

service/cloudconfig/service.go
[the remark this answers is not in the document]
Cocotton: Something we could do if we want to avoid downloading it from blob storage would be to compress the file contents, like here: https://coreos.com/os/docs/latest/cloud-config.html#write_files.

If we decide to do so, this PR would still be working properly :)

kopiczko:
Nice work @Cocotton. Thanks!

PR #22, opened 2017-09-19
Add CloudConfig service

rossf7:
This PR adds the cloudconfig service which is used by the deployment resource. The cloudconfig will be provided as parameters to the main ARM template in a later PR.

For now the cloudconfig is generated without operator specific extensions. A custom unit will be needed to download the certs from Key Vault. This will be done in a later change. This PR is just to add the service and template rendering.


client/azure.go
```diff
@@ -104,7 +104,7 @@ func NewAzureClientSet(config *AzureConfig) (*AzureClientSet, error) {
 // ResponseWasNotFound returns true if the response code from the Azure API
 // was a 404.
 func ResponseWasNotFound(resp autorest.Response) bool {
-	if resp.StatusCode == http.StatusNotFound {
+	if resp.Response != nil && resp.StatusCode == http.StatusNotFound {
```
rossf7: This is a simple bug fix I've sneaked in. The http.Response returned in autorest.Response can be nil depending on the error. So we need to check for it.

service/cloudconfig/cloud_config.go
```diff
@@ -0,0 +1,33 @@
+package cloudconfig
+
+import (
+	k8scloudconfig "github.com/giantswarm/k8scloudconfig"
+)
+
+type MasterExtension struct{}
+
+func (me *MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
```
rossf7: Files is empty because the certificates will be stored in Key Vault.

service/cloudconfig/cloud_config.go
```diff
@@ -0,0 +1,33 @@
...
+	return nil, nil
+}
+
+func (me *MasterExtension) Units() ([]k8scloudconfig.UnitAsset, error) {
```
rossf7: A custom unit will be needed to download the certs from Key Vault. This will come later.

service/cloudconfig/cloud_config.go
```diff
@@ -0,0 +1,33 @@
...
+	return nil, nil
+}
+
+func (me *MasterExtension) VerbatimSections() []k8scloudconfig.VerbatimSection {
```
rossf7: We may need verbatim sections for the cloud provider support. Not certain yet so will also come later.

service/cloudconfig/service.go
```diff
@@ -0,0 +1,98 @@
...
+func (s Service) NewWorkerCloudConfig(customObject azuretpr.CustomObject) (string, error) {
+	params := k8scloudconfig.Params{
+		Cluster:   customObject.Spec.Cluster,
+		Extension: &WorkerExtension{},
+	}
+
+	return newCloudConfig(k8scloudconfig.WorkerTemplate, params)
+}
+
+func newCloudConfig(template string, params k8scloudconfig.Params) (string, error) {
```
rossf7: Here is where the cloudconfig is generated.

service/resource/deployment/deployment.go
```diff
@@ -34,6 +34,19 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment,
 	if err != nil {
 		return Deployment{}, microerror.Mask(err)
 	}
+
+	// TODO Master CloudConfig will be passed in as a template parameter.
```
rossf7: See comment. The parameters will be integrated later via the main ARM template.

service/service.go
```diff
@@ -116,6 +130,7 @@ func New(config Config) (*Service, error) {
 		deploymentConfig := deployment.DefaultConfig()
 		deploymentConfig.URIVersion = config.Viper.GetString(config.Flag.Service.Azure.Template.URI.Version)
 		deploymentConfig.AzureConfig = azureConfig
+		deploymentConfig.CloudConfig = cloudConfigService
```
rossf7: Here the service is passed to the deployment resource.

rossf7:
Yes I do like the cloudconfig operator approach. The code here is very similar to the code in KVM and AWS operators. 

The CC operator would need to be able to watch for all 3 provider TPRs. Although only one is active per installation. It would mean we had a single component with all the extensions and reduce some of the dupe code.

service/cloudconfig/cloud_config.go
kopiczko: Those are very good comment. I wonder why you don't make them godoc.

service/cloudconfig/service.go
```diff
@@ -0,0 +1,98 @@
...
+
+		// Settings.
+		Flag:  nil,
+		Viper: nil,
+	}
+}
+
+// Service implements the cloudconfig service interface.
+type Service struct {
+	Config
```
kopiczko: I dislike embedding `Config` into the `Service`. It desn't feel right that code outside can mutate the Service.

I'm also on a mission of encouraging people to not call structs `Service`, this makes it difficult to track stacked calls. I'd suggest `CloudConfig` name for this struct.

xh3b4sd: +1

xh3b4sd:
> The CC operator would need to be able to watch for all 3 provider TPRs.

I would rather have a CC TPO that is issued which the CC operator acts upon. The operator can then have different resource implementations for each cloud provider and does its magic. The resources would then just be configured via flags. #implementationdetail #later

service/cloudconfig/cloud_config.go
rossf7: Comments added :)

service/cloudconfig/service.go
rossf7: Done

rossf7:
I've added an issue for the CloudConfig operator to the sig-operator inbox. So we can decide if we want to proceed / spec it.

https://github.com/giantswarm/giantswarm/issues/1866

PR #23, opened 2017-09-20
Add certificate secrets to key vault

rossf7:
Towards giantswarm/azure-operator#7

This PR adds a new linked template that creates an Azure Key Vault and creates secrets for the cluster certs issued by `cert-operator`. A later PR will use MSI (Managed Services Identities) to grant read access to the vault to cluster nodes.

https://azure.microsoft.com/en-us/blog/keep-credentials-out-of-code-introducing-azure-ad-managed-service-identity/

Note that Key Vault also supports storing certificates but this would involve extra file handling. By using secrets we can use the `az keyvault secret download` CLI command to download the certs to `/etc/kubernetes/ssl`. Hashicorp Vault is our certificate issuer and we just use Azure Key Vault for secure object storage.


examples/cluster.yaml
```diff
@@ -56,6 +56,8 @@ spec:
     - hostname: "worker-1"
 
   azure:
+    keyVault:
+      vaultName: "example-cluster-vault"
```
rossf7: Here is the new `azuretpr` value.

glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/giantswarm/azure-operator
 import:
 - package: github.com/giantswarm/azuretpr
-  version: master
+  version: keyvault
```
rossf7: TODO

service/arm_templates/certificates.json
```diff
@@ -0,0 +1,57 @@
+{
```
rossf7: This is a new child template for creating the vault and certs files as secrets.

service/arm_templates/certificates.json
```diff
@@ -0,0 +1,57 @@
...
+        "tags": {
+          "clusterID": "[parameters('clusterID')]"
+        },
+        "properties": {
+          "sku": {
+            "family": "[parameters('keyVaultSKUFamily')]",
+            "name": "[parameters('keyVaultSKUName')]"
+          },
+          "tenantId": "[subscription().tenantId]",
+          "accessPolicies": []
```
rossf7: The access policy will allow the cluster VMs to download the secrets files using a MSI (Managed Services Identify). This can't be added until the master nodes template has been merged.

service/arm_templates/main.json
```diff
@@ -50,12 +50,22 @@
     "kubernetesIngressInsecurePort": {
       "type": "int"
     },
+    "keyVaultName": {
+      "type": "string"
+    },
+    "keyVaultSecretsObject": {
+      "type": "secureObject"
```
rossf7: We need to use `secureObject` here so the secrets data is obfuscated.

service/key/key.go
```diff
@@ -1,17 +1,34 @@
 package key
 
 import (
+	"fmt"
+
 	"github.com/giantswarm/azuretpr"
+	"github.com/giantswarm/certificatetpr"
 )
 
+// ClusterCustomer returns the customer ID for this cluster.
```
rossf7: Boyscouting to add the missing GoDoc.

service/key/key.go
```diff
@@ -1,17 +1,34 @@
...
+	return customObject.Spec.Azure.KeyVault.VaultName
+}
+
+// Location returns the physical location where the Resource Group is deployed.
 func Location(customObject azuretpr.CustomObject) string {
 	return customObject.Spec.Azure.Location
 }
+
+// SecretName returns the name of the Key Vault secret for this certificate
+// asset.
```
rossf7: A unique secret name is generated within the vault for the certs files. They can't be the file name unfortunately.

> An object-name is a user provided name for and must be unique within a Key Vault. The name must be a string 1-127 characters in length containing only 0-9, a-z, A-Z, and -.

service/resource/deployment/deployment.go
```diff
@@ -19,6 +21,17 @@ const (
 	virtualNetworkSetupTemplate = "virtual_network_setup.json"
 )
 
+// keyVaultSecrets is used to pass secrets to Key Vault as a secure object.
```
rossf7: Here are new structs for passing the secrets data. It gets serialised to JSON when the deployment is created.

I like this and I think we can use it more to reduce the number of parameters we need to provide to the main template.

service/resource/deployment/deployment.go
```diff
@@ -35,6 +48,14 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment,
 		return Deployment{}, microerror.Mask(err)
 	}
 
+	certs, err := r.certWatcher.SearchCerts(key.ClusterID(cluster))
```
rossf7: Here we search for the certificate secrets generated by `cert-operator`.

service/resource/deployment/deployment.go
xh3b4sd: FYI we used a separate `spec.go` file for such types in the past for better separation.

xh3b4sd: I see this file is not for the resource implementation. Then this may just be good enough.

rossf7: I moved these to `spec.go` as yes I think that's tidier.

examples/cluster.yaml
kopiczko: We decided to it's `Name` in the tpr project, so probably should be `name` here.

examples/local/cluster.tmpl.yaml
```diff
@@ -56,6 +56,8 @@ spec:
     - hostname: "worker-1"
 
   azure:
+    keyVault:
+      vaultName: "$(CLUSTER_NAME)-vault"
```
kopiczko: Same.

service/arm_templates/certificates.json
kopiczko: Do we need to track that or you'll remember?

examples/local/cluster.tmpl.yaml
rossf7: Thanks missed that. Both fixed.

rossf7:
@kopiczko I added a TODO comment to the vault resource about setting the access policy. Shame we can't have regular comments since the template is JSON but the comment block is useful.

PR #29, opened 2017-09-27
fix template URI mess

kopiczko:
* extract URI generating functions and test them
* remove unused constants

### DIFF
--- a/service/resource/deployment/deployment.go
+++ b/service/resource/deployment/deployment.go
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	mainDeploymentName = "cluster-main-template"
-	templateVersion    = "1.0.0.0"
+	mainDeploymentName     = "cluster-main-template"
+	templateContentVersion = "1.0.0.0"
 )
 
 func getDeploymentNames() []string {
@@ -56,15 +56,15 @@ func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment, 
 		"kubernetesIngressInsecurePort": cluster.Spec.Cluster.Kubernetes.IngressController.InsecurePort,
 		"keyVaultName":                  key.KeyVaultName(cluster),
 		"keyVaultSecretsObject":         certSecrets,
-		"templatesBaseURI":              baseTemplateURI(r.templateURIVersion),
+		"templatesBaseURI":              baseTemplateURI(r.templateVersion),
 	}
 
 	deployment := Deployment{
-		Name:            mainDeploymentName,
-		Parameters:      convertParameters(params),
-		ResourceGroup:   key.ClusterID(cluster),
-		TemplateURI:     templateURI(r.templateURIVersion, mainTemplate),
-		TemplateVersion: templateVersion,
+		Name:                   mainDeploymentName,
+		Parameters:             convertParameters(params),
+		ResourceGroup:          key.ClusterID(cluster),
+		TemplateURI:            templateURI(r.templateVersion, mainTemplate),
+		TemplateContentVersion: templateContentVersion,
 	}
 
 	return deployment, nil
--- a/service/resource/deployment/deployment.go
+++ b/service/resource/deployment/deployment.go
@@ -1,8 +1,6 @@
 package deployment
 
 import (
-	"fmt"
-
 	"github.com/giantswarm/certificatetpr"
 
 	"github.com/giantswarm/azuretpr"
@@ -12,13 +10,8 @@ import (
 )
 
 const (
-	mainDeploymentName          = "cluster-main-template"
-	templatesBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/%s/service/arm_templates/"
-	templateVersion             = "1.0.0.0"
-	mainTemplate                = "main.json"
-	clusterSetupTemplate        = "cluster_setup.json"
-	securityGroupsSetupTemplate = "security_groups_setup.json"
-	virtualNetworkSetupTemplate = "virtual_network_setup.json"
+	mainDeploymentName = "cluster-main-template"
+	templateVersion    = "1.0.0.0"
 )
 
 func getDeploymentNames() []string {
@@ -28,15 +21,6 @@ func getDeploymentNames() []string {
 }
 
 func (r Resource) newMainDeployment(cluster azuretpr.CustomObject) (Deployment, error) {
-	templateBaseURI, err := r.templateBaseURI()
-	if err != nil {
-		return Deployment{}, microerror.Mask(err)
-	}
-	mainTemplateURI, err := r.templateURI(mainTemplate)
-	if err != nil {
-		return Deployment{}, microerror.Mask(err)
-	}
-
 	certs, err := r.certWatcher.SearchCerts(key.ClusterID(cluster))
 	if err != nil {
 		return Deployment{}, microerror.Mask(err)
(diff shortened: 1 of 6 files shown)


service/resource/deployment/deployment.go
```diff
@@ -12,13 +10,8 @@ import (
...
 const (
-	mainDeploymentName          = "cluster-main-template"
-	templatesBaseURI            = "https://raw.githubusercontent.com/giantswarm/azure-operator/%s/service/arm_templates/"
-	templateVersion             = "1.0.0.0"
-	mainTemplate                = "main.json"
-	clusterSetupTemplate        = "cluster_setup.json"
-	securityGroupsSetupTemplate = "security_groups_setup.json"
-	virtualNetworkSetupTemplate = "virtual_network_setup.json"
+	mainDeploymentName = "cluster-main-template"
+	templateVersion    = "1.0.0.0"
```
xh3b4sd: Just for my interest, what is this for a version scheme?

service/resource/deployment/deployment.go
```diff
@@ -12,13 +10,8 @@ import (
...
 )
 
 func getDeploymentNames() []string {
```
xh3b4sd: FYI I moved this to the key package in the KVM operator.

service/resource/deployment/resource.go
```diff
@@ -85,6 +84,9 @@ func New(config Config) (*Resource, error) {
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty.")
 	}
+	if config.TemplateURIVersion == "" {
+		return nil, microerror.Maskf(invalidConfigError, "config.TemplateURIVersion must not be empty.")
```
xh3b4sd: Super unimportant nitpick. These are not really sentences. So the full stop at the end of error message is wrong IMO. We never did it that way.

service/resource/deployment/resource.go
```diff
@@ -85,6 +84,9 @@ func New(config Config) (*Resource, error) {
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty.")
 	}
+	if config.TemplateURIVersion == "" {
```
xh3b4sd: Thinking loud. Do we have to ensure a certain format for the version?

service/resource/deployment/template.go
```diff
@@ -0,0 +1,18 @@
+package deployment
+
+import "fmt"
+
+const (
+	templateURIVersionDefault = "master"
+	templateURIFmt            = "https://raw.githubusercontent.com/giantswarm/azure-operator/%s/service/arm_templates/%s"
+
+	mainTemplate = "main.json"
```
xh3b4sd: I think these should be exported and have godoc.

service/resource/deployment/deployment.go
kopiczko: This comes from Azure.

service/resource/deployment/resource.go
kopiczko: copy/paste. I'll fix that.

service/resource/deployment/deployment.go
xh3b4sd: Ok, but what kind of format is this. Why 4 version parts?

service/resource/deployment/template.go
kopiczko: It's used only inside the package. I can add a godoc but I have no idea what it should be.

`mainTemplate is the name of the main template`? It's ridiculous.

service/resource/deployment/resource.go
kopiczko: Do you suggest types like in k8scc? Maybe. This is probably meant to come from the TPO. We'll see. This is also less strict than k8scc. GH URLs may change dynamically as oposed the the content of imported library.

service/resource/deployment/deployment.go
kopiczko: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates search for `contentVersion`

service/resource/deployment/template.go
xh3b4sd: It could explain what the main template is. I have no idea what name that is and why it is in JSON and so on.

service/resource/deployment/resource.go
xh3b4sd: I suggest to maybe validate a scheme. I did not talk about types.

service/resource/deployment/deployment.go
xh3b4sd: This link could be part of godoc.

service/resource/deployment/resource.go

service/resource/deployment/resource.go
kopiczko: ATM they are github branch name. Can change to something else in the future, like some blob storage on Azure. I don't want to spend time here now.

service/resource/deployment/template.go
kopiczko: This is it: https://github.com/giantswarm/azure-operator/blob/master/service/arm_templates/main.json

service/resource/deployment/spec.go
```diff
@@ -2,11 +2,16 @@ package deployment
...
+	Name          string
+	Parameters    map[string]interface{}
+	ResourceGroup string
+	TemplateURI   string
+
+	// TmplateContentVersion is value to fill in
+	// github.com/Azure/azure-sdk-for-go/arm/resources/resources.TemplateLink.ContentVersion.
+	// For more information see contentVersion documentation
+	// https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates.
+	TemplateContentVersion string
```
kopiczko: I did my best to describe it. I don't know why we have to specify this in both ARM template and sdk Deployment struct. This is confusing and it's the reason we have that here.

service/resource/deployment/template.go
kopiczko: So this is the name of the main ARM template. I don't know how to explain that better. I can rename to `mainARMTemplate` if that makes sense. ARM stands for Azure Resource Manager. Those templates are executed with deployments. The thing is I don't want to create Azure documentation in the godoc.

service/resource/deployment/deployment.go
kopiczko: Done.

kopiczko:
@xh3b4sd I think I addressed every comment. Let me know if there is something to improve there.

service/resource/deployment/resource.go
```diff
@@ -67,23 +67,26 @@ type Resource struct {
...
 	}
 	if config.CloudConfig == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.CloudConfig must not be empty.")
+		return nil, microerror.Maskf(invalidConfigError, "config.CloudConfig must not be empty")
 	}
 	if config.Logger == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty.")
+		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty")
+	}
+	if config.TemplateVersion == "" {
```
xh3b4sd: We usually differentiate between dependencies and settings here. Now the template version is validated with the dependencies but the template version is a setting.

service/resource/deployment/spec.go
```diff
@@ -2,11 +2,16 @@ package deployment
...
-	Parameters      map[string]interface{}
-	ResourceGroup   string
-	TemplateURI     string
-	TemplateVersion string
+	Name          string
+	Parameters    map[string]interface{}
+	ResourceGroup string
+	TemplateURI   string
+
+	// TmplateContentVersion is value to fill in
```
xh3b4sd: I think here is a typo.

service/resource/deployment/resource.go

service/resource/deployment/spec.go
kopiczko: Done.

PR #38, opened 2017-10-25
Vault access policies

Cocotton:


service/arm_templates/main.json
```diff
@@ -197,7 +201,25 @@
...
           "clusterID": {"value": "[parameters('clusterID')]"},
           "keyVaultName": {"value": "[parameters('keyVaultName')]"},
-          "keyVaultSecretsObject": {"value": "[parameters('keyVaultSecretsObject')]"}
+          "keyVaultSecretsObject": {"value": "[parameters('keyVaultSecretsObject')]"},
+          "masterNodesID": {
+            "copy": [{
+              "name": "value",
+              "count": "[length(parameters('mastersCustomConfig'))]",
+              "input": {
+                "id": "[resourceID('Microsoft.Compute/virtualMachines/', concat(parameters('clusterID'), '-', variables('masterPrefix'), '-', add(copyIndex('value'), 1)))]"
```
kopiczko: I think this should be `resourceId`. Does it work?

service/arm_templates/main.json
```diff
@@ -197,7 +201,25 @@
...
+              }
+            }]
+          },
+          "workerNodesID": {
+            "copy": [{
+              "name": "value",
+              "count": "[length(parameters('workersCustomConfig'))]",
+              "input": {
+                "id": "[resourceID('Microsoft.Compute/virtualMachines/', concat(parameters('clusterID'), '-', variables('workerPrefix'), '-', add(copyIndex('value'), 1)))]"
```
kopiczko: 1. I think this should be `resourceId`. Does it work?
2. Can we add `principalId` output in the virtualMachines resource, add `principalId` input here and reference outputs.principalId from virtualMachines? I find that nice, because it would encapsulate all MSI stuff in node_setup.json file.

service/arm_templates/certificates.json
```diff
@@ -18,14 +18,22 @@
       "keyVaultSKUName": {
         "type": "string",
         "defaultValue": "standard"
+      },
+      "masterNodesID": {
```
rossf7: Nitpick but masterNodeIDs reads better to me for the var name.

service/arm_templates/main.json
rossf7: +1 would be nice to have all the MSI stuff in one template.

Cocotton:
@kopiczko tested and working 😄

PR #41, opened 2017-11-07
vendor operatorkit changes

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2056


service/resource/deployment/error.go
```diff
@@ -4,11 +4,11 @@ import (
...
 
-var createTimeoutError = microerror.New("create timeout")
+var timeoutError = microerror.New("timeout")
 
-// IsCreateTimeoutError asserts createTimeoutError.
-func IsCreateTimeoutError(err error) bool {
-	return microerror.Cause(err) == createTimeoutError
+// IsTimeoutError asserts createTimeoutError.
+func IsTimeoutError(err error) bool {
+	return microerror.Cause(err) == timeoutError
```
xh3b4sd: That has nothing to do with vendoring AFAIK.

service/resource/deployment/resource.go
```diff
@@ -181,27 +181,29 @@ func (r *Resource) GetCreateState(ctx context.Context, obj, currentState, desire
 		}
 	}
 
-	return deploymentsToCreate, nil
-}
+	patch := framework.NewPatch()
+
+	if len(deploymentsToCreate) > 0 {
+		patch.SetCreateChange(deploymentsToCreate)
+	}
```
xh3b4sd: I would like to keep it the way we did it so far. There are `New*Patch` methods which are all the same and only call `new*Change` methods. Keeping this consistent makes it easier later on to refactor and unify. The tests can then stay as they are and test the private methods since there is no real business logic in the `New*Patch` methods.

service/resource/deployment/error.go
kopiczko: You are right. I created more generic error, which I think is the general direction we follow.
Do you want to revert?

xh3b4sd: No. Lets just try to do one thing in a PR. I screw this on my own quite a lot. Lets just keep trying.

service/resource/deployment/resource.go

service/resource/deployment/resource.go
```diff
@@ -161,47 +161,37 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 	}
 
-	var deploymentsToCreate []Deployment
-
-	for _, desiredDeployment := range desiredDeployments {
-		if !existsDeploymentByName(currentDeployments, desiredDeployment.Name) {
-			deploymentsToCreate = append(deploymentsToCreate, desiredDeployment)
-		}
+	if len(deploymentsToCreate) > 0 {
+		patch.SetCreateChange(deploymentsToCreate)
```
xh3b4sd: Why are we using a condition here? I think we should always set the patch to have the unified `NewUpdatePatch ` methods. In this case `newCreateChange` should just make sure to return `nil` in case no deployments should be created.

service/resource/deployment/resource_test.go
```diff
@@ -187,110 +189,14 @@ func Test_Resource_Deployment_GetCreateState(t *testing.T) {
...
-			t.Fatalf("case %d expected '%T', got '%T'", i+1, []Deployment{}, deployments)
-		}
-
-		if len(tc.ExpectedDeployments) != len(deployments) {
-			t.Fatalf("case %d expected %d deployments got %d", i+1, len(tc.ExpectedDeployments), len(deployments))
-		}
-	}
-}
-
-func Test_Resource_Deployment_GetDeleteState(t *testing.T) {
```
xh3b4sd: Why are we removing the delete tests?

service/resource/deployment/resource.go
kopiczko: If `newCreateChange` returns a nil we still need the condition. Otherwise `ApplyCreateChange` will be invoked.

service/resource/deployment/resource_test.go
kopiczko: Because `NewDeletePatch` is noop. Those tests were an unnecessary clutter IMO.

service/resource/deployment/resource.go
xh3b4sd: Yes, and `ApplyCreateChange` can just do nothing in case there is no useful create state given. This is how we did it so far in all of the resource implementations. See for instance https://github.com/giantswarm/kvm-operator/blob/7bf46cd0c5130d104516c854c420d78b99117eef/service/resource/namespace/create.go#L23-L36.

xh3b4sd: I find the debug logging that we get that way useful.

service/resource/deployment/resource_test.go
xh3b4sd: Ack.

service/resource/deployment/resource.go
kopiczko: > I think we should always set the patch to have the unified NewUpdatePatch methods.

That doesn't make sense to me. What you wrote boils down to this interface:

```
type Resource interface {
	GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error)
	GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error)

	GetCreateState(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error)
	GetDeleteState(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error)
	GetUpdateState(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error)

	ProcessCreateState(ctx context.Context, obj, createState interface{}) error
	ProcessDeleteState(ctx context.Context, obj, deleteState interface{}) error
	ProcessUpdateState(ctx context.Context, obj, updateState interface{}) error
}
```

Which differs from the old one only by this line:

```
	GetUpdateState(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, interface{}, interface{}, error)
```

By providing https://github.com/giantswarm/operatorkit/pull/92 this PR I wanted allow the Resource deciding which actions are necessary to reconcile towards the desired state.

Say a resource in order to update has to create and delete something. For example bridge name changes and the resource has to create a new one and delete the old one. To achieve that it sets create and delete patch parts.

Say another resource in order to update it just has to call update api. For example pod definition changes and the resource has to call K8s update api. To achieve that is sets update patch part.

I don't think we can have unified `NewUpdatePatch` methods. Otherwise it wouldn't make sense to provide Patch struct at all.

- - -

I will change according to your request.

xh3b4sd: Why did you do it now after writing a book about not to do it?

kopiczko: I wrote a book to make sure we're on the same page.
I did it like you want because I don't care that much and this makes you happy <3

xh3b4sd: Ok, to be on the same page. My goal was not to have fully unified `New*Patch` methods. The patch concept is powerful in that sense that the implementor can choose if create/update/delete state should be computed. In case updating only means updating it is ok to only dispatch an update change. In another case it is also ok to dispatch an update and a create change (or an additional delete change). I only think that we should not use any conditions within the `New*Patch` methods to keep them free of dependencies and logic. I see them only being used to dispatch changes, while they are free to choose which changes (create, update, delete) to dispatch.

kopiczko: I wanted to do it other way around. Keep logic in `New*Patch` and remove as much logic from `Apply*Change` as possible. The reason for that is we don't test `Apply*Change` functions, and testing `New*Patch` functions is easy. Testing `Apply*Change` would probably require some third party component like Kubernetes, or etcd, or even whole OS, so it makes sense for me to keeps those things lean.

I understand you like the logs. I think most of them can be in the `New*Patch` as well, but I didn't think much about it yet.

xh3b4sd: IMO there shouldn't be any logic but apply the state using some dependency/client. I do not mind one simple condition though, especially not when this has always the same purpose.

PR #42, opened 2017-11-09
resource: improve logs

kopiczko:


service/resource/deployment/resource.go
```diff
@@ -229,13 +231,15 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createState inter
...
 
-			r.logger.Log("cluster", key.ClusterID(customObject), "debug", fmt.Sprintf("created deployment %s", deploy.Name))
+			r.logger.Log("cluster", key.ClusterID(customObject), "debug", fmt.Sprintf("creating Azure deployments: creating %#q: created", deploy.Name))
 		}
 
-		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "created the deployments in the Azure API")
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "creating Azure deployments: created")
+	} else {
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "creating Azure deployments: already created")
 	}
```
kopiczko: Besides adding logs when resource is already reconciled I made them grep friendly. Now you can do something like: `kubectl  logs ... | grep CLUSTER_ID | grep "creating Azure deployments"`. Or `kubectl  logs ... | grep CLUSTER_ID | grep "creating Azure deployments"`.

kopiczko: This also applies to errors.

service/resource/resourcegroup/error.go
```diff
@@ -18,9 +18,9 @@ func IsWrongTypeError(err error) bool {
...
-var deleteTimeoutError = microerror.New("delete timeout")
+var timeoutError = microerror.New("timeout")
 
-// IsDeleteTimeoutError asserts deleteTimeoutError.
-func IsDeleteTimeoutError(err error) bool {
-	return microerror.Cause(err) == deleteTimeoutError
+// IsTimeoutError asserts deleteTimeoutError.
+func IsTimeoutError(err error) bool {
+	return microerror.Cause(err) == timeoutError
 }
```
kopiczko: Ninja fix: this did not have sense when the error is nicely annotated. Also we go towards generic direction.

service/resource/deployment/resource.go
```diff
@@ -57,6 +58,7 @@ func DefaultConfig() Config {
...
 }
 
+// Resource manages Azure ARM templates deployments.
 type Resource struct {
```
kopiczko: I started to use automated linter and this is a bitch :) So expect me to add some godoc.

xh3b4sd: IMO we started to remove these comments since they are quite useless and everywhere the same. Should we keep them though?

kopiczko: OK, I'll remove. I thought for generic names like this it makes sense. But probably I end up adding comments like `getCerts returns a cert.`.

PR #45, opened 2017-11-24
[WIP] Etcd dns

Cocotton:
This PR removes the etcd private load balancer since this is not a viable option due to Azure limitations (e.g. a load balancer pool member cannot reach the load balancer; the master node could not reach the etcd load balancer since they are a pool member). 

Instead, if exposes etcd with a DNS record (master node private ip address).

The security rules are also updated to allow the health probe of the api/ingress load balancers to reach their backend pool members.

The public load balancers forward the client ip address  instead of their own. So instead of making security rules accepting traffic from the load balancer public ip, we instead open the traffic to all.

**CONCERNS / WIP**
Since etcd  CANNOT be frontend by a load balancer, we can't map the port 443 to 2379. Instead, we need to directly reach the port 2379.

I managed to make everything work and have a, what seems to be, functioning guest cluster by updating the port (from 443 to 2379) on the following lines:

- https://github.com/giantswarm/azure-operator/blob/master/vendor/github.com/giantswarm/k8scloudconfig/templates.go#L51
- https://github.com/giantswarm/azure-operator/blob/master/vendor/github.com/giantswarm/k8scloudconfig/templates.go#L1095
- https://github.com/giantswarm/azure-operator/blob/master/vendor/github.com/giantswarm/k8scloudconfig/templates.go#L1263

Any ideas of what would be the best way to set the etcd port other than hardcoding it?


service/arm_templates/main.json
```diff
@@ -366,7 +325,7 @@
         "parameters": {
           "dnsZone": { "value": "[parameters('dnsZones').etcd]" },
           "prefix": { "value": "[concat('etcd.', parameters('clusterID'))]" },
-          "ipAddress": { "value": "[reference('etcd_load_balancer_setup').outputs.ipAddress.value]" }
+          "ipAddress": { "value": "[reference('master_nodes_setup_1').outputs.ipAddress.value]" }
```
Cocotton: This is a bit dirty. I assume there is only a single master.

If you prefer, I can try to find a way to set a round robin DNS in case there is more than one master nodes.

service/arm_templates/private_load_balancer_setup.json
```diff
@@ -1,114 +0,0 @@
-{
```
Cocotton: This template is not required anymore since we don't create any private load balancer.

service/arm_templates/security_groups_setup.json
```diff
@@ -30,22 +21,23 @@
...
-    },
     "kubernetesIngressSecurePort": {
       "type": "int"
     },
     "kubernetesIngressInsecurePort": {
       "type": "int"
+    },
+    "azureHealthProbeCidr": {
+      "type": "string",
+      "defaultValue": "168.63.129.16/32"
```
Cocotton: Azure uses this fixed public IP address from which it probes the load balancer backend pool members.

See this for more information: https://blogs.msdn.microsoft.com/mast/2015/05/18/what-is-the-ip-address-168-63-129-16/

service/cloudconfig/template.go
```diff
@@ -6,6 +6,11 @@ const (
 	getKeyVaultSecretsFileName = "/opt/bin/get-keyvault-secrets"
 	getKeyVaultSecretsTemplate = `#!/bin/bash -e
 
+until $(curl --fail --output /dev/null http://localhost:50342/oauth2/token --data "resource=https://vault.azure.net" -H Metadata:true); do
```
Cocotton: This script launches before the azure vault is created, so it fails by default.

By adding this (kind of ugly?) until condition, it just waits for the vault and the secrets to be ready before fetching them.

r7vme:
@Cocotton regarding etcd port.

It's not a problem (almost):

1) Calico in Azure will be in policy-only mode and will use k8s, no need to specify etcd endpoint. But this will be done later, i still don;t know best way to put other calico configuration to k8scc.
2) advertise url imo we can just skip it for now, it won't break anything as far as i know
3) api server, please update k8scc it was fixed already  https://github.com/giantswarm/k8scloudconfig/commit/be9f399432713d5ec7223c3eca6fd9ecfb922456

tl;dr only apiserver on Azure will be using etcd and it was already fixed to localhost in latest k8scc

Cocotton:
@r7vme, updating the k8scloudconfig vendor brings 2 problems 

1. The master cloudconfig template is now too big for Azure. It looks like the whole base64 encoded cloudconfig template weights a bit more than 95kb, but Azure [has a limit of 65.5kb](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines#osprofile-object). I managed to work around this by Gzipping and base64 encoding some of the biggest elements in the `write_files:` section of the template.
2. It looks like the k8s api service is not booting up correctly because the `/etc/kubernetes/encryption/k8s-encryption-config.yaml` file can't be found. I did not have much time to dig on that today, but will investigate that asap :)

service/cloudconfig/template.go
kopiczko: Let's please split it to a separate PR so we can merge it quickly.

kopiczko:
@Cocotton: I discussed with @r7vme. Let's make a branch of k8scloudconfig repo with changes you applied. Then we'll vendor the branch. After that we'll figure out how to deal with merging that branch.

r7vme:
@Cocotton 

i managed to w/a this same way by gzipping whole cloud-config.

https://github.com/giantswarm/azure-terraform/blob/master/cloud-config/cloud-config.sh.tmlp
https://github.com/giantswarm/azure-terraform/blob/master/modules/cloud-config-compressed/cloud-config.tf#L59-L67

I think in long-term it makes sense to implement gzipping in azure-operator. For me ~90k compresses to ~22k. So we still have big window

cc: @kopiczko

service/arm_templates/main.json
r7vme: Fine for now.

service/arm_templates/security_groups_setup.json
r7vme: 1) there is special tag "AzureLoadBalancer" that should be used for loadbalancer traffic. See link
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-nsg

2) This is probably not needed at all, because LoadBalancer traffic opened by default.
See Inbound default rules in link above.

If it does not work for you, let's chat in slack, because i have this working in host cluster.

service/arm_templates/security_groups_setup.json
```diff
@@ -100,39 +92,39 @@
...
+              "sourceAddressPrefix": "[parameters('workerSubnetCidr')]",
               "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
               "access": "Allow",
               "direction": "Inbound",
               "priority": "3800"
             }
           },
           {
-            "name": "workerSubnetToMasterSubnet",
+            "name": "azureHealthProbeRule",
```
r7vme: not needed probably , see above

service/arm_templates/security_groups_setup.json
```diff
@@ -230,6 +222,20 @@
               "destinationAddressPrefix": "[parameters('workerSubnetCidr')]",
               "access": "Allow",
               "direction": "Inbound",
+              "priority": "3700"
+            }
+          },
+          {
+            "name": "azureHealthProbeRule",
```
r7vme: not needed probably, see above

service/arm_templates/security_groups_setup.json
```diff
@@ -100,39 +92,39 @@
           {
             "name": "apiLoadBalancerRule",
             "properties": {
-              "description": "Allow the API load balancer to reach the kubernetes API.",
+              "description": "Allow anyone to reach the kubernetes API.",
               "protocol": "tcp",
               "sourcePortRange": "*",
```
r7vme: SourcePortRange 443

service/arm_templates/security_groups_setup.json
```diff
@@ -100,39 +92,39 @@
...
+              "sourceAddressPrefix": "*",
               "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
               "access": "Allow",
               "direction": "Inbound",
               "priority": "3900"
             }
           },
           {
-            "name": "etcdLoadBalancerRule",
+            "name": "workerSubnetToMasterSubnet",
```
r7vme: This is already allowed by default. 

See https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-nsg

AllowVNetInBound allows ALL traffic within virtual network.

Cocotton: As discussed, we override the default Azure rules. Because of that, we need to explicitly open traffic between the subnets.

Cocotton:
@kopiczko @r7vme, I tested the deployment using a cloudconfig with compressed files. The deployment is successful, but there seems to be some other problems related to the cloud-config file.

As we said in the talk, the cloud-config is out of scope of this PR.

Are you fine with me merging this PR now?

Next step will be to setup a script for the cloud-config like we discussed on the talk, as well as create the empty route table (2 different PRs)

service/arm_templates/security_groups_setup.json
```diff
@@ -198,7 +172,7 @@
               "protocol": "tcp",
               "sourcePortRange": "*",
               "destinationPortRange": "[string(parameters('kubernetesIngressSecurePort'))]",
-              "sourceAddressPrefix": "[parameters('ingressLoadBalancerCidr')]",
+              "sourceAddressPrefix": "*",
```
kopiczko: Nitpick: I think the description should be updated.

service/arm_templates/security_groups_setup.json
```diff
@@ -212,7 +186,7 @@
               "protocol": "tcp",
               "sourcePortRange": "*",
               "destinationPortRange": "[string(parameters('kubernetesIngressInsecurePort'))]",
-              "sourceAddressPrefix": "[parameters('ingressLoadBalancerCidr')]",
+              "sourceAddressPrefix": "*",
```
kopiczko: I think this rule is a dupe of previous one.

service/arm_templates/security_groups_setup.json
Cocotton: Description updated

service/arm_templates/security_groups_setup.json
Cocotton: This one is for the insecure port, the previous for the secure port.

PR #46, opened 2017-11-24
glide: update vendor

kopiczko:


kopiczko:
Won't work because of chnages in azuretpr.

PR #48, opened 2017-11-27
Adds Helm chart

r7vme:
Minimal drop required to use in `gollum` Azure host cluster.

Related change: https://github.com/giantswarm/installations/pull/153

### DIFF
--- a/helm/azure-operator-chart/templates/configmap.yaml
+++ b/helm/azure-operator-chart/templates/configmap.yaml
@@ -8,10 +8,5 @@ data:
     server:
       listen:
         address: 'http://0.0.0.0:8000'
-    service:
-      azure:
-        template:
-          uri:
-            version: "master"
       kubernetes:
         incluster: true
--- a/helm/azure-operator-chart/templates/configmap.yaml
+++ b/helm/azure-operator-chart/templates/configmap.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: azure-operator-configmap
+  namespace: giantswarm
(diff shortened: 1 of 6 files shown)


helm/azure-operator-chart/templates/configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+data:
+  config.yaml: |
+    server:
+      listen:
+        address: 'http://0.0.0.0:8000'
+    service:
+      azure:
+        template:
+          uri:
+            version: "master"
```
kopiczko: This should be removed. azure-operator defaults to the build git SHA. That way we have predictable template versions.

helm/azure-operator-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+        args:
+        - daemon
+        - --config.dirs=/var/run/azure-operator/configmap/
+        - --config.dirs=/var/run/azure-operator/secret/
+        - --config.files=config
+        - --config.files=secret
+        resources:
+          requests:
+            cpu: 100m
+            memory: 20Mi
```
kopiczko: We started to assign the same values for requests and limits in other operators I think.

helm/azure-operator-chart/templates/configmap.yaml
r7vme: done

helm/azure-operator-chart/templates/deployment.yaml
r7vme: done

PR #51, opened 2017-11-28
service: do not pass logger explicitly

kopiczko:


kopiczko:
@xh3b4sd I'd like to discuss that.

(This won't compile, it's on purpose I don't want to change everything at once, only show an example)

Here I replaced Logger dependency with LoggerConfig. What I find nice is we can have a default so we save lines passing logger all over the place. in many cases we'll be able to use just `packagex.DefaultConfig()` w/o setting anything, or passing only important stuff. WDYT?

xh3b4sd:
I don't understand the problem this tries to solve. I do not see any. I understand you want to get rid of the need to configure something and you try to do this by configure something which adds another level of configuration. I would just keep it as it is. When you do not care about the logger then set `microloggertest.New()`. Same for the Kubernetes client.

kopiczko:
Let's keep it like it is.

PR #54, opened 2017-11-29
Adds Azure cloud provider config

r7vme:
this config used by Kubernetes Azure cloud provider.

### DIFF
--- a/service/cloudconfig/cloud_config.go
+++ b/service/cloudconfig/cloud_config.go
@@ -17,8 +17,14 @@ func (me *MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
 		return nil, microerror.Mask(err)
 	}
 
+	cloudProviderConf, err := me.masterCloudProviderConf()
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
 	newFiles = []k8scloudconfig.FileAsset{
 		getSecretsScript,
+		cloudProviderConf,
 	}
 
 	return newFiles, nil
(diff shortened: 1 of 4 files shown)


service/key/key.go
```diff
@@ -22,13 +38,48 @@ func KeyVaultName(customObject azuretpr.CustomObject) string {
 	return customObject.Spec.Azure.KeyVault.Name
 }
 
+// MasterSecurityGroupName returns name of the security group attached to master subnet.
+func MasterSecurityGroupName(customObject azuretpr.CustomObject) string {
```
kopiczko: Let's track adding those to ARM template parameters as well.

service/cloudconfig/cloud_config.go
```diff
@@ -48,6 +54,53 @@ func (me *MasterExtension) VerbatimSections() []k8scloudconfig.VerbatimSection {
 	return nil
 }
 
+// masterCloudProviderConf returns Kubernetes cloud provider config for Azure.
+func (me *MasterExtension) masterCloudProviderConf() (k8scloudconfig.FileAsset, error) {
```
kopiczko: Please add TODO we should make it `cloudProviderConf` as soon as we start creating masters and worker in the same subnet and SG.

r7vme: Hmm, do we plan to use same subnet?

kopiczko: Did I misunderstand you? I thought you suggested merging those two guys:


* pawel-cluster-VirtualNetwork-MasterSubnet
* pawel-cluster-VirtualNetwork-WorkerSubnet

r7vme: i forgot why i suggested this... if only for simplicity, Than we can avoid this and leave as it is right now as two networks will allow us properly assign security rules for master.

I mean for master we need to open some extra ports so proper is to have it done only for master subnet. Sorry for confusion :)

service/key/key.go
r7vme: Can we add this in TODO?

I can do this as separate PR, because it's quite big change.

kopiczko: Yes, TODO is ok.

PR #55, opened 2017-11-29
service: provide AzureConfig to CloudConfig

kopiczko:
We need AzureConfig inside CloudConfig to finish https://github.com/giantswarm/azure-operator/pull/54

### DIFF
--- a/main.go
+++ b/main.go
@@ -31,6 +31,13 @@ var (
 )
 
 func main() {
+	err := mainWithError()
+	if err != nil {
+		panic(fmt.Sprintf("%#v\n", err))
+	}
+}
+
+func mainWithError() error {
 	var err error
 
 	// Create a new logger which is used by all packages.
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -38,7 +40,7 @@ func main() {
 		loggerConfig.IOWriter = os.Stdout
 		newLogger, err = micrologger.New(loggerConfig)
 		if err != nil {
-			panic(err)
+			panic(fmt.Sprintf("%#v", microerror.Maskf(err, "micrologger.New")))
```
xh3b4sd: Would be cool do it like this: https://github.com/giantswarm/draughtsman-eventer/blob/fd040324c12b324e801e72a6e3cfda10611155d8/main.go#L31 as well.

kopiczko: I think there it's even worse: https://github.com/giantswarm/draughtsman-eventer/blob/fd040324c12b324e801e72a6e3cfda10611155d8/main.go#L71

I can save I panic for crating a logger.

kopiczko: I added mainWithError. 2 panic calls saved. It's something :)

kopiczko: I'd like to get rid of this server factory at some point. Did you have some thoughts about it?

xh3b4sd: My thinking is it will not go away. I cannot think of any way. It is also super unimportant. What would be nice would be to extend its interface to be able to return an error but this seems to be difficult as well and also super unimportant.

PR #63, opened 2017-12-01
Msi roles assignment

Cocotton:
This PR adds the parameter `assignContributorRole` to the node_setup.json template.

When passing `true`, to node is assigned the "contributor role" in the resource group.

I added that flag because I was not sure we wanted to give that role to the worker nodes.

Unfortunately, I cannot test this completely because I don't have the permission to give the contributor role to a VM. See the following error
```
The client 'XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX' with object id 'XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX' does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write'
```

### DIFF
--- a/service/arm_templates/main.json
+++ b/service/arm_templates/main.json
@@ -438,7 +438,8 @@
           "osImageSKU": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.sku]"},
           "osImageVersion": {"value": "[parameters('mastersCustomConfig')[copyIndex()].osImage.version]"},
           "cloudConfigData": {"value": "[parameters('masterCloudConfigData')]"},
-          "loadBalancerBackendPoolIDs": {"value": "[createArray(reference('api_load_balancer_setup').outputs.loadBalancerBackendPoolID.value)]"}
+          "loadBalancerBackendPoolIDs": {"value": "[createArray(reference('api_load_balancer_setup').outputs.loadBalancerBackendPoolID.value)]"},
+          "assignContributorRole": {"value": "[bool('true')]"}
         }
       }
     },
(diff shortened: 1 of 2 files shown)


service/arm_templates/node_setup.json
```diff
@@ -71,6 +78,16 @@
     },
     "loadBalancerBackendPoolIDs": {
       "type": "array"
+    },
+    "assignContributorRole": {
+      "type": "bool"
+    },
+    "contributorRoleDefinitionId": {
+      "type": "string",
+      "defaultValue": "b24988ac-6180-42a0-ab88-20f7382dd24c",
```
Cocotton: This seems to be the GUID that is always attributed to the contributor role, no matter the account/subscription/resource group. 

Take a look here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-api-authentication#get-azure-rbac-role-identifier

service/arm_templates/main.json
```diff
@@ -468,7 +469,8 @@
           "osImageSKU": {"value": "[parameters('workersCustomConfig')[copyIndex()].osImage.sku]"},
           "osImageVersion": {"value": "[parameters('workersCustomConfig')[copyIndex()].osImage.version]"},
           "cloudConfigData": {"value": "[parameters('workerCloudConfigData')]"},
-          "loadBalancerBackendPoolIDs": {"value": "[createArray(reference('ingress_load_balancer_setup').outputs.loadBalancerBackendPoolID.value)]"}
+          "loadBalancerBackendPoolIDs": {"value": "[createArray(reference('ingress_load_balancer_setup').outputs.loadBalancerBackendPoolID.value)]"},
+          "assignContributorRole": {"value": "[bool('false')]"}
```
Cocotton: If we also need to give the contributor role to the worker nodes, I'll just remove that condition as well as in the node_setup template and assign the role to all VMs

r7vme: We need MSI for all nodes in cluster.

kopiczko: Se we can remove the flag as well.

service/arm_templates/node_setup.json
kopiczko: OMG Couldn't that be a `Contributor`...

Cocotton: I wish, that feels so dirty..

PR #64, opened 2017-12-03
Fix msi rbac

r7vme:
- Fix role assignment. Azure complained about name, because it should be in `guid` format.
- We need additional role for azure-operator account to assign MSI roles. Added section how to create azure-operator service principal and assign additional role.

### DIFF
--- a/service/arm_templates/node_setup.json
+++ b/service/arm_templates/node_setup.json
@@ -210,7 +210,7 @@
     {
       "type": "Microsoft.Authorization/roleAssignments",
       "apiVersion": "2014-10-01-preview",
-      "name": "[variables('roleAssignmentName')]",
+      "name": "[guid(variables('roleAssignmentName'))]",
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -60,6 +60,19 @@ docker build -t quay.io/giantswarm/azure-operator:local-lab .
...
+
+Azure operator requires API access. To archieve this please create service principal.
+
+```
+az ad sp create-for-rbac --name=azure-operator-sp --role="Contributor" --scopes="/subscriptions/${SUBSCRIPTION}"
+```
+
+Assign additional role "User Access Administrator". Required to assign roles on Virtual Machines (MSI).
+```
+az role assignment create --assignee ${APPID} --role "18d7d88d-d35e-4fb5-a5c3-7773c20a72d9"
```
kopiczko: This ID will change from subscription to subscription.

PR #68, opened 2017-12-05
dnsrecord: add new resource

kopiczko:
This resource takes name servers from the DNS zone created inside guest cluster
and creates an NS record inside host cluster zone.

Example:

Guest cluster DNS zone:

eggs2.k8s.gollum.westeurope.azure.gigantic.io
@ ns1-01.azure-dns.com.

Host cluster DNS zone:

gollum.westeurope.azure.gigantic.io
@ ns5-05.azure-dns.com.
NS eggs2.k8s ns1-01.azure-dns.com.


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/giantswarm/azure-operator
 import:
 - package: github.com/giantswarm/azuretpr
-  version: master
+  version: resource-group
```
xh3b4sd: Are we going with TPRs still here? Do we have to migrate here to CRDs as well?

service/operator/framework.go
```diff
@@ -3,6 +3,7 @@ package operator
 import (
 	"github.com/giantswarm/azure-operator/service/cloudconfig"
 	"github.com/giantswarm/azure-operator/service/resource/deployment"
+	"github.com/giantswarm/azure-operator/service/resource/dnsrecord"
```
xh3b4sd: I would like to start versioning this by name as well. Should be `dnsrecordv1` then and the package directive and the resource name should reflect this as well.

service/operator/framework.go
```diff
@@ -39,33 +40,45 @@ func newFramework(config Config) (*framework.Framework, error) {
...
 
-		resourceGroupResource, err = resourcegroup.New(resourceGroupConfig)
+		resourceGroupResource, err = resourcegroup.New(c)
 		if err != nil {
 			return nil, microerror.Maskf(err, "resourcegroup.New")
 		}
 	}
 
-	var deploymentResource framework.Resource
+	var deploymentResource *deployment.Resource
```
xh3b4sd: Why this change?

service/operator/framework.go
```diff
@@ -39,33 +40,45 @@ func newFramework(config Config) (*framework.Framework, error) {
...
 	{
-		deploymentConfig := deployment.DefaultConfig()
-		deploymentConfig.TemplateVersion = config.TemplateVersion
-		deploymentConfig.AzureConfig = config.AzureConfig
-		deploymentConfig.CertWatcher = certWatcher
-		deploymentConfig.CloudConfig = cloudConfig
-		deploymentConfig.Logger = config.Logger
-
-		deploymentResource, err = deployment.New(deploymentConfig)
+		c := deployment.DefaultConfig()
```
xh3b4sd: I like the short `c` here, but I personally would appreciate new lines between logical blocks.

service/resource/dnsrecord/conv.go
```diff
@@ -0,0 +1,33 @@
+package dnsrecord
+
+import (
+	"github.com/giantswarm/azuretpr"
+	"github.com/giantswarm/microerror"
+)
+
+func toCustomObject(v interface{}) (azuretpr.CustomObject, error) {
```
xh3b4sd: IMO this should be inside the key package. At least the other operators are doing it this way. Also note that we should version the key packages as well then. Also, what does `conv.go` stand for? Would be nice to keep this with other operators aligned.

service/resource/dnsrecord/resource.go
```diff
@@ -0,0 +1,351 @@
...
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/framework"
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/key"
+)
+
+const (
+	// Name is the identifier of the resource.
+	Name = "resourcegroup"
```
xh3b4sd: The package name of the resource is `dnsrecord`, so the resource name should be aligned with that IMO.

service/resource/dnsrecord/resource.go
```diff
@@ -0,0 +1,351 @@
...
+	return newService, nil
+}
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	o, err := toCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return r.getCurrentState(ctx, o)
+}
```
xh3b4sd: The other operators define this in different files. I thought we already had consensus on this practise.

service/resource/dnsrecord/resource.go
```diff
@@ -0,0 +1,351 @@
...
+	patch := framework.NewPatch()
+
+	updateChange, err := r.newUpdateChange(ctx, o, c, d)
+	if err != nil {
+		return nil, microerror.Maskf(err, "NewUpdatePatch")
+	}
+
+	patch.SetUpdateChange(updateChange)
+	return patch, nil
+}
```
xh3b4sd: We have a more simplified pattern with this function in other operators as well.

service/operator/framework.go
kopiczko: I believe exposing the original type as long as possible is a good thing, because hiding it behind an interface hides implementation details.

service/operator/framework.go
kopiczko: This is 1:1 change. I only changed name. Where are the logical blocks?

service/resource/dnsrecord/conv.go
kopiczko: It was copy/paste from other resource. I only moved that from `resource.go` to `conv.go`. conv stands for conversion. Maybe it isn't the best name since in Go terms it's type assertion.

I can move toCustomObject to `key` package, that makes sense, since it's reused between resources, but for `toDNSRecords` it's impossible since it returns a private type.

service/resource/dnsrecord/resource.go
kopiczko: Yes, copy/paste. Thanks!

service/resource/dnsrecord/resource.go
kopiczko: Again copy/paste. I'm not big on that, but will change.

service/resource/dnsrecord/resource.go
kopiczko: I extracted more private methods. E.g. for `GetDesiredState` I created a method `getDesiredState` which takes proper types. I think this allows to skip the boilerplate and focus on important parts when reading the code.
Natural consequence of that was to make `newUpdateChange` properly typed so I did conversions here. I could extract `newUpdatePatch` so it'd be more aligned maybe. WDYT?

service/operator/framework.go
kopiczko: I'm not yet fluent with package versioning scheme we have. I want to leave that for later and change all packages at once.

glide.yaml
kopiczko: We're going with TPR. I don't want to bother with vendor hacks, and update operatorkit first. Migration is cheap as long as we don't have customers we can burn all clusters to the ground.

kopiczko:
> The review would be easier if the PR would dedicate to smaller change sets.

I think this change is very atomic. It just introduces a new resource. What I could improve?

glide.yaml
xh3b4sd: Ack. Should I sort the migration then or do you guys want to sort it?

service/operator/framework.go
xh3b4sd: Disagree and commit.

service/operator/framework.go
xh3b4sd:
```
c := deployment.DefaultConfig()

<dependencies>

<settings>

<constructor magic>
```

service/resource/dnsrecord/conv.go
xh3b4sd: `toDNSRecords` is about the resource so this is fine here. The custom object assertion should move to the key package though.

service/resource/dnsrecord/resource.go
xh3b4sd: Thanks.

glide.yaml
kopiczko: If you want to do it you're welcome. Otherwise I'll do that. But let's first finish with this PR to avoid merging conflicts.

xh3b4sd:
We already split introductions of new resources in separate PRs. 

- Create the resource package with interface but without logic. 
- Split into separate files. 
- Add logic for creation. 
- Add logic for deletion. 
- Add logic for updates. 
- Add logic for metrics.
- You name it.

service/resource/dnsrecord/resource.go
xh3b4sd: AFAIK we already talked about keeping `New*Patch` methods dumb to only gather the required changes. Adding specific type assertions here makes them more unique and always different which makes potential additional abstractions and human reading harder. All logic and type assertion happens in the `new*Change` methods so far and this project here should not differ out of a sudden. We can talk about how to do it in general but I would prefer keeping it consistent across the projects since this is a very powerful property of our projects in general.

service/operator/framework.go
kopiczko: I tried to do that, but there are so many unrelated changes, I'd like to do that in separate PR.

glide.yaml
xh3b4sd: It would take some workload off of me when you sort this here.

service/operator/framework.go
xh3b4sd: Easy. All good.

glide.yaml
kopiczko: Ok, don't worry then.

service/resource/dnsrecord/resource.go
kopiczko: Done.

client/azure.go
```diff
@@ -73,6 +74,10 @@ type AzureClientSet struct {
 	DeploymentsClient *resources.DeploymentsClient
 	// GroupsClient manages ARM resource groups.
 	GroupsClient *resources.GroupsClient
+	// DNSRecordSetsClient manages DNS zones' records.
```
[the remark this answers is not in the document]
kopiczko: I think it's ok for plural.

service/resource/dnsrecord/delete.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return microerror.Maskf(err, "deleting host cluster DNS records")
+	}
+	c, err := toDNSRecords(change)
+	if err != nil {
+		return microerror.Maskf(err, "deleting host cluster DNS records")
+	}
+
+	return r.applyDeleteChange(ctx, o, c)
+}
```
kopiczko: @xh3b4sd I like having type assertions in separation and then call private method with non-interface{} types. That makes it easier to filter this boilerplate. I'd like to do the same for New*Patch methods. I'd like this to become our "best practice". Let's discuss here.

xh3b4sd: I think this is backwards. Having the common ground being abstracted and not polluted with custom code is a good thing. I would like to aim for some more abstraction and even code generation in the long run and I think your suggestion would not ease that but create more rules. So far we agreed on keeping the common ground as broad as possible and I would like to keep that. You can then do all custom magic in the private methods. The boilerplate you need to filter is always the first block. I do not see an issue with this personally. When you are big on this go ahead and lets see how it works out in this project but I would like to keep the majority of operators as they are for now until we have a stronger agreement.

kopiczko: > I would like to aim for some more abstraction and even code generation in the long run and I think your suggestion would not ease that but create more rules. 

I could imagine generated code could look like:

```
func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
	o, err := toCustomObject(obj)
	if err != nil {
		return nil, microerror.Maskf(err, "GetDesiredState")
	}

	return r.getDesiredState(ctx, o)
}
```

You'd just have to implement `getDesiredState`.

xh3b4sd: But this is just one part of the puzzle and the missing rest does not easily work that way because how should you guess the types being required? And doing it half that way and half another way is not going to work out.

kopiczko: Types can be passed as arguments.

What would you like to generate then?

xh3b4sd: Things like this for instance. 

```
func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*framework.Patch, error) {
	create, err := r.newCreateChange(ctx, obj, currentState, desiredState)
	if err != nil {
	}

	update, err := r.newUpdateChange(ctx, obj, currentState, desiredState)
	if err != nil {
	}

	patch := framework.NewPatch()
	patch.SetCreateChange(create)
	patch.SetUpdateChange(update)

	return patch, nil
}
```

This is super generic without anything special. Guessing the private "change" methods is easy because this is a strict pattern and the outcome is predictable because it is always the same based on the private "change" methods that are implemented. Generating things like this is not going to work when we put the custom type assertion in their like you suggest, because where does the knowledge of the types come from and how does the naming of the methods than work? I think this makes it just more complicated.

kopiczko: We were discussing already this can't be generalized as sometimes update means creation & deletion. This is when things are immutable.

I thought of code generation as a compensation of lack of generic types in Go. So we had very different image of that.

kopiczko: I extracted `newUpdatePatch` and `newDeletePatch` which are as simple as possible, but have proper types.

I don't want to spend more time on that. I can get rid of all the private methods if that's the only thing missing to get an approval.

PR #71, opened 2017-12-12
Managed disks

Cocotton:
Move the virtual machines disks and availability sets from classic to managed.

We provide the main ARM template with the new specs found in [this azuretpr pull request](https://github.com/giantswarm/azuretpr/pull/13). More information about the decision of that structure are document in the linked PR.

Right now I'm having trouble with the roleAssignment. I'm receiving an error saying I don't have the rights to add roles to the VM. I tested this PR without that so I can't be certain everything still is all good. 

The disk are created and attached to the machine, but that's where my validation ends.

This PR is kind of work in progress, I think it's nice to start reviewing it while I dig deeper into validating it completely.


service/arm_templates/node_setup.json
```diff
@@ -179,22 +177,27 @@
...
+            "name": "[concat(parameters('clusterID'), '-', parameters('prefix'), '-', parameters('osDisk').name)]",
+            "caching": "[parameters('osDisk').caching]",
+            "managedDisk": {
+              "storageAccountType": "[parameters('osDisk').storageAccountType]"
             },
-            "caching": "ReadWrite",
             "createOption": "FromImage"
           },
-          "dataDisks": [
+          "copy": [
```
r7vme: what this "copy" means?

kopiczko: This is loop in ARM templates. Value of `name` is `dataDisks`, then you have # of copies in `count`, then `input `is the content. You can use `copyIndex(name)` to obtain the index of the current loop.

Cocotton:
@kopiczko cool, let me know when this will be figured out :)

service/arm_templates/node_setup.json
```diff
@@ -179,22 +177,27 @@
             "version": "[parameters('osImageVersion')]"
           },
           "osDisk": {
-            "name": "[variables('osDiskName')]",
-            "vhd": {
-              "uri": "[concat('http://', parameters('storageAccountName'), '.blob.core.windows.net/vhds/', variables('osDiskName'), '.vhd')]"
+            "name": "[concat(parameters('clusterID'), '-', parameters('prefix'), '-', parameters('osDisk').name)]",
+            "caching": "[parameters('osDisk').caching]",
```
kopiczko: What is caching?

service/arm_templates/node_setup.json
```diff
@@ -179,22 +177,27 @@
...
+              "input": {
+                "name": "[concat(parameters('clusterID'), '-', parameters('prefix'), '-', parameters('dataDisks')[copyIndex('dataDisks')].name)]",
+                "caching": "[parameters('dataDisks')[copyIndex('dataDisks')].caching]",
+                "diskSizeGB": "[parameters('dataDisks')[copyIndex('dataDisks')].sizeGB]",
+                "managedDisk": {
+                  "storageAccountType": "[parameters('dataDisks')[copyIndex('dataDisks')].storageAccountType]"
+                },
+                "createOption": "Empty",
+                "lun": "[add(copyIndex('dataDisks'), 1)]"
+              }
```
kopiczko: I did my homework. We don't want to expose that in the custom object at all. Let's create disks with Standard_LRS type, 50gb for data disk. All fixed in the template. Maybe extract to variables if makes sense.

Ping @r7vme

Cocotton: @kopiczko I had a chat with @r7vme last week and the solution seemed to add a docker disk on both the master and worker, as well as an etcd disk on the master.

If I hardcode that, the etcd disk will also be premium with a size of 50gb, which feels a bit overkill.

Should we rather merge the etcd and docker disk together? Or maybe there's a better solution?

kopiczko: @r7vme I think we can live with HDD for docker and SSD for etcd. WDYT?

Let's create as many disks as we need. Separate disk for etcd is legit. As well as separate disk for docker and data. What I mean is just not exposing that in the custom object (azuretpr) because we don't need that. This gives us freedom to change that any time.

r7vme: Looks like we should also take into account VM size

> Premium Storage supports DS-series, DSv2-series, GS-series, Ls-series, and Fs-series VMs. You can use standard and premium storage disks with these VM types. You cannot use premium storage disks with VM series that are not Premium Storage-compatible.

We can just replicate what acs-engine does: They select storage based on VM size [2].

For example, if user requested `Standard_A3` we can only use Standard storage. But with `Standard_DS1_v2` we can use both, but i would stick with Premium always if it's possible.

1 -  https://docs.microsoft.com/en-us/azure/virtual-machines/windows/premium-storage#supported-vms
2 - https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/azureconst.go#L525-L1049

kopiczko: I'd go with the Premium always and don't care at this point. As I said we don't have that in the custom object. We don't have to worry.

Cocotton: Good idea, I like that. @kopiczko I'll close the azuretpr since we don't need that anymore.

I'll keep you guys posted with updates to this PR.

Cocotton:
@kopiczko @r7vme I updated the PR with a new logic.

Instead of passing the osDisk and dataDisks object and array to the node_setup.json template, I simply pass an hardcoded dataDisks simplified array to the node template. The array only contains dataDisks with a name and size.

Then, from the [acs VM sizes list](https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/azureconst.go#L525-L1049), I made a list of all the VM only supporting Standard_LRS. I check if the vmSize parameter from the azuretpr is found in that list. If it is, I set the `storageAccountType` variable to "Standard_LRS". Otherwise, I set it to "Premium_LRS".

I chose to match the provided vmSize parameter on a list of standard machines, since there's less of them than premium machines.

Maybe we should do some validation somewhere to make sure we pass vmSize that exists? I could do that in the template, but it seems a bit deep to start validating stuff.

If also decided to match the vmSize on an array, rather than getting the [acs size map](https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/azureconst.go#L525-L1049) because it is so big and takes up so much space :P.

I tested the PR and the disks sets themselves to the right storageAccountType depending on the vmSize :)

service/arm_templates/node_setup.json
```diff
@@ -85,17 +79,104 @@
...
     "adminSSHKeyPath": "[concat('/home/', parameters('adminUsername'), '/.ssh/authorized_keys')]",
     "ipName": "[concat(parameters('clusterID'), '-', parameters('prefix'))]",
     "nicName": "[concat(parameters('clusterID'), '-', parameters('prefix'),'-NIC')]",
     "vmName": "[concat(parameters('clusterID'), '-', parameters('prefix'))]",
-    "osDiskName": "[concat(parameters('clusterID'), '-', parameters('prefix'), '-OSDisk')]",
-    "dataDiskName": "[concat(parameters('clusterID'), '-', parameters('prefix') ,'-DataDisk')]",
     "roleAssignmentName": "[concat(parameters('clusterID'), '-', parameters('prefix') ,'-RoleAssignment')]",
-    "contributorRoleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', parameters('contributorRoleDefinitionId'))]"
+    "contributorRoleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', parameters('contributorRoleDefinitionId'))]",
+    "storageAccountType": "[if(contains(parameters('standardLRSSize'), parameters('vmSize')), 'Standard_LRS', 'Premium_LRS')]"
```
r7vme: Nice :)

service/arm_templates/node_setup.json
```diff
@@ -85,17 +79,104 @@
...
+        "Standard_H8",
+        "Standard_H8m",
+        "Standard_NC12",
+        "Standard_NC24",
+        "Standard_NC24r",
+        "Standard_NC6",
+        "Standard_NV12",
+        "Standard_NV24",
+        "Standard_NV6"
+      ]
```
kopiczko: I wouldn't even make it a variable not  a parameter. I don't think there is a reason to expose that.

service/arm_templates/availability_sets_setup.json
```diff
@@ -35,8 +35,11 @@
       "name": "[variables('availabilitySetName')]",
       "location": "[resourceGroup().location]",
       "properties": {
-        "PlatformFaultDomainCount": "[parameters('faultDomainCount')]",
-        "PlatformUpdateDomainCount": "[parameters('updateDomainCount')]"
+        "platformFaultDomainCount": "[parameters('faultDomainCount')]",
+        "platformUpdateDomainCount": "[parameters('updateDomainCount')]"
+      },
+      "sku": {
+        "name": "aligned"
```
kopiczko: What is this?

service/arm_templates/node_setup.json
```diff
@@ -179,22 +260,27 @@
...
+              "input": {
+                "name": "[concat(parameters('clusterID'), '-', parameters('prefix'), '-', parameters('dataDisks')[copyIndex('dataDisks')].name)]",
+                "caching": "ReadWrite",
+                "diskSizeGB": "[parameters('dataDisks')[copyIndex('dataDisks')].sizeGB]",
+                "managedDisk": {
+                  "storageAccountType": "[variables('storageAccountType')]"
+                },
+                "createOption": "Empty",
+                "lun": "[copyIndex('dataDisks')]"
+              }
```
kopiczko: I think now we should add some mount service extensions to the cloudconfig, right?

r7vme: right, but let's do it later, when disks will be in place.

service/arm_templates/availability_sets_setup.json
Cocotton: This sets the availability set as managed instead of classic. A managed availability set is required for managed disks.

https://docs.microsoft.com/en-us/rest/api/compute/availabilitysets/availabilitysets-create#request

PR #72, opened 2017-12-12
service: use apiextensions

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1384

I wanted change to be minimal, so this leaves the azure operator in broken state (because of cloud-configs).

- switch to dep
- change k8scloudconfig to v_2_0_0 (this is subject to change to v_3_0_0 I just wanted a minimal diff)
- use apiextensions AzureConfig

TODO (in follow up PRs):
- use certs@master
- use v_3_0_0 cloud config
- update local guides

I'm sorry I switched to dep in one go, but other projects are using dep already. I tried to do that with glide, but it was too painful and this is the diff of two files.


service/operator/service.go
```diff
@@ -15,9 +17,11 @@ import (
 type Config struct {
 	// Dependencies.
 
-	Logger    micrologger.Logger
-	K8sClient kubernetes.Interface
-	Resources []framework.Resource
+	Logger       micrologger.Logger
```
xh3b4sd: The order broke here.

kopiczko: Done.

PR #77, opened 2017-12-21
cloudconfig: refactor extensions rendering

kopiczko:
If you prefer looking at diff: https://github.com/giantswarm/azure-operator/pull/77/commits/4541b3c654b5fb4007feea349996fd279993bd93

Second commit is only moving stuff around.

I looked at diff and decided it's I'll do moving in one go.


service/cloudconfig/master_extension.go
```diff
@@ -0,0 +1,100 @@
+package cloudconfig
+
+import (
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/key"
+	"github.com/giantswarm/certs"
+	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v_2_0_0"
+	"github.com/giantswarm/microerror"
+)
```
kopiczko: `Files` and `Units` look the same. They call `render*File` and `render*Unit` and return slice of results.

All `render*` are unified. They call corresponding `render*` from `render.go`.

service/cloudconfig/render.go
```diff
@@ -0,0 +1,71 @@
...
+
+	file := k8scloudconfig.FileAsset{
+		Metadata: fileMeta,
+		Content:  content,
+	}
+
+	return file, nil
+}
+
+// TODO remove obj from parameters. This doesn't need params to render.
```
kopiczko: Will be resolved in follow up PR. I find that small, but important fix.

service/cloudconfig/render.go
```diff
@@ -0,0 +1,71 @@
+package cloudconfig
+
+import (
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v_2_0_0"
+	"github.com/giantswarm/microerror"
+)
```
kopiczko: All unified. Takes params and renders content.

service/cloudconfig/types.go
```diff
@@ -3,27 +3,41 @@ package cloudconfig
 import (
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/key"
 )
```
kopiczko: Here I added `*Params` struct for each asset.

service/cloudconfig/worker_extension.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (we *workerExtension) renderGetKeyVaultSecretsFile() (k8scloudconfig.FileAsset, error) {
+	params := getKeyVaultSecretsFileParams{
+		VaultName: key.KeyVaultName(we.CustomObject),
+		Secrets:   []getKeyVaultSecretsFileParamsSecret{},
+	}
+
+	// Only file paths are needed here, so we don't care if certs.Cluster
+	// is empty.
+	for _, f := range certs.NewFilesClusterWorker(certs.Cluster{}) {
```
kopiczko: workerExtension differs only by this line from masterExtension.

service/cloudconfig/master_extension.go
```diff
@@ -0,0 +1,100 @@
...
+	asset, err := renderGetKeyVaultSecretsFile(params)
+	if err != nil {
+		return k8scloudconfig.FileAsset{}, microerror.Mask(err)
+	}
+
+	return asset, nil
+}
+
+func (me *masterExtension) renderGetKeyVaultSecretsUnit() (k8scloudconfig.UnitAsset, error) {
+	params := me.CustomObject
```
jgsqware: why this assignation instead of passing directly me.CustomObject?

service/cloudconfig/render.go
```diff
@@ -0,0 +1,71 @@
...
+
+func renderGetKeyVaultSecretsFile(params getKeyVaultSecretsFileParams) (k8scloudconfig.FileAsset, error) {
+	fileMeta := k8scloudconfig.FileMetadata{
+		AssetContent: getKeyVaultSecretsFileTemplate,
+		Path:         getKeyVaultSecretsFileName,
+		Owner:        getKeyVaultSecretsFileOwner,
+		Permissions:  getKeyVaultSecretsFilePermission,
+	}
+
+	content, err := k8scloudconfig.RenderAssetContent(fileMeta.AssetContent, params)
```
jgsqware: line 38,47 could be extracted eg:

```
func renderFiles(f k8scloudconfig.FileMetadata, p interface{}) (k8scloudconfig.FileAsset,erroe)
```
 and reused in all render function

kopiczko: I would rather improve FileAsset and FileMeta instead.

service/cloudconfig/master_extension.go
kopiczko: To make it similar to other `render*` functions. This will be removed anyway.

service/cloudconfig/render.go
jgsqware: in a separate PR?

kopiczko: This is part of k8scloudconfig. I don't promise I'll do that immediately. Also feel free to make a follow up PR if you find it important. I feel like adding `renderFiles` to every operator means the library should be improved.

PR #82, opened 2017-12-27
Add rbac and sa

pipo02mix:
Towards [#2296](https://github.com/giantswarm/giantswarm/issues/2296)

Although we don't have yet an environment with services running for Azure, I have left this PR open to track and remember the RBAC implementation. For now, I have just added namespaces and azureconfigs

### DIFF
--- a/helm/azure-operator-chart/templates/rbac.yaml
+++ b/helm/azure-operator-chart/templates/rbac.yaml
@@ -1,8 +1,8 @@
-  apiVersion: rbac.authorization.k8s.io/v1beta1
-  kind: ClusterRole
-  metadata:
-    name: azure-operator
-  rules:
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: azure-operator
+rules:
   - apiGroups:
       - apiextensions.k8s.io
     resources:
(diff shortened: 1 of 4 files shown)


helm/azure-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - namespaces
+    verbs:
+      - get
+      - create
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: userd
```
r7vme: userd -> azure-operator

pipo02mix: Good catch!

helm/azure-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - patch
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+    verbs:
+      - get
+      - create
+      - delete
```
kopiczko: Do we need namespaces at all?

kopiczko: I didn't mean namespace in `metadata:` JFYI. I think it was good, the deployment is in the giantswarm namespace.

pipo02mix: I removed namespace from metadata because it is a cluster role and it does not need. About the resource namespace if the azure operator does not need it at all then I remove it.

pipo02mix: indeed if azure operator won't use namespaces then we can do a role binding instead of cluster role binding for the cluster role, so it would be more restricted

corest:
We will need also `PodSecurityPolicy` - currently `azure-operator` fails because of this

pipo02mix:
working on it, thanks for the help

pipo02mix:
@corest I have added PSP too. PTAL

PR #88, opened 2018-01-02
dep: update k8scloudconfig

kopiczko:
Also adds **insecure** encryption key just to satisfy newest master of k8scloudconfigs.


kopiczko:
@puja108 to make my testing easier. When we integrate it fully with host cluster there will be a secret with randomized key for an installation. The operator should use the value from the secret. 

There is `Use proper apiserver encryption key with use of randomekeys.` todo in https://github.com/giantswarm/giantswarm/issues/1384 to track that.

PR #89, opened 2018-01-02
cloudconfig: extract loopbackAddr const

kopiczko:
This is to enable clean removal in the future. When removing const and
there is still some loopback address left, the code will not compile.

### DIFF
--- a/service/cloudconfig/service.go
+++ b/service/cloudconfig/service.go
@@ -21,7 +21,7 @@ const (
 	// TODO: use randomkeys to fill the value properly.
 	apiserverEncryptionKey = "UShZb2zOWvY5Svkf8+oSSa0dEZxprPWz0xYYsAsFuP0="
 
-	// loopbackAddr is used to set varioius cloud config template
+	// loopbackAddr is used to set various cloud config template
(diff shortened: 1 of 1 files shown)


service/cloudconfig/service.go
```diff
@@ -15,11 +15,15 @@ import (
 )
 
 const (
-	// apiserverEncryptionKey is **insecure** encryption key just to satisfy newest
-	// master of k8scloudconfigs.
+	// apiserverEncryptionKey is **insecure** encryption key just to
+	// satisfy newest master of k8scloudconfigs.
```
rossf7: <3

service/cloudconfig/service.go
```diff
@@ -15,11 +15,15 @@ import (
...
 	//
 	// TODO: use randomkeys to fill the value properly.
 	apiserverEncryptionKey = "UShZb2zOWvY5Svkf8+oSSa0dEZxprPWz0xYYsAsFuP0="
+
+	// loopbackAddr is used to set varioius cloud config template
```
rossf7: * various

PR #92, opened 2018-01-04
Network imporovements for Calico

r7vme:
- allow ip forwarding
- add security rules to allow pods/calico traffic
- fix ingress load balancer ports

TL;DR allow pods subnet in security groups

For context:
In Azure Calico uses native Azure routing. So pods will (should) use ip range from virtual network range.

For example
virtual network: 10.0.0.0/16
calico/pods network: 10.0.128.0/17 <-  my recommendation to just pick up second half of virt.network range. Every Kubernetes node will pick /24 from this subnet.

### DIFF
--- a/service/arm_templates/security_groups_setup.json
+++ b/service/arm_templates/security_groups_setup.json
@@ -12,6 +12,9 @@
     "clusterID": {
       "type": "string"
     },
+    "calicoSubnetCidr": {
+      "type": "string"
+    },
     "masterSubnetCidr": {
       "type": "string"
     },
@@ -54,7 +57,7 @@
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "*",
-              "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "destinationAddressPrefix": "*",
               "access": "Deny",
               "direction": "Inbound",
               "priority": "4096"
@@ -67,7 +70,7 @@
               "protocol": "*",
               "sourcePortRange": "*",
               "destinationPortRange": "*",
-              "sourceAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "sourceAddressPrefix": "*",
               "destinationAddressPrefix": "*",
               "access": "Allow",
               "direction": "Outbound",
@@ -82,7 +85,7 @@
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "[parameters('masterSubnetCidr')]",
-              "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "destinationAddressPrefix": "[parameters('virtualNetworkCidr')]",
               "access": "Allow",
               "direction": "Inbound",
               "priority": "4094"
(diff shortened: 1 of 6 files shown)


service/arm_templates/security_groups_setup.json
```diff
@@ -54,7 +57,7 @@
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "*",
-              "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "destinationAddressPrefix": "*",
```
r7vme: We should deny all traffic, not only one for our subnet (i.e. it does not deny pods traffic).

service/arm_templates/security_groups_setup.json
```diff
@@ -67,27 +70,13 @@
               "protocol": "*",
               "sourcePortRange": "*",
               "destinationPortRange": "*",
-              "sourceAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "sourceAddressPrefix": "*",
```
r7vme: Simplify here. Allow all outbound traffic.

Alternative would be allow VMs and Pods subnets and deny anything else. But i don't really care about outbound.

service/arm_templates/security_groups_setup.json
```diff
@@ -67,27 +70,13 @@
...
               "destinationAddressPrefix": "*",
               "access": "Allow",
               "direction": "Outbound",
               "priority": "4095"
             }
           },
-          {
```
r7vme: this rule does nothing as internal traffic (within subnet) will not hit security groups.

service/arm_templates/security_groups_setup.json
```diff
@@ -117,18 +106,32 @@
...
-            "name": "workerSubnetToMasterSubnet",
+            "name": "allowWorkerSubnet",
             "properties": {
               "description": "Allow the worker machines to reach the master machines on any ports.",
               "protocol": "*",
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "[parameters('workerSubnetCidr')]",
-              "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "destinationAddressPrefix": "[parameters('virtualNetworkCidr')]",
```
r7vme: Destination can be VMs subnet or Pods subnet (always within virtual network range). This makes sure that pods from workers can access master.

Alternatively we can add additional rule specifically for Pods. But i prefer to keep rules list as slim as possible. Azure says that many rules affect network performance.

service/arm_templates/security_groups_setup.json
```diff
@@ -117,18 +106,32 @@
...
               "access": "Allow",
               "direction": "Inbound",
               "priority": "3800"
             }
+          },
+          {
+            "name": "allowCalicoSubnet",
```
r7vme: Allows pods traffic from workers.

service/arm_templates/security_groups_setup.json
```diff
@@ -148,7 +151,7 @@
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "*",
-              "destinationAddressPrefix": "[parameters('workerSubnetCidr')]",
+              "destinationAddressPrefix": "*",
```
r7vme: all changes for worker security group reflect changes for master. See comments above

service/arm_templates/security_groups_setup.json
```diff
@@ -82,7 +85,7 @@
               "sourcePortRange": "*",
               "destinationPortRange": "*",
               "sourceAddressPrefix": "[parameters('masterSubnetCidr')]",
-              "destinationAddressPrefix": "[parameters('masterSubnetCidr')]",
+              "destinationAddressPrefix": "[parameters('virtualNetworkCidr')]",
```
r7vme: pods also live inside this network, so allow whole virtual network as destination (pods network always part of virtual network).

PR #94, opened 2018-01-07
dep,service: add version bundle

kopiczko:


service/version_bundles.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			Changelogs: []versionbundle.Changelog{},
+			Components: []versionbundle.Component{
+				{
+					Name:    "calico",
+					Version: "3.0.1",
+				},
+				{
+					Name:    "docker",
+					Version: "1.12.6",
```
kopiczko: Docker version is just a copy/paste. Is there an easy way to check this?

xh3b4sd: It comes from the CoreOS version. See https://coreos.com/releases, I am just not sure which CoreOS version is being used right now.

kopiczko: It would be maybe nice to add the info about the os as well?

xh3b4sd: We could do that. It is only that we show the users what we define here and we would maybe need to filter certain things. Don't know. Anyway, when we add it in Azure, we should do so for AWS and KVM as well.

kopiczko: So here in the version bundle we can't be really sure what the OS (and in consequence etcd) version is. In case of azure it depends on settings in the custom object. This is the same case with AWS AMI. Not sure how it's done for KVM.

My opinion is it should be baked into the bundle. I.e. not configurable with custom object or whatsoever.

cc @r7vme as this may be interesting for you.

xh3b4sd: > My opinion is it should be baked into the bundle. I.e. not configurable with custom object or whatsoever.

+1

rossf7: +1 from me too. We did that with the kubernetes version and k8scloudconfig and I think was the right call.

kopiczko: So this version is correct now as the CoreOS version used is 1465.7.0 (kind of stale - from September last year). But the operator doesn't have any control over that.

New coreos versions have  17.09.0 version of docker. Versioning approach changed?

I'm leaving it how it is. This is still wip.

kopiczko:
Is there something I should do to receive LGTM?

service/version_bundles.go
```diff
@@ -0,0 +1,47 @@
...
+					Name:    "kubernetes",
+					Version: "1.9.0",
+				},
+				{
+					Name:    "nginx-ingress-controller",
+					Version: "0.9.0",
+				},
+			},
+			Dependencies: []versionbundle.Dependency{},
+			Deprecated:   true,
```
xh3b4sd: Deprecated must not be true in case WIP is true. You should actually not be able to start the operator using this version bundle.

service/version_bundles.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"time"
+
+	"github.com/giantswarm/versionbundle"
+)
+
+func newVersionBundles() []versionbundle.Bundle {
+	return []versionbundle.Bundle{
+		{
+			Changelogs: []versionbundle.Changelog{},
```
xh3b4sd: Changelog must not be empty. Same with starting the operator should fail.

service/version_bundles.go
r7vme: yes, docker switched to new versioning scheme, so 17.09 is fine.

regarding this change, CoreOS version is taken from [CRO](https://github.com/giantswarm/azure-operator/blob/master/examples/azure-resource-lab-chart/templates/cluster.yaml#L85). In aws-operator it statically set in installations repo.

no more comments.

xh3b4sd:
> taking into account that we don't have proper version bundles

For my interest, what is wrong with them? How could they be more proper?

r7vme:
> For my interest, what is wrong with them? How could they be more proper?

who knows. 

I've left some comments https://github.com/giantswarm/giantswarm/issues/2335

service/version_bundles.go
kopiczko: Done.

service/version_bundles.go
kopiczko: Done.

service/version_bundles.go
```diff
@@ -0,0 +1,78 @@
...
+					Component:   "calico",
+					Description: "remove calico-ipip-pinger",
+					Kind:        versionbundle.KindRemoved,
+				},
+				{
+					Component:   "calico",
+					Description: "remove calico-node-controller",
+					Kind:        versionbundle.KindRemoved,
+				},
+			},
```
kopiczko: I followed k8scloudconfig changelog: https://github.com/giantswarm/k8scloudconfig/blob/52cd0230d6fb7346b7b33b75e74e8e12990def21/CHANGELOG.md

And http://keepachangelog.com/en/1.0.0/ format

service/version_bundles.go
```diff
@@ -0,0 +1,78 @@
...
+					Kind:        versionbundle.KindChanged,
+				},
+				{
+					Component:   "calico",
+					Description: "remove calico-ipip-pinger",
+					Kind:        versionbundle.KindRemoved,
+				},
+				{
+					Component:   "calico",
+					Description: "remove calico-node-controller",
```
xh3b4sd: The description was defined as a sentence so far. I would like to keep it that way. See for instance https://github.com/giantswarm/kvm-operator/blob/master/service/version_bundles.go.

kopiczko: Would you also like to use past form?
I'd like to to keep imperative form so we follow "standard", but I can live with either.

xh3b4sd: I actually don't mind. I only want to have one convention that we use everywhere. I started with just something. When there is something better lets agree on it and get it straight everywhere. AKA business as usual.

xh3b4sd: @marians probably has a strong opinion on this.

r7vme: everything that we have here is related to k8scc, so  imo "source of thruth" should be there.

Can we move this piece to k8scc?

kopiczko: I'd be fine to keep `[]versionbundle.Changelog` & `[]versionbundle.Component` instance for every version in k8scloudconfig repo. But we'd have to merge it with operator specific stuff. Like OS and docker version. But there is `append` :)

I'd like to hear @xh3b4sd opinion on that.

xh3b4sd: Doesn't make sense to me. It is good to have it hard coded in the operator. That makes us thing. We always have to think about updates and its implications. We would also only redirect for a portion of the changelogs, but not for all. Then everything would be different. The consistency we have right now is way more valuable IMO.

kopiczko: Portion of changelogs is an argument for me.
IMO consistency is the main problem here. For example v3.0.0 now switched to coredns. It's very easy to forget about updating version_bundles.go on next `dep ensure -update` run.

xh3b4sd: When updating vendor deps will ever be an issue it means somebody did not do his job right, which can be a problem for everything. When you add something to k8scloudconfig you have to make sure it lands in the operators accordingly. When we version everything and somebody adds something new then nothing changes for the running guest clusters when updating vendor deps so this is not an issue at all. When you screw the process though and update existing templates and you do not fix this in the operators and modify the version bundles accordingly then you fucked up greatly. Nobody should do this anyway. Something like this should not pass any review.

marians: Regarding the description text: As we will present this to end users without any change, please use these rules:

- past tense
- Uppercase first letter
- Punctuation only required if you use more than one sentence.

This would turn `remove calico-node-controller` into `Removed calico-node-controller`

kopiczko:
Going with what we have now. Let's solve the rest in further discussions.

marians:
I was a second too late, it seems. So I copy my comment for further reference.

Regarding the description text: As we will present this to end users without any change, please use these rules:

- past tense
- Uppercase first letter
- Punctuation only required if you use more than one sentence.

This would turn `remove calico-node-controller` into `Removed calico-node-controller`.

PR #95, opened 2018-01-10
deployment: bake-in OS version

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2333.

This is step towards encapsulating OS version into the version bundle.


service/resource/deployment/types.go
```diff
@@ -0,0 +1,45 @@
...
+	Offer string `json:"offer" yaml:"offer"`
+	// Publisher is the image publisher (e.g GiantSwarm).
+	Publisher string `json:"publisher" yaml:"publisher"`
+	// SKU is the image SKU (e.g. Alpha).
+	SKU string `json:"sku" yaml:"sku"`
+	// Version is the image version (e.g. 1465.7.0).
+	Version string `json:"version" yaml:"version"`
+}
+
+func newNodeOSImage() nodeOSImage {
```
xh3b4sd: I would like to have either a good godoc on this because of the hard coded information within the function or constants for the magic strings which are documented properly.

kopiczko: I'm afraid it will be difficult to provide better documentation. Even azure official doc is pretty generic: https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines#ImageReference

I will include this link in the godoc though.

xh3b4sd: The important thing for me here is to have the used CoreOS version in the godoc.

PR #96, opened 2018-01-10
service: use past tense in version bundle changelog

kopiczko:


service/version_bundles.go
```diff
@@ -12,32 +12,32 @@ func newVersionBundles() []versionbundle.Bundle {
...
 				{
 					Component:   "kubernetes",
-					Description: "Update to version 1.9.0.",
+					Description: "Updated to version 1.9.0.",
 					Kind:        versionbundle.KindChanged,
 				},
 				{
 					Component:   "kubernetes",
-					Description: "Use vanilla (previously coreos) hyperkube image.",
+					Description: "Used vanilla (previously coreos) hyperkube image.",
```
marians: For my English, this should actually be `Use vanilla`, as it refers more to the effect of the change. I know, it's not what I wrote in those simple guidelines. Seems like it's not so easy to make guidelines for writing English sentences.

kopiczko: Everything refers to the effect of the change. I don't understand. You maybe wanted to make it noun like `Use of`, but let's please avoid mixing-in nouns here. 

Another links to convince you to the imperative form (beyond simple guidelines):
https://stackoverflow.com/questions/3580013/should-i-use-past-or-present-tense-in-git-commit-messages
https://coreos.com/releases/#1576.5.0

marians: New suggestion in past rense:

`Changed to use vanilla (previously coreos) hyperkube image`

`Switched to vanilla (previously coreos) hyperkube image`

PR #97, opened 2018-01-11
deployment: reuse SSH keys from Cluster

kopiczko:


service/resource/deployment/types.go
```diff
@@ -39,7 +26,7 @@ type nodeOSImage struct {
 }
 
 // newNodeOSImage provides OS information CoreOS 1465.7.0.
-func newNodeOSImage() nodeOSImage {
+func newNodeOSImageCoreOS_1465_7_0() nodeOSImage {
```
kopiczko: I renamed that to be more explicit during usage.

service/resource/deployment/types.go
```diff
@@ -1,27 +1,14 @@
...
-func newNodes(azureNodes []providerv1alpha1.AzureConfigSpecAzureNode) []node {
-	var ns []node
-	for _, n := range azureNodes {
-		ns = append(ns, newNode(n))
-	}
-	return ns
-}
-
 type node struct {
-	providerv1alpha1.AzureConfigSpecAzureNode
```
kopiczko: No embedding. Only VMSize field will be left in AzureConfigSpecAzureNode, so it's just copied.

service/resource/deployment/deployment.go
```diff
@@ -19,49 +19,71 @@ func getDeploymentNames() []string {
 	}
 }
 
-func (r Resource) newMainDeployment(cluster providerv1alpha1.AzureConfig) (Deployment, error) {
-	certs, err := r.certsSearcher.SearchCluster(key.ClusterID(cluster))
+func (r Resource) newMainDeployment(obj providerv1alpha1.AzureConfig) (Deployment, error) {
```
kopiczko: Renamed to obj. `cluster` was weird. In general I favour `obj` over `customOjbect`. It's shorter and we use so often that we can reserve that name excessively for that.

service/resource/deployment/deployment.go
```diff
@@ -19,49 +19,71 @@ func getDeploymentNames() []string {
...
 
 	// Convert certs files into a collection of key vault secrets.
 	certSecrets := convertCertsToSecrets(certs)
 
-	masterCloudConfig, err := r.cloudConfig.NewMasterCloudConfig(cluster)
+	var masterNodes []node
+	for _, m := range obj.Spec.Azure.Masters {
+		n := node{
+			AdminUsername:   key.AdminUsername(obj),
+			AdminSSHKeyData: key.AdminSSHKeyData(obj),
```
kopiczko: This data is now taken from `obj.Spec.Cluster.Kubernetes.SSH.UserList[0]`.

service/key/key.go
```diff
@@ -24,6 +24,24 @@ func AzureCloudType(customObject providerv1alpha1.AzureConfig) string {
 	return defaultAzureCloudType
 }
 
+func AdminUsername(customObject providerv1alpha1.AzureConfig) string {
+	users := customObject.Spec.Cluster.Kubernetes.SSH.UserList
+	// We don't want panics when someone is doing something nasty.
+	if len(users) == 0 {
+		return ""
+	}
+	return users[0].Name
```
xh3b4sd: Why is admin the first?

service/resource/deployment/deployment.go
xh3b4sd: FYI I saw in the k8s community the working seems to be `CRD` and `CR`.

kopiczko:
@r7vme yes, Azure will fail. We should fail in general before if there isn't any SSH user provided in the Cluster object. I just did basic protection for panics. Then something will fail before or after and we'll know. This is conscious and I'm not concerned at all.

service/key/key.go
kopiczko: Because there is no particular preference who should be that one. And there is an requirement that someone has to be. First one is the most natural choice for me in this case.

service/resource/deployment/deployment.go
kopiczko: Not sure if you aren't confusing things here. My understanding is:

- `CRD` is a resource (/customresourcedefinitions endpoint)
- `CR` is an object of the CRD resource (this object is reconciled by the `CRD` controller and new resource is created with new endpoint, e.g. /azureconifgs); the object can be obtained with `kubectl get customresourcedefinitions cr_name`.
- `obj` is an object of the new resource (/azureconfigs endpoint), created by reconciling the `CR`; the object can be obtained with `kubectl get azureconfigs obj_name`.

xh3b4sd: For me `obj` and `CR` are the same here.

PR #99, opened 2018-01-11
service: update to Container Linux 1576.5.0

kopiczko:


service/version_bundles.go
```diff
@@ -42,13 +52,21 @@ func newVersionBundles() []versionbundle.Bundle {
 				},
 			},
 			Components: []versionbundle.Component{
+				{
+					Name:    "containerlinux",
```
kopiczko: Not sure if that should be the name.

service/resource/deployment/deployment.go
```diff
@@ -44,7 +44,7 @@ func (r Resource) newMainDeployment(obj providerv1alpha1.AzureConfig) (Deploymen
 		n := node{
 			AdminUsername:   key.AdminUsername(obj),
 			AdminSSHKeyData: key.AdminSSHKeyData(obj),
-			OSImage:         newNodeOSImageCoreOS_1465_7_0(),
+			OSImage:         newNodeOSImageCoreOS_1576_5_0(),
```
r7vme: what is the purpose of version bundle if you're still hard coding CoreOS version like this?

service/resource/deployment/types.go
```diff
@@ -25,12 +25,12 @@ type nodeOSImage struct {
 	Version string `json:"version" yaml:"version"`
 }
 
-// newNodeOSImage provides OS information CoreOS 1465.7.0.
-func newNodeOSImageCoreOS_1465_7_0() nodeOSImage {
+// newNodeOSImage provides OS information CoreOS 1576.5.0.
+func newNodeOSImageCoreOS_1576_5_0() nodeOSImage {
```
r7vme: For me this function shoud take as argument channel and version and output proper struct. But now it's hardcoded smth.

service/version_bundles.go
xh3b4sd: Towards the naming I would ask @marians and @puja108 since this is customer facing. 

I think it is good to have the changelogs but I am not sure we should add the components now. You actually only want to update the version. The changelog is there. This is good. When we add CoreOS as component here we should also do this in the other operators. I think we should maybe just not do it now and keep it to the changelogs. When we later have the cloud config operator or whatever we can add the component there.

service/version_bundles.go
```diff
@@ -42,13 +52,21 @@ func newVersionBundles() []versionbundle.Bundle {
...
+					Version: "1576.5.0",
+				},
 				{
 					Name:    "calico",
 					Version: "3.0.1",
 				},
+				{
+					Name:    "containerlinux",
+					Version: "1576.5.0",
+				},
```
xh3b4sd: This is now duplicated.

service/version_bundles.go
```diff
@@ -30,6 +30,16 @@ func newVersionBundles() []versionbundle.Bundle {
 					Description: "Updated to version 3.0.1.",
 					Kind:        versionbundle.KindChanged,
 				},
+				{
```
r7vme: no comments, only thought in my mind: "data" and "code" should live SEPARATELY.

service/version_bundles.go
r7vme: name is fine

xh3b4sd: I also saw that there is a docker component which the other operators do not have AFAIK. We should align this then.

service/version_bundles.go
kopiczko: I forgot to remove the previous one.

service/version_bundles.go
kopiczko: They do: https://github.com/giantswarm/kvm-operator/blob/8f35c674414d9098aa415d2d5774c01da08b44c3/service/version_bundles.go#L49-L52

This is why I started the discussion about docker version in version bundles.
@calvix says this is important to show docker version to the customer.

service/resource/deployment/types.go
kopiczko: I think this answers this comment: https://github.com/giantswarm/azure-operator/pull/99#discussion_r160970150

service/version_bundles.go
xh3b4sd: Cool.

calvix: I think I meant its not important for the customer :)

kopiczko:
@r7vme 

> Do we want to move changelog outside of operator as we discussed yesterday?

I'm fine with that and think it's a good thing. @xh3b4sd said he can disagree and commit.
I'm occupied with other stuff ATM and it isn't trivial to do it nicely. If you have time please go ahead.

service/version_bundles.go
xh3b4sd: lol

service/version_bundles.go
kopiczko: I agree with you in general, but TBH there is no code. This is pure data file if you analize it. We could name it `version_bundles.json`, read and parse to VersionBundle struct, but there is no reason to do so. You can think about it as "Go Object Notation" format, like "JavaScript Object Notation" format, aka. JSON.

service/version_bundles.go
kopiczko: Done.

kopiczko:
I think we can decide whether we need to expose containerlinux and/or docker component later on. I think it's important information though. I can imagine it may be used when specifying dependencies between version bundles.

We can change that down the road.

I don't want to change the canon in this PR. If we don't want to make OS part of the version bundle this is a good place to discuss: https://github.com/giantswarm/giantswarm/issues/2333

Re changelogs and version bundles in general, I don't want introduce real changes in this PR as it isn't related.

What should I do to get a LGTM?

puja108:
I kinda like that we have the coreos version in the operator here, I'd like to have this similarly on KVM and AWS, where we currently keep it in the installations config (which I don't like, cause it creates differences). In AWS we can keep a list of AMIs so the operator can choose based on the region it's deployed on and we don't need to configure it by hand.

kopiczko:
My idea in AWS is to keep a map of OSVersion&Region pairs to AMI. That way it would be sufficient to specify version bundle and region in the custom object.

puja108:
the map sounds like a good idea, but in the custom object I'd like to only have the OSVersion then and the operator should know in which region it (or the guest cluster) is. Or does the current custom object already contain the region? The only use I see for region in the custom object is if we manage guest clusters in different regions by an operator in another region.

kopiczko:
What can I do to get LGTM?

PR #100, opened 2018-01-11
service: update to Container Linux 1576.5.0

kopiczko:


service/resource/deployment/types.go
```diff
@@ -34,3 +48,14 @@ func newNodeOSImageCoreOS_1465_7_0() nodeOSImage {
...
+// keyVault is used to pass secrets to Key Vault as a secure object.
+type keyVault struct {
+	Secrets []keyVaultSecret `json:"secrets"`
+}
+
+// keyVaultSecret is a secret stored in Key Vault.
+type keyVaultSecret struct {
+	SecretName  string `json:"secretName"`
+	SecretValue string `json:"secretValue"`
+}
```
kopiczko: Fields didn't changed. Stuff is moved and unexported.

kopiczko:
- I renamed `keyVaultSecrets` to `keyVault` because it has field `Secrets` of type `keyVaultSecret` and I wanted to somehow conform to the practice we have in apiextensions.

- I unexported `Deployment` type to make it clear it isn't used outside the package. I think this is good practice in general. Besides that all other types are private.

PR #109, opened 2018-01-23
Updates vendor with k8s 1.9.2

corest:


kopiczko:
Stale?

kopiczko:
Correct.

kopiczko:
I think you also want to change imorts inside cloudconfig packge.

corest:
I thought about this, but then moved those changes into separate branch, because usually we are not mixing vendoring with version bundle changes

kopiczko:
Yeah. Feel free to proceed with that. You don't need approval for pure vendor updates.

Issue #110, opened 2018-01-23
Set an active release.


oponder:
There is currently no active versionbundle for the azure-operator.

This means non-admin users will not be able to create a cluster, and Happa is not able to suggest what Release Version a user should pick by default.

Is the current versionbundle done? If so we should change the WIP flag to false.

ping @kopiczko @r7vme

r7vme:
@kopiczko i think we need this ASAP. can you take a look?

oponder:
@r7vme we will need this at least before letting a customer attempt to use it.

Until then it's ok like this, since we are not comfortable with breaking the process.

I made a workaround so that at least we can create clusters using happa as an admin for now. That workaround is in progress, but you can see it here:

https://github.com/giantswarm/happa/issues/282

kopiczko:
Can we close that? We'll have the release when it's needed.

kopiczko:
Ping @oponder

oponder:
Ok, we can close it. Let's not forget though ok!

PR #111, opened 2018-01-24
Create an active release

oponder:
Fixes: https://github.com/giantswarm/azure-operator/issues/110

There is currently no active release for azure, since this component does not have a non-wip versionbundle.

This means customers won't be able to create clusters.

Is this ready to be set to WIP: false ?

### DIFF
--- a/service/version_bundles.go
+++ b/service/version_bundles.go
@@ -72,7 +72,7 @@ func newVersionBundles() []versionbundle.Bundle {
 			Name:         "azure-operator",
 			Time:         time.Date(2018, time.January, 7, 8, 35, 0, 0, time.UTC),
 			Version:      "0.1.0",
-			WIP:          true,
+			WIP:          false,
 		},
 	}
 }


xh3b4sd:
I would like to keep the process up. Having WIP version bundles is fine and we can work on inactive releases via the API when we use our giantswarm org member users or admin tokens.

kopiczko:
Does it mean it should remain as it is?

xh3b4sd:
IMO yes. Do we have customers demanding Azure guest clusters? If this is true we should come to a state and freeze a version bundle so customers can use an active release. Then we can continue developing on the new version bundles. I just don't know what the motivation here is and in which state Azure is or what we want to give out now.

kopiczko:
I think it's for happa testing, but I can't speak for @oponder.
BTW you can always deploy this branch for your testing.

oponder:
I'll find a different way around this in Happa.

PR #121, opened 2018-02-08
Encryption key

TheoBrigitte:
Towards #118


examples/azure-resource-lab-chart/templates/cluster.yaml
```diff
@@ -1,5 +1,5 @@
-apiVersion: "giantswarm.io/v1"
-kind: AzureCluster
+apiVersion: "provider.giantswarm.io/v1alpha1"
+kind: AzureConfig
```
rossf7: OK in this case as its small. But in general upgrading the chart from TPR to CRD would be better as a separate PR since its a separate change.

service/operator/framework.go
```diff
@@ -42,6 +43,18 @@ func newFramework(config Config) (*framework.Framework, error) {
...
 
+	var randomkeysSearcher *randomkeys.Searcher
+	{
+		c := randomkeys.DefaultConfig()
+		c.K8sClient = config.K8sClient
+		c.Logger = config.Logger
+
+		randomkeysSearcher, err = randomkeys.NewSearcher(c)
+		if err != nil {
+			return nil, microerror.Maskf(err, "randomkeys.NewSearcher")
```
rossf7: Here you can just mask the error. This is so we have the full call stack when it gets logged.

service/resource/deployment/deployment.go
```diff
@@ -50,12 +50,17 @@ func (r Resource) newMainDeployment(obj providerv1alpha1.AzureConfig) (deploymen
...
 		return deployment{}, microerror.Mask(err)
 	}
 
-	workerCloudConfig, err := r.cloudConfig.NewWorkerCloudConfig(obj)
+	masterCloudConfig, err := r.cloudConfig.NewMasterCloudConfig(obj, string(cluster.APIServerEncryptionKey))
+	if err != nil {
+		return deployment{}, microerror.Mask(err)
+	}
+
+	workerCloudConfig, err := r.cloudConfig.NewWorkerCloudConfig(obj, string(cluster.APIServerEncryptionKey))
```
rossf7: The encryption key is only needed by the master. This is because its set via a flag on the api-server.

kopiczko: Yes, it was my broken thinking back then and it sneaked. It would be nice to fix that in scope of that change.

service/operator/framework.go
TheoBrigitte: Isn't `microerror.Maskf` already doing that by providing both the call stack and an annotation?

service/resource/deployment/deployment.go
```diff
@@ -50,7 +50,12 @@ func (r Resource) newMainDeployment(obj providerv1alpha1.AzureConfig) (deploymen
 		workerNodes = append(workerNodes, n)
 	}
 
-	masterCloudConfig, err := r.cloudConfig.NewMasterCloudConfig(obj)
+	cluster, err := r.randomkeysSearcher.SearchCluster(key.ClusterID(obj))
+	if err != nil {
+		return deployment{}, microerror.Mask(err)
+	}
```
kopiczko: Wouldn't it be nice to encapsulate that knowledge in `cloudconfig` package w/o interface changes?

TheoBrigitte: yeah make sense, so we would not even touch the signature of `NewMasterCloudConfig`

kopiczko: Yes, that would be the goal.

service/operator/framework.go
rossf7: It still works but AFAIK the annotation won't be logged. We usually use `Maskf` when creating an error. If you're returning it you can just use `Mask`.

kopiczko: Site node: annotation will be logged.

rossf7: Good to know, thanks!

service/cloudconfig/service.go
```diff
@@ -167,6 +173,14 @@ func (c CloudConfig) NewWorkerCloudConfig(customObject providerv1alpha1.AzureCon
 	return newCloudConfig(k8scloudconfig.WorkerTemplate, params)
 }
 
+func (c CloudConfig) getEncryptionkey(customObject providerv1alpha1.AzureConfig) (string, error) {
+	cluster, err := c.randomkeysSearcher.SearchCluster(key.ClusterID(customObject))
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+	return string(cluster.APIServerEncryptionKey), nil
+}
```
kopiczko: I don't see a reason to extract that method, but when it feels cleaner to you that's ok.

TheoBrigitte:
Tested and working on gollum.

encryption key (base64 encoded twice) for the guest cluster on gollum
```
$ k --context=giantswarm-gollum get secret io5vg-encryption -o yaml
apiVersion: v1
data:
  encryption: S0NmTXpBQktkNjdXWlRrNHRaeVRvejlGQ3dtbVdnOEVKZnhlcFNsY3J3ST0=
[...]
```

encryption key (base64 encoded once) on the guest cluster
```
io5vg-Master-1 ~ # cat /etc/kubernetes/encryption/k8s-encryption-config.yaml 
kind: EncryptionConfig
apiVersion: v1
resources:
    - aescbc:
        keys:
        - name: key1
          secret: KCfMzABKd67WZTk4tZyToz9FCwmmWg8EJfxepSlcrwI=
   [...]
```

created a test secret on guest cluster
```
$ k --context=giantswarm-io5vg create secret generic secret1 -n default --from-literal=mykey=mydata
secret "secret1" created
$ k --context=giantswarm-io5vg get secret secret1 -o yaml
apiVersion: v1
data:
  mykey: bXlkYXRh
kind: Secret
[...]

io5vg-Master-1 ~ # ETCDCTL_API=3 etcdctl --cacert=/etc/kubernetes/ssl/etcd/client-ca.pem --cert=/etc/kubernetes/ssl/etcd/client-crt.pem --key=/etc/kubernetes/ssl/etcd/client-key.pem get /giantswarm.io/secrets/default/secret1
/giantswarm.io/secrets/default/secret1
k8s:enc:aescbc:v1:key1:[...]
```

PR #122, opened 2018-02-12
Update k8scloudconfig from 3.0.0 to 3.1.1

corest:


corest:
Sure

corest:
I see nodes in Azure, but can't connect to machines, API is also not reachable. Not sure if that is related to this change, or operator in general

kopiczko:
What's the cluster ID?

corest:
`mew47`

kopiczko:
Looks like your PR introduces braking changes. @TheoBrigitte tired to create a cluster with your branch and it wasn't working. Route Table (azure resource) is empty, that means something is wrong with the cloud provider.

When reverted to latest master operator creates working clusters.

kopiczko:
If you want to ssh to your cluster to debug further I can help you with that tomorrow.

corest:
It can be related to the self-hosted k8s components

corest:
For now, cloudconfig 3.1.0 can't be used with azure-operator. New bundle will be released only after `3.1.1` release  (required change https://github.com/giantswarm/k8scloudconfig/pull/305)

service/cloudconfig/service.go
```diff
@@ -100,20 +100,37 @@ func (c CloudConfig) NewMasterCloudConfig(customObject providerv1alpha1.AzureCon
...
+							ReadOnly: true,
+						},
+						k8scloudconfig.HostMount{
+							Name:     "identity-settings",
+							Path:     "/var/lib/waagent/",
+							ReadOnly: true,
+						},
 					},
 					CommandExtraArgs: []string{
 						"--cloud-config=/etc/kubernetes/config/azure.yaml",
```
corest: We are using `azure.yaml` here, but don't have a mount. Is that correct?

r7vme: it is in k8scloudconfig already.

https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_0_0/master_template.go#L2210

corest:
Tested on gollum, works fine

PR #123, opened 2018-02-13
key: fix KeyVault name

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2491

Prefixing KeyVault with cluster ID is risky because of naming
constraints. KeyVault name must not start with a number. The name has to
be globally unique.

I decided to prefix it (and eventually all other resources) with gs- to
guarantee we always start with a letter and avoid such situations in the
future.

Follow up issue https://github.com/giantswarm/giantswarm/issues/2535

### DIFF
--- a/service/key/key.go
+++ b/service/key/key.go
@@ -9,7 +9,14 @@ import (
 )
 
 const (
-	defaultAzureCloudType     = "AZUREPUBLICCLOUD"
+	defaultAzureCloudType = "AZUREPUBLICCLOUD"
+
+	// globalPrefix is added to all resources created by azure-operator.
+	// This is to avoid problems witch resources which names has to start
+	// with a letter, as sometimes guest cluster IDs start with number.
+	globalPrefix = "gs"
+
+	keyVaultSuffix            = "keyvault"
 	routeTableSuffix          = "RouteTable"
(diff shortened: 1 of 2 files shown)


service/key/key_test.go
```diff
@@ -58,22 +58,6 @@ func Test_ClusterCustomer(t *testing.T) {
...
-			Cluster: providerv1alpha1.Cluster{
-				ID: "test-cluster",
-			},
-		},
-	}
-
-	if KeyVaultName(customObject) != expectedVaultName {
-		t.Fatalf("Expected key vault name %s but was %s", expectedVaultName, KeyVaultName(customObject))
-	}
-}
```
kopiczko: The test is moved to the table test below.

kopiczko:
Tested locally.

service/key/key.go
```diff
@@ -9,7 +9,14 @@ import (
 )
 
 const (
-	defaultAzureCloudType     = "AZUREPUBLICCLOUD"
+	defaultAzureCloudType = "AZUREPUBLICCLOUD"
+
+	// globalPrefix is added to all resources created by azure-operator.
+	// This is to avoid problems witch resources which names has to start
```
tuommaki: Small typo: "problems **witch** resources" -> "problems **with** resources" I believe?

PR #127, opened 2018-02-19
Remove loopback reference to kubernetes API

corest:
As this limitation (https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-troubleshoot#cause-3-accessing-the-load-balancer-from-the-same-vm-and-network-interface) is not valid anymore - we can use the same approach for accessing API as in other operators. 

Do we want to remove the possibility of overriding this in k8scloudconfig? E.g. here:
https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_1_1/cloudconfig.go#L39-L44

### DIFF
--- a/service/cloudconfig/service.go
+++ b/service/cloudconfig/service.go
@@ -16,12 +16,6 @@ import (
 	"github.com/giantswarm/randomkeys"
 )
 
-const (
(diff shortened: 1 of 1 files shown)


corest:
I removed also BindAddress. As i Understood - it was the part of a workaround

PR #129, opened 2018-02-22
service: pass certificates with cloudconfig

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2483

This removes KeyVault completely. Certificates are passed with cloud configs.
This is important step towards switching to scale sets, as I had difficulties to obtain principal ID from them.

As a bonus clusters boots faster, and we have simpler arm templates and cloudconfigs.


kopiczko:
Tested locally.

PR #132, opened 2018-02-26
azureconfig/v1: prepare for operatorkit update

kopiczko:


kopiczko:
Tested locally.

kopiczko:
As title says it's an preparation. Of course many things aren't prefect.

PR #133, opened 2018-02-27
azureconfig/v1: remove etcd DNS

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2490 https://github.com/giantswarm/giantswarm/issues/2483

On Azure only master nodes use etcd, so it is locked down for
simplification and security.


corest:
Did you discover it only now or there were some limitations before?

kopiczko:
@corest the idea comes from this comment https://github.com/giantswarm/giantswarm/issues/2483#issuecomment-368447760 (yesterday).

PR #134, opened 2018-02-27
azureconfig: add resource router

kopiczko:


service/azureconfig/v1/resource_set.go
```diff
@@ -140,11 +143,39 @@ func NewResourceSet(config ResourceSetConfig) ([]framework.Resource, error) {
...
+	resources := []framework.Resource{
 		resourceGroupResource,
 		deploymentResource,
 		dnsrecordResource,
 	}
 
+	handlesFunc := func(obj interface{}) bool {
+		customObject, err := key.ToCustomObject(obj)
+		if err != nil {
+			config.Logger.Log("level", "warning", "message", fmt.Sprintf("invalid object: %s", err), "stack", fmt.Sprintf("%v", err))
```
xh3b4sd: As soon as we have multiple versions this log will be more or less useless IMO. It is also that the framework emits an error log in case there is no single resource set for an observed custom object.

kopiczko: This log is about wrong type assertion. It should never happen. But when it happens we should notice it. I don't see relation between number or resource sets and wrong assertions, can you explain?

xh3b4sd: Right, the number of resource sets is irrelevant here. I still think this log is useless. I assume it never will be a thing because we have the informer that only gives us `AWSConfig` types. Do you have a concrete case in which this here is necessary? I feel like I still miss something.

kopiczko: I can't give an example. This is a defensive programming. If we are 100% this will never return error (which it shouldn't I agree) we should either `panic` or `customObject, _ := key.ToCustomObject(obj)`. But I'm not 100% thus log.

xh3b4sd: Fair. Lets see how this goes.

PR #136, opened 2018-03-06
dep: add prune settings

kopiczko:


xh3b4sd:
A description would help me understand what exactly you did and why.

xh3b4sd:
We should make everybody aware of updating its dep then.

Issue #115, opened 2018-01-29
No peering with host cluster


r7vme:
At the moment guest clusters does not have network peering with host cluster, which is required for SSH and monitoring.

cc: @kopiczko

xh3b4sd:
What is with this issue? IIRC Theo is working on it. Can we move this to WIP then?

TheoBrigitte:
This issue is about providing network peering between host and guest cluster on azure, so we can reach into guest cluster from the host (e.g. ssh into guest master node).

And yes I am working on this.

xh3b4sd:
Sweet. Thank you for checking back.

r7vme:
Done as per @TheoBrigitte

PR #145, opened 2018-03-13
host <-> guest cluster peering

TheoBrigitte:
Towards https://github.com/giantswarm/azure-operator/issues/115

Provide virtual network peering between host and guest cluster.


service/azureconfig/v1/arm_templates/main.json
```diff
@@ -8,12 +8,21 @@
...
+    },
     "GiantSwarmTags": {
       "type": "object",
       "defaultValue": {
         "provider": "F80D01C0-7AAC-4440-98F6-5061511962AD"
       }
     },
+    "virtualNetworkName": {
+      "type": "string"
+    },
```
rossf7: Why is there a description on `hostClusterResourceGroup` but not on `virtualNetworkName`?

service/azureconfig/v1/arm_templates/peering.json
```diff
@@ -0,0 +1,41 @@
...
+    "sourceNetwork": {
+      "type": "string"
+    },
+    "targetNetwork": {
+      "type": "string"
+    },
+    "resourceGroup": {
+      "type": "string"
+    },
+    "apiVerison": {
```
rossf7: apiVersion?

service/azureconfig/v1/resource/vnetpeering/delete.go
```diff
@@ -0,0 +1,66 @@
+package vnetpeering
+
+import (
+	"context"
+	// "fmt"
```
rossf7: Can this be removed?

service/azureconfig/v1/resource/vnetpeering/delete.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/azure-operator/service/azureconfig/v1/key"
```
rossf7: Imports should be in 3 blocks.

service/azureconfig/v1/resource/vnetpeering/desired.go
```diff
@@ -0,0 +1,39 @@
+package vnetpeering
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+	"github.com/Azure/go-autorest/autorest/to"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/azure-operator/service/azureconfig/v1/key"
```
rossf7: Same here.

Gopkg.toml
```diff
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/giantswarm/apiextensions"
-  branch = "master"
+  branch = "azure-cluster-peering"
```
kopiczko: Reminder: revert to master

client/azure.go
```diff
@@ -78,6 +79,8 @@ type AzureClientSet struct {
 	DNSRecordSetsClient *dns.RecordSetsClient
 	// DNSRecordSetsClient manages DNS zones.
 	DNSZonesClient *dns.ZonesClient
+	// VirtualNetworkPeeringsClient manages virtual network peerings.
```
JosephSalisbury: This Godoc variable name should match

xh3b4sd: Same one above.

service/azureconfig/v1/resource/vnetpeering/delete.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, microerror.Maskf(err, "NewDeletePatch")
+	}
+
+	return r.newDeletePatch(ctx, a, c, d)
+}
+
+func (r Resource) newDeletePatch(ctx context.Context, azureConfig providerv1alpha1.AzureConfig, current, desired network.VirtualNetworkPeering) (*framework.Patch, error) {
+	patch := framework.NewPatch()
+	patch.SetDeleteChange(current)
```
xh3b4sd: Why always only deleting current and not desired?

service/azureconfig/v1/resource/vnetpeering/delete.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return microerror.Maskf(err, "deleting host vnet peering")
+	}
+
+	_, err = vnetPeeringClient.Delete(ctx, key.HostClusterResourceGroup(azureConfig), key.HostClusterResourceGroup(azureConfig), key.ResourceGroupName(azureConfig))
+	if err != nil {
+		return microerror.Maskf(err, "deleting host vnet peering")
+	}
+
+	r.logger.LogCtx(ctx, "debug", "deleting host vnet peering: deleted")
```
xh3b4sd: There is a message scheme established which I would like to stick with. This `deleting: deleted` thingy is not really helpful on a daily basis.

service/azureconfig/v1/resource/vnetpeering/desired.go
```diff
@@ -0,0 +1,39 @@
...
+			RemoteVirtualNetwork: &network.SubResource{
+				ID: to.StringPtr(r.getVirtualNetworkID(key.ResourceGroupName(azureConfig), key.VnetName(azureConfig))),
+			},
+		},
+	}, nil
+}
+
+func (r Resource) getVirtualNetworkID(resourceGroupName, vnetName string) string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s", r.azureConfig.SubscriptionID, resourceGroupName, vnetName)
+}
```
xh3b4sd: This looks to me like it should be in the key package.

service/azureconfig/v1/resource/vnetpeering/error.go
```diff
@@ -0,0 +1,18 @@
+package vnetpeering
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
+var (
+	invalidConfigError = microerror.New("invalid config")
+	wrongTypeError     = microerror.New("wrong type")
+)
```
xh3b4sd: This is not how we do it usually. I would like to stick to our style. It makes code easier to scan when patterns stay the same.

service/azureconfig/v1/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return r.newUpdatePatch(ctx, a, c, d)
+}
+
+func (r Resource) newUpdatePatch(ctx context.Context, azureConfig providerv1alpha1.AzureConfig, current, desired network.VirtualNetworkPeering) (*framework.Patch, error) {
+	patch := framework.NewPatch()
+
+	if needUpdate(current, desired) {
+		patch.SetUpdateChange(desired)
+	}
```
xh3b4sd: This condition does not belong here. The private `new*Patch` methods only gather information and forward them. Conditionals and any logic should be implemented in some other private method. At least this is how we are doing it so far and I would like to stick to consistency.

service/azureconfig/v1/resource_set.go
```diff
@@ -163,10 +164,30 @@ func NewResourceSet(config ResourceSetConfig) (*framework.ResourceSet, error) {
...
 	}
 
+	var vnetPeeringResource framework.Resource
+	{
+		ops, err := vnetpeering.New(
+			vnetpeering.Config{
+				Logger: config.Logger,
+
+				AzureConfig: config.AzureConfig,
+			},
```
xh3b4sd: The config is usually separated.

service/azureconfig/v1/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return patch, nil
+}
+
+func needUpdate(current, desired network.VirtualNetworkPeering) bool {
```
xh3b4sd: I would like to see tests for this.

service/azureconfig/v1/resource/vnetpeering/current.go
```diff
@@ -0,0 +1,35 @@
...
+func (r Resource) GetCurrentState(ctx context.Context, azureConfig interface{}) (interface{}, error) {
+	a, err := key.ToCustomObject(azureConfig)
+	if err != nil {
+		return nil, microerror.Maskf(err, "GetCurrentState")
+	}
+
+	return r.getCurrentState(ctx, a)
+}
+
+func (r Resource) getCurrentState(ctx context.Context, azureConfig providerv1alpha1.AzureConfig) (interface{}, error) {
```
kopiczko: Idea of having corresponding private method is to pass strong type parameter and return strong typed values. Here you return an interface `interface{}`. It should be `network.VirtualNetworkPeering{}`. There is aleadyt mess with nil values, which returning strong typed values prevents.

service/azureconfig/v1/resource/vnetpeering/desired.go
kopiczko: I don't think this has to be a separate function, you can create this string inside `getDesiredState`. Your preference though.

service/azureconfig/v1/arm_templates/main.json
kopiczko: virtualNetworkName is for guest cluster.

client/azure.go
```diff
@@ -174,6 +183,18 @@ func newDNSZonesClient(config AzureConfig) (*dns.ZonesClient, error) {
 	return &client, nil
 }
 
+func newVnetPeeringClient(config AzureConfig) (*network.VirtualNetworkPeeringsClient, error) {
```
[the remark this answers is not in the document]
TheoBrigitte: I would rather boyscout this in a different PR, since there is more to be renamed here

client/azure.go
```diff
@@ -174,6 +183,18 @@ func newDNSZonesClient(config AzureConfig) (*dns.ZonesClient, error) {
...
+	spt, err := newServicePrincipalToken(config, azure.PublicCloud.ResourceManagerEndpoint)
+	if err != nil {
+		return nil, microerror.Maskf(err, "creating service principal token")
```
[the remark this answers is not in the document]
TheoBrigitte: I wanted to follow the already present structure here. But I also experienced harder debugging with those annotations, usually the traces speak by itself.

Would it make sense to you @kopiczko to removes those annotations in a different PR ?

service/azureconfig/v1/arm_templates/main.json
```diff
@@ -179,8 +193,8 @@
           "contentVersion": "1.0.0.0"
         },
         "parameters": {
-          "clusterID": {"value": "[parameters('clusterID')]"},
           "GiantSwarmTags": {"value": "[parameters('GiantSwarmTags')]"},
+          "virtualNetworkName": {"value": "[parameters('virtualNetworkName')]"},
```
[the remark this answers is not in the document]
TheoBrigitte: This is for consistency with host cluster peering, the value is no longer generated within the template but rather from `key.VnetName()`.
So we can then pass the same network name taken here (for guest cluster) and to host cluster peering later on.

service/azureconfig/v1/resource/vnetpeering/delete.go
TheoBrigitte: Deleting current ensure there is something to be deleted, since it has te be retrieved from azure first. Deleting desired does not make sense to me since it is always there.

I should probably also check for emptiness of current here, in case it is already gone.

service/azureconfig/v1/resource/vnetpeering/delete.go
TheoBrigitte: could you point me to this message scheme ?

client/azure.go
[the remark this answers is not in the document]
xh3b4sd: Fine with me.

service/azureconfig/v1/resource/vnetpeering/delete.go
xh3b4sd: I see. My thinking is often that `GetCurrentState` has higher chances of being wrong for several reasons whereas `GetDesiredState` is often more static and not that error prone in comparison. So my thinking is usually considering the desired state to be deleted and be fine in case it is not found at all, which we have to deal with the other way around anyway.

service/azureconfig/v1/resource/vnetpeering/delete.go
xh3b4sd: https://github.com/giantswarm/aws-operator/blob/master/service/awsconfig/v9/resource/cloudformation/create.go

client/azure.go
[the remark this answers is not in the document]
kopiczko: It's actually @xh3b4sd request. For me those annotations make sense but for me either approach isn't a blocker. I think for errors that we don't create it makes sense to add additional information in case they are cryptic.

service/azureconfig/v1/arm_templates/main.json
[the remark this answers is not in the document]
kopiczko: Got it. Thanks.

service/azureconfig/v1/resource/vnetpeering/delete.go
kopiczko: This message scheme is used in azure-operator in all resources and I like it a lot. Makes debugging things easy as messages are visually aligned.

service/azureconfig/v1/resource/vnetpeering/current.go
```diff
@@ -0,0 +1,35 @@
...
+	vnetPeeringClient, err := r.getVnetPeeringClient()
+	if err != nil {
+		return nil, microerror.Maskf(err, "getCurrentState")
+	}
+
+	vnetPeering, err := vnetPeeringClient.Get(ctx, key.HostClusterResourceGroup(azureConfig), key.HostClusterResourceGroup(azureConfig), key.ResourceGroupName(azureConfig))
+	if client.ResponseWasNotFound(vnetPeering.Response) {
+		return network.VirtualNetworkPeering{}, nil
```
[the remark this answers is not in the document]
TheoBrigitte: I do not really understand your point here

service/azureconfig/v1/resource/vnetpeering/delete.go
TheoBrigitte: Makes actual sense, I will update that to desired and handle the not found case later on in `ApplyDeleteChange`

service/azureconfig/v1/resource/vnetpeering/delete.go
TheoBrigitte: I will keep this to be consistent with what is already in azure-operator

client/azure.go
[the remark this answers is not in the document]
xh3b4sd: Just sharing my opinion here.

service/azureconfig/v1/resource/vnetpeering/delete.go
xh3b4sd: Disagree and commit.

service/azureconfig/v1/resource/vnetpeering/current.go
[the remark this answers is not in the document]
kopiczko: In one place you return zero-value of a struct (which isn't nil) in other zero value of the struct.

Addressing https://github.com/giantswarm/azure-operator/pull/145#discussion_r174393635 will fix that.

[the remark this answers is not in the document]
xh3b4sd: `network.VirtualNetworkPeering` is a type. `network.VirtualNetworkPeering{}` initializes a value. `&network.VirtualNetworkPeering{}` initializes a pointer. Zero value for `network.VirtualNetworkPeering` is `network.VirtualNetworkPeering{}`. Zero value for `&network.VirtualNetworkPeering{}` is `nil`. `network.VirtualNetworkPeering{}` is not the same as `nil`. The return values within a resource implementation should always be the same. So either a resource deals with pointers and is able to return `nil` in error cases or a resource deals with values and has to return the initializes zero value which cannot be `nil`. In the code here values and `nil` return values are mixed. `nil` implies there are pointers. There are no pointers. We should not mix things up.

service/azureconfig/v1/resource/vnetpeering/current.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// GetCurrentState retrieve the current host cluster virtual network peering resource from azure.
+func (r Resource) GetCurrentState(ctx context.Context, azureConfig interface{}) (interface{}, error) {
+	a, err := key.ToCustomObject(azureConfig)
+	if err != nil {
+		return nil, microerror.Maskf(err, "GetCurrentState")
+	}
+
+	return r.getCurrentState(ctx, a)
```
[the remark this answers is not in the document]
TheoBrigitte: Good point

service/azureconfig/v1/arm_templates/peering.json
TheoBrigitte: actually I renamed this to `virtualNetworksPeeringAPIVersion` and provided a default value and description

TheoBrigitte:
This is quite lot of changes, sorry I should have splitted that.

service/azureconfig/v1/resource/vnetpeering/update_test.go
```diff
@@ -0,0 +1,140 @@
...
+		if c.isError {
+			if err == nil {
+				t.Errorf("expected error got '%t'", err)
+			}
+		} else {
+			if result != c.expected {
+				t.Errorf("expected '%t' got '%t'", c.expected, result)
+			}
+		}
+	}
```
kopiczko: It would be good to create new tests on par with https://github.com/giantswarm/fmt/blob/master/go/general_style_guide.md#unit-tests

service/azureconfig/v1/resource/vnetpeering/desired.go
```diff
@@ -0,0 +1,40 @@
...
+// Peering resource is named after cluster's resource group and targeting its virtual network.
+func (r Resource) GetDesiredState(ctx context.Context, azureConfig interface{}) (interface{}, error) {
+	a, err := key.ToCustomObject(azureConfig)
+	if err != nil {
+		return nil, microerror.Maskf(err, "GetDesiredState")
+	}
+
+	vnetPeering, err := r.getDesiredState(ctx, a)
+	if err != nil {
+		return nil, microerror.Maskf(err, "GetDesiredState")
```
kopiczko: Same about `nil` returning should be `network.VirtualNetworkPeering{}`

service/azureconfig/v1/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,142 @@
...
+//     VirtualNetworkPeeringPropertiesFormat.AllowVirtualNetworkAccess
+//     VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork.ID
+func needUpdate(current, desired network.VirtualNetworkPeering) (bool, error) {
+	if desired.Name == nil ||
+		desired.VirtualNetworkPeeringPropertiesFormat == nil ||
+		desired.VirtualNetworkPeeringPropertiesFormat.AllowVirtualNetworkAccess == nil ||
+		desired.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork == nil ||
+		desired.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork.ID == nil {
+		return false, microerror.Maskf(invalidDesiredState, "got %+#v", desired)
+	}
```
TheoBrigitte: This block make sense from a semantic point of view but on the other side should not be required since `desired` is mostly static and should always be there.

service/azureconfig/v1/resource/vnetpeering/update_test.go
```diff
@@ -0,0 +1,173 @@
...
+				VirtualNetworkPeeringPropertiesFormat: &network.VirtualNetworkPeeringPropertiesFormat{
+					AllowVirtualNetworkAccess: to.BoolPtr(true),
+					RemoteVirtualNetwork: &network.SubResource{
+						ID: to.StringPtr("some ID"),
+					},
+				},
+			},
+			true,
+			nil,
+		},
```
TheoBrigitte: Those test should be enough but I could also add cases for nil pointers

service/azureconfig/v1/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,143 @@
...
+		return true, nil
+	}
+
+	if current.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork == nil ||
+		current.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork.ID == nil ||
+		*current.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork.ID != *desired.VirtualNetworkPeeringPropertiesFormat.RemoteVirtualNetwork.ID {
+		return true, nil
+	}
+
+	if current.VirtualNetworkPeeringPropertiesFormat.PeeringState == network.Disconnected {
```
TheoBrigitte: Handle case where peering is in `disconnected` state and need to be updated in order to be in `initiated` state and ready for a new connection.

TheoBrigitte:
Need one approval please

service/azureconfig/v1/resource/vnetpeering/update.go
kopiczko: I'd just remove it.

PR #146, opened 2018-03-14
Align resource tag names and use key package

rossf7:


rossf7:
Not possible due to Azure naming restrictions. 

`Tag names can't contain these characters: <, >, %, &, \, ?, /`

PR #147, opened 2018-03-14
Resource tags

rossf7:
Towards giantswarm/giantswarm#2775

Adds the tags below to each Resource Group for cost tracking.

```
GiantSwarmCluster: "rue77"
GiantSwarmOrganization: "giantswarm"
GiantSwarmInstallation: "gollum"
```

Replaces the previous tags.

```
ClusterID: "rue77"
CustomerID: "giantswarm"
```

### DIFF
--- a/flag/service/installation/installation.go
+++ b/flag/service/installation/installation.go
@@ -1,0 +1,5 @@
+package installation
+
+type Installation struct {
+	Name string
+}
--- a/service/azureconfig/v1/key/key.go
+++ b/service/azureconfig/v1/key/key.go
@@ -11,6 +11,7 @@ const (
 	defaultAzureCloudType = "AZUREPUBLICCLOUD"
 
 	clusterTagName      = "GiantSwarmCluster"
+	installationTagName = "GiantSwarmInstallation"
 	organizationTagName = "GiantSwarmOrganization"
 
 	routeTableSuffix          = "RouteTable"
@@ -62,9 +63,10 @@ func ClusterOrganization(customObject providerv1alpha1.AzureConfig) string {
 }
 
 // ClusterTags returns a map with the resource tags for this cluster.
-func ClusterTags(customObject providerv1alpha1.AzureConfig) map[string]string {
+func ClusterTags(customObject providerv1alpha1.AzureConfig, installationName string) map[string]string {
 	tags := map[string]string{
 		clusterTagName:      ClusterID(customObject),
+		installationTagName: installationName,
 		organizationTagName: ClusterOrganization(customObject),
 	}
 
--- a/service/azureconfig/v1/key/key.go
+++ b/service/azureconfig/v1/key/key.go
@@ -10,8 +10,8 @@ import (
 const (
 	defaultAzureCloudType = "AZUREPUBLICCLOUD"
 
-	clusterTagName      = "giantswarm.io/cluster"
-	organizationTagName = "giantswarm.io/organization"
+	clusterTagName      = "GiantSwarmCluster"
+	organizationTagName = "GiantSwarmOrganization"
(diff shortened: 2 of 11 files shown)


flag/service/installation/installation.go
```diff
@@ -0,0 +1,5 @@
+package installation
+
+type Installation struct {
+	Name string
+}
```
rossf7: Installation name is set via a flag.

service/azureconfig/v1/key/key.go
```diff
@@ -10,6 +10,10 @@ import (
 const (
 	defaultAzureCloudType = "AZUREPUBLICCLOUD"
 
+	clusterTagName      = "GiantSwarmCluster"
+	installationTagName = "GiantSwarmInstallation"
+	organizationTagName = "GiantSwarmOrganization"
```
rossf7: These are the tag names. We can't use the `giantswarm.io/cluster` format we use in AWS because `/` is a restricted character.

> Tag names can't contain these characters: <, >, %, &, \, ?, /

service/azureconfig/v1/key/key.go
```diff
@@ -52,6 +56,23 @@ func ClusterID(customObject providerv1alpha1.AzureConfig) string {
...
+// ClusterTags returns a map with the resource tags for this cluster.
+func ClusterTags(customObject providerv1alpha1.AzureConfig, installationName string) map[string]string {
+	tags := map[string]string{
+		clusterTagName:      ClusterID(customObject),
+		installationTagName: installationName,
+		organizationTagName: ClusterOrganization(customObject),
+	}
+
+	return tags
+}
```
rossf7: This key function creates the tags.

service/azureconfig/v1/resource/resourcegroup/resource.go
```diff
@@ -92,14 +96,10 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 
-	tags := map[string]string{
-		clusterIDTag:  key.ClusterID(customObject),
-		customerIDTag: key.ClusterCustomer(customObject),
-	}
 	resourceGroup := Group{
 		Name:     key.ClusterID(customObject),
 		Location: key.Location(customObject),
-		Tags:     tags,
+		Tags:     key.ClusterTags(customObject, r.installationName),
```
rossf7: Here we set the tag for the Resource Group. We have a separate resource group per guest cluster.

service/azureconfig/v1/resource/resourcegroup/resource.go
```diff
@@ -19,34 +19,38 @@ const (
...
 	deleteTimeout = 30 * time.Minute
 	managedBy     = "azure-operator"
 )
 
 type Config struct {
-	AzureConfig client.AzureConfig
-	Logger      micrologger.Logger
+	AzureConfig      client.AzureConfig
+	InstallationName string
+	Logger           micrologger.Logger
```
kopiczko: Nitpick InstallationName is more of a setting than dep.

service/azureconfig/v1/key/key.go
kopiczko: Yes, that's sad :(
I like the choice you made given the restrictions.

service/azureconfig/v1/resource/resourcegroup/resource.go
rossf7: Good catch, done.

rossf7:
Yes, I think we'll need to tag at least some of resources for enabling the Azure cloud provider. This should help with cost tracking and I wanted to get it in before we have customer clusters.

Issue #12, opened 2017-08-30
Add healthz endpoint that checks Azure API access


rossf7:
The operator should have a healthz endpoint that checks the Azure API credentials are valid. If possible it should also validate that the operator has the correct API permissions.

The endpoint should be created in https://github.com/giantswarm/microendpoint so it can be reused.

- [ ] add check
- [ ] enable checks in helm charts

xh3b4sd:
Hey guys. What about this issue? Sounds fairly simple. Lets pick it or rip it.

rossf7:
Let's add the `healthz`. I'll take this week as boyscouting.

PR #148, opened 2018-03-15
Add healthz

rossf7:
Fixes giantswarm/azure-operator#12

Adds the healthz endpoint and service. The check just gets the first resource group in the subscription to check the credentials.


service/healthz/service.go
```diff
@@ -0,0 +1,61 @@
...
+	s := &Service{
+		azureConfig: config.AzureConfig,
+		logger:      config.Logger,
+	}
+
+	return s, nil
+}
+
+// Check implements the health check which lists the first Resource Group
+// in the subscriptiontion to check we can authenticate.
```
rossf7: Here is the check. It's pretty simple but should check the credentials without causing a lot of load.

server/endpoint/healthz/endpoint.go
```diff
@@ -0,0 +1,94 @@
+package healthz
```
xh3b4sd: All this should not be necessary. Check https://github.com/giantswarm/kvm-operator/blob/ebab431ef35478a4a7efc6830564a3e47f38e6a9/server/endpoint/endpoint.go.

service/healthz/service.go
xh3b4sd: `subscriptiontion` -> `subscription`

server/middleware/middleware.go
```diff
@@ -1,33 +0,0 @@
...
-}
-
-// New creates a new configured middleware.
-func New(config Config) (*Middleware, error) {
-	return &Middleware{}, nil
-}
-
-// Middleware is middleware collection.
-type Middleware struct {
-}
```
rossf7: I removed the middleware as well as the endpoint. As both are not used.

service/healthz/service.go
rossf7: I can never spell that. Thx!

server/endpoint/healthz/endpoint.go
rossf7: Ack. I based this on aws-operator which has the old structure. I'll update once this is merged.

server/middleware/middleware.go
xh3b4sd: I think this will throw an error on daemon start because so far the middleware is expected to not be empty.

service/healthz/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return response, nil
+}
+
+// Check implements the health check which lists the first Resource Group
+// in the subscription to check we can authenticate.
+func (s *Service) Check(ctx context.Context, request Request) (*Response, error) {
+
+	return DefaultResponse(), nil
+}
```
xh3b4sd: This should be removed now I guess. It does not really do anything.

rossf7: Ack. Will get rid of the dead code.

PR #157, opened 2018-03-22
take host cluster and location from flags

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2849

First step is to make sure we know host cluster CIDR globally.


Gopkg.lock
```diff
@@ -94,7 +94,7 @@
   version = "v1.0.0"
 
 [[projects]]
-  branch = "master"
+  branch = "azureconfig-rm-host"
```
kopiczko: TODO: revert

Gopkg.toml
```diff
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/giantswarm/apiextensions"
-  branch = "master"
+  branch = "azureconfig-rm-host"
```
kopiczko: TODO: revert

client/azure.go
```diff
@@ -31,28 +28,11 @@ type AzureConfig struct {
...
-		// Dependencies.
-		Logger: nil,
-
-		// Settings.
-		ClientID:       "",
-		ClientSecret:   "",
-		SubscriptionID: "",
-		TenantID:       "",
-	}
-}
```
kopiczko: Boyscouted in the meantime.

service/azureconfig/v1/key/types.go
```diff
@@ -0,0 +1,39 @@
+package key
+
+import "fmt"
+
+type Azure struct {
+	HostCluster AzureHostCluster
+	Location    string
+}
```
kopiczko: For a lack of better idea I added a pure data type, which provides Validate method, so it's reused in `New`. I'm going to include information required to create a new client as well. So we don't pass client.AzureConfig around.

service/azureconfig/v1/key/key.go
```diff
@@ -156,11 +142,6 @@ func WorkerSubnetName(customObject providerv1alpha1.AzureConfig) string {
 	return fmt.Sprintf("%s-%s-%s", ClusterID(customObject), virtualNetworkSuffix, workerSubnetSuffix)
 }
 
-// Location returns the physical location where the Resource Group is deployed.
-func Location(customObject providerv1alpha1.AzureConfig) string {
-	return customObject.Spec.Azure.Location
-}
-
```
kopiczko: Now those values are taken from flags. They aren't really configurable per guest cluster.

service/azureconfig/framework.go
```diff
@@ -40,26 +42,29 @@ func NewFramework(config FrameworkConfig) (*framework.Framework, error) {
 	var err error
 
 	if config.G8sClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.G8sClient must not be empty")
+		return nil, microerror.Maskf(invalidConfigError, "%T.G8sClient must not be empty", config)
```
rossf7: <3

service/azureconfig/v1/key/types.go
rossf7: I like the approach and the use of the Validate method.

xh3b4sd: I would have created a new package for this. This is not about the things the key package is usually concerned about.

service/azureconfig/v1/resource/resourcegroup/resource.go
```diff
@@ -24,33 +24,43 @@ const (
 )
 
 type Config struct {
+	Logger micrologger.Logger
+
 	AzureConfig      client.AzureConfig
-	Logger           micrologger.Logger
+	Azure            key.Azure
```
xh3b4sd: Azure comes before AzureConfig in the alphabet. Also, what is the difference between both? I think the naming is a bit unfortunate. One is a client and one is installation/host-cluster configuration as I see it.

service/azureconfig/v1/key/types.go
kopiczko: Bad thing about this place is that `v1/key` is imported in the main service. This thing must be resource version independent like the CR is. I would put that into `azureconfig/types.go`, but that will create cyclic dependency. I'm going to create a `azureconfig/config/types.go` and put it there.

Thoughts?

service/azureconfig/v1/resource/resourcegroup/resource.go
kopiczko: Yes check https://github.com/giantswarm/azure-operator/pull/157#discussion_r176372168 I'm going to remove AzureConfig in general. Just not in this PR.

service/azureconfig/v1/key/types.go
xh3b4sd: Why not keeping it in this repo in `service/config/` or something?

kopiczko: I meant this repo. `/service/azureconfig/config`

service/service.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/giantswarm/azure-operator/client"
 	"github.com/giantswarm/azure-operator/flag"
 	"github.com/giantswarm/azure-operator/service/azureconfig"
+	azureconfigconfig "github.com/giantswarm/azure-operator/service/azureconfig/config"
```
kopiczko: Now this is kind of annoying that it has to be aliased because `config` is a `New` parameter.

Any ideas for a more unique name? `configuration`, `settings`?

xh3b4sd: What about `flags`? It actually maps information from the command line flags to the service.

kopiczko: I thought about it as well but there is also `"github.com/giantswarm/azure-operator/flag"` import which would collide... Unless we make one plural and other singular, but I think we both agree that's awful.

I think `settings` makes sense as this is previous `// Settings.` part of the `Config`s.

xh3b4sd: I do not mind. Don't have better ideas either.

kopiczko: Also `settings` or `setting`?

kopiczko:
Tested on godsmack.

PR #161, opened 2018-03-28
add helm/azure-resource-chart

TheoBrigitte:
Towards https://github.com/giantswarm/giantswarm/issues/2441


helm/azure-resource-chart/values.yaml
```diff
@@ -0,0 +1,14 @@
+clusterName: "test-cluster"
+commonDomain: "godsmack.westeurope.azure.gigantic.io"
+commonDomainResourceGroup: "godsmack"
+hostClusterCIDR: "10.0.0.0/16"
+hostClusterResourceGroup: "godsmack"
+hostClusterVirtualNetwork: "godsmack"
```
TheoBrigitte: Might want to remove those installation specific default values

helm/azure-operator-resource-chart/values.yaml
```diff
@@ -0,0 +1,14 @@
+clusterName: "test-cluster"
+commonDomain: "godsmack.westeurope.azure.gigantic.io"
+commonDomainResourceGroup: "godsmack"
+hostClusterCIDR: "10.0.0.0/16"
+hostClusterResourceGroup: "godsmack"
+hostClusterVirtualNetwork: "godsmack"
```
TheoBrigitte: Might want to remove those installation specific values

TheoBrigitte:
Is a public quay repo ok for this ?

helm/azure-operator-resource-chart/templates/cluster.yaml
```diff
@@ -0,0 +1,77 @@
...
+      etcd:
+        name: "{{.Values.commonDomain}}"
+        resourceGroup: "{{.Values.commonDomainResourceGroup}}"
+      ingress:
+        name: "{{.Values.commonDomain}}"
+        resourceGroup: "{{.Values.commonDomainResourceGroup}}"
+    hostCluster:
+      cidr: "{{.Values.hostClusterCIDR}}"
+      resourceGroup: "{{.Values.hostClusterResourceGroup}}"
+      virtualNetwork: "{{.Values.hostClusterVirtualNetwork}}"
```
kopiczko: `hostCluster` is not used anymore.

helm/azure-operator-resource-chart/values.yaml
kopiczko: I think this is only maybe wrong in political sense.

kopiczko:
You have to create the repo

helm/azure-operator-resource-chart/values.yaml
```diff
@@ -0,0 +1,11 @@
+clusterName: "test-cluster"
+commonDomain: ""
+commonDomainResourceGroup: ""
```
kopiczko: WDYT about putting here something like

```
commonDomain: "test-host-cluster.westeurope.azure.gigantic.io"
commonDomainResourceGroup: "test-host-cluster"
```

Then it's easy to just replace the value with the actual cluster name. The same thing you didi for `sshPublicKey`

kopiczko:
@indyfree I never knew which way to go. Small fmt entry maybe, to spread the knowledge? Or you think it's enough to just nag ppl in the PRs?

indyfree:
@kopiczko Your right. Altough nagging is fun I'm gonna get a `fmt` on the way :wink:

TheoBrigitte:
While we are at nitpicking, what about quotes, do we usually prefer to have quotes around string ?

helm/azure-operator-resource-chart/values.yaml
```diff
@@ -0,0 +1,15 @@
+clusterName: "test-cluster"
+commonDomain: "test-host-cluster.westeurope.azure.gigantic.io"
+commonDomainResourceGroup: "test-host-cluster"
+sshUser: "test-user"
+sshPublicKey: "ssh-rsa XXXXXXXXXXXXXXXXXXXXXXXXX"
```
rossf7: We use this in aws-operator too but if it doesn't get overridden it causes problems.

As it's not valid base64 the cloudconfig fails. I think it's better to put a default key or just anything base64 here. So you can test the cluster comes up even if you can't SSH to the nodes.

helm/azure-operator-resource-chart/templates/cluster.yaml
```diff
@@ -0,0 +1,78 @@
...
+        insecurePort: 8080
+        ip: "172.31.0.1"
+        securePort: 443
+      cloudProvider: "azure"
+      dns:
+        ip: "172.31.0.10"
+      domain: "cluster.local"
+      hyperkube:
+        docker:
+          image: "quay.io/giantswarm/hyperkube:v1.8.4_coreos.0"
```
rossf7: This docker image isn't used anymore. This is now set in k8scc so we can have the k8s version in the version bundles. Can we remove it?

helm/azure-operator-resource-chart/templates/cluster.yaml
```diff
@@ -0,0 +1,78 @@
...
+      ingressController:
+        docker:
+          image: "quay.io/giantswarm/nginx-ingress-controller:0.9.0-beta.11"
```
rossf7: Same for ingress controller.

kopiczko:
> While we are at nitpicking, what about quotes, do we usually prefer to have quotes around string ?

I like them.

PR #165, opened 2018-04-05
Opens required ports for monitoring

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/2372

Master:
screen shot 2018-04-05 at 12 28 23
Worker:
screen shot 2018-04-05 at 12 28 36

and when i submit the namespace, service and endpoint, monitoring targets come up correctly in prometheus bam bam

screen shot 2018-04-05 at 13 03 33

### DIFF
--- a/service/azureconfig/v1/arm_templates/security_groups_setup.json
+++ b/service/azureconfig/v1/arm_templates/security_groups_setup.json
@@ -48,7 +48,12 @@
     "masterSecurityGroupID": "[resourceId('Microsoft.Network/networkSecurityGroups', variables('masterSecurityGroupName'))]",
 
     "workerSecurityGroupName": "[concat(parameters('clusterID'), '-WorkerSecurityGroup')]",
-    "workerSecurityGroupID": "[resourceId('Microsoft.Network/networkSecurityGroups', variables('workerSecurityGroupName'))]"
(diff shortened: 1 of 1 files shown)


service/azureconfig/v1/resource/deployment/deployment.go
```diff
@@ -10,6 +10,11 @@ import (
 const (
 	mainDeploymentName     = "cluster-main-template"
 	templateContentVersion = "1.0.0.0"
+
```
JosephSalisbury: Pulling these all the way from the CR doesn't really make sense to me, it's not like they're configurable.

service/azureconfig/v1/arm_templates/security_groups_setup.json
```diff
@@ -41,6 +41,18 @@
     },
     "kubernetesIngressInsecurePort": {
       "type": "int"
+    },
+    "cadvisorPort": {
+        "type": "int"
+    },
```
kopiczko: I would make those variables and not pass it from main. This makes templates more usable on their own. You can also set default values and omit the parameters if you think it's important to have them configurable.

JosephSalisbury:
@TheoBrigitte || @r7vme || @kopiczko ptalaaaaaaaaaa

PR #166, opened 2018-04-06
Adds namespace resource

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/2372

Most of the implementation is lifted from `aws-operator`.

e.g:
```
$ kubectl get ns 6s34q -o yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: 2018-04-06T12:24:34Z
  labels:
    cluster: 6s34q
    customer: giantswarm
  name: 6s34q
  resourceVersion: "2516566"
  selfLink: /api/v1/namespaces/6s34q
  uid: 773a83b0-3995-11e8-af51-000d3a2870da
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```


service/azureconfig/v1/resource/namespace/current.go
```diff
@@ -0,0 +1,38 @@
...
+		if apierrors.IsNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find the namespace in the Kubernetes API")
+			// fall through
+		} else if err != nil {
+			return nil, microerror.Mask(err)
+		} else {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "found the namespace in the Kubernetes API")
+			namespace = manifest
+		}
+	}
```
xh3b4sd: This is not from the latest version. See https://github.com/giantswarm/aws-operator/blob/b0d90cde1ca7f66ae9edacbb00840d2d74a88012/service/awsconfig/v10/resource/namespace/current.go#L38-L47. The resource cancelation is missing in case the namespace is already in terminating state.

service/azureconfig/v1/resource/namespace/resource.go
```diff
@@ -0,0 +1,56 @@
...
+
+func New(config Config) (*Resource, error) {
+	if config.K8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.K8sClient must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	newResource := &Resource{
```
xh3b4sd: `newResource` -> `r`

PR #176, opened 2018-04-16
controller,service: controller.ClusterFramework -> controller.Cluster

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2995


service/service.go
```diff
@@ -38,11 +37,11 @@ type Config struct {
...
 type Service struct {
-	ClusterFramework *framework.Framework
-	Healthz          *healthz.Service
-	Version          *version.Service
+	Healthz *healthz.Service
+	Version *version.Service
 
-	bootOnce sync.Once
+	bootOnce          sync.Once
+	clusterController *controller.Cluster
```
kopiczko: No reason to make it exported. It's booted only inside the service.

service/controller/cluster.go
```diff
@@ -29,7 +29,11 @@ type ClusterFrameworkConfig struct {
 	TemplateVersion string
 }
 
-func NewClusterFramework(config ClusterFrameworkConfig) (*framework.Framework, error) {
+type Cluster struct {
+	*framework.Framework
+}
```
kopiczko: This is just a basic type embedding `*framework.Framework`, future `*controller.Controller`. This embedding is to avoid imports and aliased imports in `service.go`. This could be `framework *framework.Framework`, but I don't see strong reason to do so since it isn't a library.

kopiczko:
Covered by #179

PR #178, opened 2018-04-17
arm_templates: fix worker VMSS subnet

kopiczko:
I noted that today. Tested locally.

### DIFF
--- a/service/controller/v1/arm_templates/main.json
+++ b/service/controller/v1/arm_templates/main.json
@@ -423,7 +423,7 @@
           "vmssOsImageVersion": {"value": "[parameters('workerNodes')[0].osImage.version]"},
           "vmssVmCustomData": {"value": "[parameters('workerCloudConfigData')]"},
           "vmssLbBackendPoolId": {"value": "[reference('ingress_load_balancer_setup').outputs.backendPoolId.value]"},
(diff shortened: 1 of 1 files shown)


kopiczko:
Rolling.

service/controller/v1/arm_templates/main.json
```diff
@@ -423,7 +423,7 @@
           "vmssOsImageVersion": {"value": "[parameters('workerNodes')[0].osImage.version]"},
           "vmssVmCustomData": {"value": "[parameters('workerCloudConfigData')]"},
           "vmssLbBackendPoolId": {"value": "[reference('ingress_load_balancer_setup').outputs.backendPoolId.value]"},
-          "vmssVnetSubnetId": {"value": "[reference('virtual_network_setup').outputs.masterSubnetID.value]"}
+          "vmssVnetSubnetId": {"value": "[reference('virtual_network_setup').outputs.workerSubnetID.value]"}
```
xh3b4sd: How did you notice this? Can we ensure this with tests or something?

kopiczko: I just looked at the templates :) This works in both cases. The previous version was potentially less secure, as there were less isolation between masters and workers.

Only e2e test can cover this. It also would have to to use azure-sdk-for-go to check to take the private IP of machines and assert they are in correct network.

My common sense says it isn't worth it at all.

PR #181, opened 2018-04-18
v1: put master in VMSS

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2483


kopiczko:
Tested locally

kopiczko:
@r7vme 

> can you please describe how update will work? (e.g. new for new k8scc)

https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#how-to-bring-vms-up-to-date-with-the-latest-scale-set-model

> are you sure that volume will be reattached to the vm?

No. I will test that.

> btw, i don't see ARM vmss manifest

The same manifest as for workers is reused. It's `vmss.json`.

kopiczko:
Also check https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#vm-specific-updates

kopiczko:
We had some discussions yesterday. I will sum up here.

## Purpose

- Being able to reconcile ARM deployments. Otherwise we deal with immutable customData field in VMs. Step towards https://github.com/giantswarm/giantswarm/issues/2483 (as in description).
- Maintain only VMSS. Get rid of VM deployment, and availability set deployment. Less things to maintain is better most of the time.
- Allow easy master upgrades in the long run, by just [upgrading machines to the latest model](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#how-to-bring-vms-up-to-date-with-the-latest-scale-set-model). I checked that data on disks remains after the upgrade.

## Anti-Purpose

- Being able to scale masters. However that opens that possibility.

PR #182, opened 2018-04-18
v1: explicitly set cloud in azure.yaml

kopiczko:


xh3b4sd:
Why do we want to do this?

kopiczko:
It's confusing. We support only single `cloud` type. I'm not even sure what else can we support. Let's make it truly configurable when we need it.

PR #184, opened 2018-04-18
e2e testing - ipam

TheoBrigitte:
End to end tests are meant to run concurrently on CircleCI.
Each test have its own host cluster running within CircleCI and spin up a guest cluster on Azure.

Because we will have concurrent guest cluster running, we need a way to compute non overlapping guest network CIDR. This is usually done in host cluster, but because of the nature of e2e test (no single host cluster) this is not possible here.

Here I compute the CIDR based on the CircleCI job number by doing: `subnet = CircleCI job number % Quantity of network in e2e CIDR`.
So each job will always have the same CIDR, even if manually re-run.

```
Computation example:
e2e CIDR = 10.128.0.0/9
e2e Mask = 16
Allow for: 128 Subnets

CircleCI Job Number 988
988 % 128 = 92

10.128.0.0/16 + 92 = 10.220.0.0/16
```
Then job 989 has 10.221.0.0/16, job 990 has 10.222.0.0/16 an so on ...


TheoBrigitte:
ping @giantswarm/sre

integration/network/network.go
```diff
@@ -0,0 +1,173 @@
...
+func init() {
+	var err error
+
+	var azureNetwork *net.IPNet
+	{
+		azureNetwork, err = AzureNetwork()
+		if err != nil {
+			panic(err)
+		}
+		os.Setenv(EnvAzureCIDR, azureNetwork.String())
```
xh3b4sd: Why are we setting so many env vars? IMO we should try to use the process environment and less as possible. I see that it is necessary for initialization and such but once we run our code we should stay there. This is also why we want to migrate from shell out magic to use appr and helm client libs. We quickly run into situations in which we want to deal with concurrency or some such and then we are trapped. So I would like to approach this here via getters which are only set up via the env. See also https://github.com/giantswarm/aws-operator/blob/c13a7eeb4a574fc5678dfed969c9cc373107f10c/integration/env/env.go#L102. When we have getters for the test relevant configuration we can inject it where needed and we do not need to assume there is data in the process environment.

integration/network/network.go
```diff
@@ -0,0 +1,173 @@
...
+		os.Setenv(EnvAzureCalicoSubnetCIDR, azureCalicoSubnet.String())
+	}
+}
+
+// AzureNetwork either return network from CIDR found at EnvAzureCIDR environement variable or
+// determine network using EnvE2ECIDR, EnvE2EMask and EnvCircleBuildNumber.
+func AzureNetwork() (azureNetwork *net.IPNet, err error) {
+	azureCDIR := os.Getenv(EnvAzureCIDR)
+	if azureCDIR == "" {
+		e2eCIDR := os.Getenv(EnvE2ECIDR)
```
kopiczko: Why is there an option here? Why not require one of those?

integration/network/network.go
TheoBrigitte: Good catch, I wanted in fact to move this into `AzureNetwork` and `AzureSubnet` logic, so they would only setup the env vars if they are empty.

I do not really understand how your approach suggestion with getters. Could you elaborate on this.

integration/network/network.go
TheoBrigitte: `EnvE2ECIDR` is only required when `EnvAzureCIDR` is not provided. I can make this requirement explicit in init.

integration/network/network.go
xh3b4sd: I think we can use the information we computed here and dispatch where necessary without writing them into the process environment. Lets drop the `os.Setenv` magic and have a getter instead that returns the value of a variable we initialize here. I provided examples in the link of the aws-operator. Please ask more specific questions when things are unclear I feel like I cannot be more specific from the top of my head now.

integration/network/network.go
```diff
@@ -0,0 +1,173 @@
+package network
```
kopiczko: In general I'm confused. You provided nice description in the PR, but the code provides a lot of customizations which I doubt are necessary.

I like the idea of using job ID to create a subnet. That will allow us to run ~120 jobs in parallel (assuming the ID is increasing integer number). Why not just provide the CIDR depending on job ID and disallow all other options?

I'd like to know what is "host cluster" vnet that the peering is created with.

integration/network/network.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	return subnet, nil
+}
+
+// DetermineSubnet is deterministic, it will always return the same network given the same input.
+// It compute the number of available networks between cidr and mask.
+// And use `decider mod available networks` to determine which one to pick.
+// e.g. DetermineSubnet("10.255.0.0/9", 16, 42)  >  10.170.0.0/16
```
kopiczko: `10.255.0.0/9` this isn't correct CIDR network notation (even though parsable by golang library) - it contains network + IP. I don't want to read the code below, because I don't understand how it is supposed to work. I guess it always changes the second octet. If so why not just use string formatting? That's way simpler to reason about.

integration/network/network.go
```diff
@@ -0,0 +1,173 @@
...
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	return azureNetwork, nil
+}
+
+// AzureSubnet either return subnet from CIDR found at envVar environement variable or
+// compute the next free subnet using network, mask and allocatedSubnet (if any).
+func AzureSubnet(envVar string, network net.IPNet, mask int, allocatedSubnet ...net.IPNet) (subnet *net.IPNet, err error) {
```
kopiczko: Where does allocatedSubnets come from? Why we need such level of flexibility?

integration/network/network.go
TheoBrigitte: The use case is the following: if the CIDR is provided we use it, otherwise we compute a value for it.

On CircleCI we will run multiple e2e tests at the same time and each test need to have its own CIDR to avoid network addresses collision.

kopiczko: > if the CIDR is provided we use it

Why do we need to provide CIDR? Why can't we always use computed one?

> each test need to have its own CIDR to avoid network addresses collision

Why? Peering? What will be the created cluster peered with?

integration/network/network.go
TheoBrigitte: What the unnecessary options/customizations in your opinion ?


https://github.com/giantswarm/azure-operator/blob/1df1de711fc8dbdb3593fc176949dcadbd385520/integration/template/azure_operator_charts_values.go#L19-L23

So the host cluster running in ci has: `CIDR: "10.0.0.0/16"`
But vnet peering is happening with godsmack

integration/network/network.go
kopiczko: Why this function is public though? That's why I was so confused.

integration/network/network.go
TheoBrigitte: Yeah true, the correct CIDR here is `10.128.0.0/9` .

It is not always a whole octet it changes depending on the given mask and CIDR's mask. Let me explain with a full example
```
CIDR: 10.128.0.0/9    Mask: 17    Decider: 1

CIDR mask: 255.0.0.0 = 9        11111111.10000000.00000000.00000000
Mask: 255.255.0.0 = 17          11111111.11111111.10000000.00000000
                                          ^       ^ = 8 bits

Leaving us with (7 bits=) 256 possible subnets.
We wrap the decider in those possibilities: 1 mod 256 = 1

How much do we need to shift 1: 32 (bits in IP) - 17 (mask) = 15
1 bit shifted 15 times left = 10000000.00000000 > 0.0.128.0

We compute the final CIDR:
10.128.0.0 >    00001010.10000000.00000000.00000000
OR
0.0.128.0 >     00000000.00000000.10000000.00000000
                -----------------------------------
10.128.128.0 >  0001010.10000000.10000000.00000000
```
I am not sure there is a easier way to compute this.

integration/env/env.go
```diff
@@ -29,6 +38,37 @@ func init() {
...
+			panic(err)
+		}
+
+		os.Setenv(EnvAzureCIDR, cidrs.AzureCIDR)
+		os.Setenv(EnvAzureMasterSubnetCIDR, cidrs.MasterSubnetCIDR)
+		os.Setenv(EnvAzureWorkerSubnetCIDR, cidrs.WorkerSubnetCIDR)
+		os.Setenv(EnvAzureCalicoSubnetCIDR, cidrs.CalicoSubnetCIDR)
+	} else {
+		if os.Getenv(EnvAzureCalicoSubnetCIDR) == "" {
+			panic(fmt.Sprintf("env var '%s' must not be empty", EnvAzureCalicoSubnetCIDR))
```
kopiczko: Would be good to add `when AZURE_CIDR is set`.

integration/env/env.go
```diff
@@ -29,6 +38,37 @@ func init() {
...
+
+		cidrs, err := network.ComputeCIDR(uint(buildNumber))
+		if err != nil {
+			panic(err)
+		}
+
+		os.Setenv(EnvAzureCIDR, cidrs.AzureCIDR)
+		os.Setenv(EnvAzureMasterSubnetCIDR, cidrs.MasterSubnetCIDR)
+		os.Setenv(EnvAzureWorkerSubnetCIDR, cidrs.WorkerSubnetCIDR)
+		os.Setenv(EnvAzureCalicoSubnetCIDR, cidrs.CalicoSubnetCIDR)
```
kopiczko: Why setting env vars instead of populating more `var`s (next to cicrcleSHA)?

integration/network/network.go
TheoBrigitte: I did not manage to come up with a more elegant way of passing those CIDR values around. since they are read from environment by e2e-harness.

integration/env/env.go
TheoBrigitte: Because those values are later on read from environment by e2e-harness, when templating files under `template` folder.

kopiczko: Ack

PR #190, opened 2018-04-25
v1: freeze 0.1.0

kopiczko:


service/controller/v1/version_bundle.go
```diff
@@ -50,6 +50,6 @@ func VersionBundle() versionbundle.Bundle {
 		Name:         "azure-operator",
 		Time:         time.Date(2018, time.January, 7, 8, 35, 0, 0, time.UTC),
```
webwurst: This says something about "January". Is that correct?

puja108:
I have not gotten any e2e running on 1.10 yet, it's always timing out waiting for nodes to be schedulable, but they are clearly schedulable, not sure what is wrong there. I can try again with recent release, is 0.3.0 the right one to choose 0.4.0 looked strange to me.

puja108:
running e2e right now for 0.3.0 release version

PR #191, opened 2018-04-26
v1: cancel deployment deletion

kopiczko:
The problem is the cert secrets can be already deleted.

### DIFF
--- a/service/controller/v1/resource/deployment/resource.go
+++ b/service/controller/v1/resource/deployment/resource.go
@@ -11,6 +11,7 @@ import (
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/operatorkit/controller"
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
 
 	"github.com/giantswarm/azure-operator/client"
 	"github.com/giantswarm/azure-operator/service/controller/setting"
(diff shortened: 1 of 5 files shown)


kopiczko:
Need LGTM <3

service/controller/v1/resource/deployment/resource.go
```diff
@@ -39,19 +38,15 @@ type Config struct {
...
 
 	azure           setting.Azure
 	azureConfig     client.AzureConfig
 	templateVersion string
 }
 
 func New(config Config) (*Resource, error) {
-	if config.CertsSearcher == nil {
-		return nil, microerror.Maskf(invalidConfigError, "%T.CertsSearcher must not be empty", config)
-	}
```
xh3b4sd: Why is the certs searcher now gone?

service/controller/v1/resource/deployment/resource.go
```diff
@@ -90,6 +84,13 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	// Deleting the resource group will take care about cleaning
+	// deployments.
```
xh3b4sd: A bit more information on why we cancel would be cool. IIRC there was a nice comment already in your previous commit containing these information.

kopiczko:
I'm going with this now. e2e tests takes time and I need to run a few more. If comments need refining let's discuss further.

xh3b4sd:
All good from my side. Thank you.

PR #192, opened 2018-04-26
v1: cancel endpoints & service deletion

kopiczko:
They are deleted with the namespace. Endpoints does failing Azure API
calls on deletion so this change is necessary.

### DIFF
--- a/service/controller/v1/resource/deployment/resource.go
+++ b/service/controller/v1/resource/deployment/resource.go
@@ -87,6 +87,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// Deleting the resource group will take care about cleaning
 	// deployments.
 	if key.IsDeleted(customObject) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling deployment deletion: deleted with the resource group")
 		resourcecanceledcontext.SetCanceled(ctx)
 		return nil, nil
 	}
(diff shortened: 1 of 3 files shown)


service/controller/v1/resource/deployment/resource.go
```diff
@@ -87,6 +87,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// Deleting the resource group will take care about cleaning
 	// deployments.
 	if key.IsDeleted(customObject) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling deployment deletion: deleted with the resource group")
```
xh3b4sd: I find these log messages more confusing than helpful. It states canceling and then it states deleted and I do not know as a developer looking at these logs what this means. In the other operators we try to put simple messages without big implications or colons or brackets or other magic semantics or meanings. I do not understand this here.

kopiczko: What should be the message?

xh3b4sd: In this case probably `canceling deployment deletion`.

xh3b4sd: So far I always did it that way. See https://github.com/giantswarm/aws-operator/blob/b322a4d5cc1c7202742b544f546972ae26a5e2af/service/controller/v10/resource/cloudformation/current.go#L52-L54. One line for the context, one line for the cancelation. This action reason pattern is not cool here because we work with structured logging. When you want to add this then make use of the key/val pairs. Then the child gets actually a name and your intentions become more comprehensible.

PR #194, opened 2018-04-27
client, service, vendor: update azure-sdk-for-go

kopiczko:


service/controller/v1/resource/deployment/resource.go
```diff
@@ -112,11 +112,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 			}
 
 			d := deployment{
-				Name:                   *deploymentExtended.Name,
-				Parameters:             *deploymentExtended.Properties.Parameters,
-				ResourceGroup:          resourceGroupName,
-				TemplateURI:            *deploymentExtended.Properties.TemplateLink.URI,
-				TemplateContentVersion: *deploymentExtended.Properties.TemplateLink.ContentVersion,
+				Name: *deploymentExtended.Name,
 			}
```
kopiczko: Only name is checked in newCreateChange. I don't want to deal with `Parameters` type change. From `map` to `interface`. This is subject to change anyway.

PR #195, opened 2018-04-27
Vendor update latest operatorkit

MarcelMue:
Towards https://github.com/giantswarm/giantswarm/issues/3078


MarcelMue:
Going to fix build when I am back from uni

MarcelMue:
Going with this as it's just vendor and I want to get it through.

PR #203, opened 2018-05-13
Add germancloud support

r7vme:
There are multiple changes:
- added new configuration flag `--service.azure.cloud`
- added cloud support to azure client package (made it similar as done in Kubernetes)
- updated calico to 3.0.5 (as it was in changelog, but template was not updated)
- make `v2` active release and deprecate `v1`

Related: https://github.com/giantswarm/installations/pull/310
Related: https://github.com/giantswarm/k8scloudconfig/pull/368

Status: Tested in `tarantula` cluster creation started, but vmss fail to be created with "Internal server error" from Azure. Working on it.

### DIFF
--- a/client/azure.go
+++ b/client/azure.go
@@ -22,10 +22,19 @@ type AzureConfig struct {
 	ClientID string
 	// ClientSecret is the secret of the Active Directory Service Principal.
 	ClientSecret string
+	// The cloud environment identifier. Takes values from https://github.com/Azure/go-autorest/blob/ec5f4903f77ed9927ac95b19ab8e44ada64c1356/autorest/azure/environments.go#L13
+	Cloud string
 	// SubscriptionID is the ID of the Azure subscription.
 	SubscriptionID string
 	// TenantID is the ID of the Active Directory tenant.
 	TenantID string
+}
+
+// azureClientConfig contains all essential information to create an Azure client.
+type azureClientConfig struct {
+	subscriptionID          string
+	resourceManagerEndpoint string
+	servicePrincipalToken   *adal.ServicePrincipalToken
 }
 
 func (c AzureConfig) Validate() error {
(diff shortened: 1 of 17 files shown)


client/azure.go
```diff
@@ -22,12 +22,21 @@ type AzureConfig struct {
...
 	// TenantID is the ID of the Active Directory tenant.
 	TenantID string
 }
 
+// azureClientConfig contains all essential information to create an Azure client.
+type azureClientConfig struct {
+	subscriptionID          string
+	resourceManagerEndpoint string
+	servicePrincipalToken   *adal.ServicePrincipalToken
+}
```
r7vme: Made this similar to https://github.com/kubernetes/kubernetes/blob/v1.10.2/pkg/cloudprovider/providers/azure/azure_client.go#L146-L155

Idea is to get oAuth token once (put to this object) and use it for clients.

client/azure.go
```diff
@@ -123,83 +150,81 @@ func ResponseWasNotFound(resp autorest.Response) bool {
...
-	spt, err := newServicePrincipalToken(config, azure.PublicCloud.ResourceManagerEndpoint)
-	if err != nil {
-		return nil, microerror.Maskf(err, "creating service principal token")
-	}
-
-	client := resources.NewDeploymentsClient(config.SubscriptionID)
-	client.Authorizer = autorest.NewBearerAuthorizer(spt)
+func newDeploymentsClient(config *azureClientConfig) (*resources.DeploymentsClient, error) {
+	client := resources.NewDeploymentsClientWithBaseURI(
+		config.resourceManagerEndpoint,
```
r7vme: We need to use `config.resourceManagerEndpoint` of specific cloud. By default `NewDeploymentsClient` uses public one.

service/controller/v1/resource/deployment/resource_test.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 var testAzure = setting.Azure{
+	Cloud: "AZUREPUBLICCLOUD",
```
r7vme: Just a unit test fix for `v1`. Otherwise it does not pass validation.

service/controller/v2/cloudconfig/template.go
```diff
@@ -523,7 +523,7 @@ const (
 	cloudProviderConfFileName       = "/etc/kubernetes/config/azure.yaml"
 	cloudProviderConfFileOwner      = "root:root"
 	cloudProviderConfFilePermission = 0600
-	cloudProviderConfFileTemplate   = `cloud: AZUREPUBLICCLOUD
+	cloudProviderConfFileTemplate   = `cloud: {{ .Cloud }}
```
r7vme: This is second important thing after Azure client. We need to put proper value in cloud provider.

client/azure.go
```diff
@@ -123,83 +150,81 @@ func ResponseWasNotFound(resp autorest.Response) bool {
...
 	if err != nil {
 		return nil, microerror.Maskf(err, "creating OAuth config")
 	}
 
-	return adal.NewServicePrincipalToken(*oauthConfig, config.ClientID, config.ClientSecret, scope)
+	return adal.NewServicePrincipalToken(
+		*oauthConfig,
+		config.ClientID,
+		config.ClientSecret,
+		env.ServiceManagementEndpoint)
```
xh3b4sd: Can we please not do this? It reads awfully IMHO.

client/azure.go
```diff
@@ -123,83 +150,81 @@ func ResponseWasNotFound(resp autorest.Response) bool {
...
+// parseAzureEnvironment returns azure environment by name.
+func parseAzureEnvironment(cloudName string) (*azure.Environment, error) {
+	var env azure.Environment
+	var err error
+	if cloudName == "" {
+		env = azure.PublicCloud
+	} else {
+		env, err = azure.EnvironmentFromName(cloudName)
+	}
+	return &env, err
```
xh3b4sd: There should be error masking and I would question the need of returning a pointer here.

client/azure.go
r7vme: done.

client/azure.go
r7vme: done. Returning value and masked error in this function and wrpa error above, where function called.

service/controller/v2/version_bundle.go
```diff
@@ -9,16 +9,21 @@ import (
...
+			{
+				Component:   "cloudconfig",
+				Description: "Removed kube-state-metrics related components (will be managed by chart-operator).",
+				Kind:        versionbundle.KindRemoved,
+			},
+			{
+				Component:   "azure-operator",
+				Description: "Added German cloud support.",
+				Kind:        versionbundle.KindAdded,
+			},
```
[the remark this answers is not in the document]
r7vme: done

r7vme:
@xh3b4sd @kopiczko addressed comments. Please take a look.

Changed back v2 to WIP, because we have to fix lack of MSI in german cloud before considering active release.

Want to merge this changes as they are all necessary.

client/azure.go
xh3b4sd: You only fixed this in one place but the other places are still using the "undesired" style. So I would not like to wrap the function calls and I would also like to keep the error handling explicit. So instead of this. 

```
return foo(
  a, 
  b, 
  c, 
  d)
```

Lets please do this, **everywhere**. 

```
err := foo(a, b, c, d)
if err != nil {
  return microerror.Mask(err)
}

return nil
```

client/azure.go
```diff
@@ -123,83 +150,73 @@ func ResponseWasNotFound(resp autorest.Response) bool {
...
+func newServicePrincipalToken(config AzureConfig, env azure.Environment) (*adal.ServicePrincipalToken, error) {
+	oauthConfig, err := adal.NewOAuthConfig(env.ActiveDirectoryEndpoint, config.TenantID)
 	if err != nil {
 		return nil, microerror.Maskf(err, "creating OAuth config")
 	}
 
-	return adal.NewServicePrincipalToken(*oauthConfig, config.ClientID, config.ClientSecret, scope)
+	token, err := adal.NewServicePrincipalToken(*oauthConfig, config.ClientID, config.ClientSecret, env.ServiceManagementEndpoint)
+	if err != nil {
+		return nil, microerror.Maskf(err, "getting token")
```
xh3b4sd: Just FYI in practise the annotations are rather useless when masking existing errors. The masking mostly only makes sense when an error is created and a root cause is annotated. I noticed this is done that way in the azure-operator though. Just sharing my opinion here again.

r7vme:
CI ok, addressed all comments. Merging.

PR #208, opened 2018-05-24
improves import format

xh3b4sd:


service/controller/v2/resource/deployment/resource.go
```diff
@@ -21,8 +20,6 @@ import (
 const (
 	// Name is the identifier of the resource.
 	Name = "deploymentv2"
-
-	createTimeout = 30 * time.Minute
```
xh3b4sd: There are certain variables/constants that are not used at all as it seems. Boyscouting them away. Swoooosh.

service/controller/v2/resource/dnsrecord/create.go
```diff
@@ -4,8 +4,9 @@ import (
 	"context"
 
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
-	"github.com/giantswarm/azure-operator/service/controller/v2/key"
 	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/service/controller/v2/key"
```
xh3b4sd: The rest is just aligning the imports formatting to maintain the 3 blocks as we have them defined in https://github.com/giantswarm/fmt/blob/master/go/imports.md.

PR #211, opened 2018-05-24
improves log messages

xh3b4sd:
This PR is the start of a series of PRs towards log message improvements. The goal is to not diverge anymore from the other operators we build. This should make working across projects a lot easier.

### DIFF
--- a/service/controller/v2/resource/deployment/resource.go
+++ b/service/controller/v2/resource/deployment/resource.go
@@ -83,8 +83,9 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// Deleting the resource group will take care about cleaning
 	// deployments.
 	if key.IsDeleted(customObject) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling deployment deletion: deleted with the resource group")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "redirecting responsibility of deletion of deployments to resource group termination")
 		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


service/controller/v2/resource/deployment/resource.go
```diff
@@ -83,8 +83,9 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// Deleting the resource group will take care about cleaning
 	// deployments.
 	if key.IsDeleted(customObject) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling deployment deletion: deleted with the resource group")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "redirecting responsibility of deletion of deployments to resource group termination")
```
rossf7: It's good that the message is improved but this is really long and I think will be hard to read in logs. What about something like this?

`deletion of deployments is handled by resource group termination`

service/controller/v2/resource/deployment/resource.go
```diff
@@ -182,7 +181,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createState inter
 		}
 
 		for _, deploy := range deploymentsToCreate {
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating Azure deployments: creating %#q", deploy.Name))
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating deployment '%s'", deploy.Name))
```
rossf7: Good to remove Azure. This is azure-operator. 😄

service/controller/v2/resource/deployment/resource.go
xh3b4sd: The important part here is the redirecting. I made it to this. 

```
deletion of deployments is redirected to resource group termination
```

rossf7: Thanks, I like that.

PR #212, opened 2018-05-25
simplifies resourcegroup resource

xh3b4sd:


xh3b4sd:
The missing logging is my problem with the simplified resource. It is not really useful at all. When we now put logging in there it just says the same all the time and that does actually not have any meaning. I will test this change manually on `gollum` or something and see how it goes.

kopiczko:
IMO it now says updating/deleting instead of creating/updating/deleting. I think it's good to know azure-operator reconciles those resources, even w/o details, because there aren't many. Yesterday in aws-operator we noticed only drainer controller was working because of error in InitCtx function. Would be difficult to catch w/o any logs. Also when there is an error you have some reference point.

xh3b4sd:
There were some issues with the error handling on repeated deletion calls. I fixed this with a specific error matcher. This works now nicely and is an improvement as compared to the former version. I go with this now.

kopiczko:
Nice.

Issue #171, opened 2018-04-13
Asynchronous API calls


TheoBrigitte:
### Current state
azure-operator currently sits waits for every asynchronous request made to Azure API. This time could used to perform other operations and to improve efficiency towards handling large amount of guest clusters.
e.g. If two successive operation for creation and deletion of the same guest cluster are made. The delete operation would have to wait until the create command has completed, which is usually between 5 and 10 minutes on azure.

### Issue
We ran into an issue where e2e tests failed because azure-operator did not provided any output during 10 minutes, while it was operating a delete cluster operation.
see: https://github.com/giantswarm/azure-operator/pull/162#issuecomment-380468647

## Proposal
Let asynchronous azure API request operate asynchronously.
On each resync period we can read the current operation's [provision state](https://github.com/Azure/azure-sdk-for-go/blob/master/services/network/mgmt/2018-01-01/network/models.go#L974-L986) and proceed with adequate reconciliation operation.

## Implementation
* Provide resource's `ProvisioningState` in `CurrentState` and `DesiredState`
* Take `ProvisioningState` during reconciliation
* Do not wait and return an error during asynchronous operation. So we do not reconcile next resource, and next resync period will check for the current state of the current resource.

kopiczko:
Ping @giantswarm/sig-operator

xh3b4sd:
[...]

PR #213, opened 2018-05-25
simplifies deployment resource

xh3b4sd:
This PR aims to simplify the resource implementation by not using the CRUD resource. What I notice though is that every time the deployment is ensured, it takes way too much time. Like Azure is doing something heavy all the time, which should actually not be necessary. Here is an excerpt from the logs of the creation of the cluster until further reconciliations. 

```
ensuring deployment is created    17:02:19
 ensured deployment is created    17:08:25

ensuring deployment is created    17:08:32
 ensured deployment is created    17:21:35

ensuring deployment is created    17:21:39
 ensured deployment is created    17:23:44

ensuring deployment is created    17:23:49
 ensured deployment is created    17:25:51
```

We have to decide if this is a good thing or not. One weird thing we see here is that the time between these checks is not aligned with the resync period.


kopiczko:
> What I notice though is that every time the deployment is ensured, it takes way too much time.

That's the reason why we need async API calls first. See https://github.com/giantswarm/azure-operator/issues/171#issuecomment-382691891

Otherwise it won't be a big deal to fix https://github.com/giantswarm/giantswarm/issues/2483

xh3b4sd:
I would like to get the refactoring merged and do the asynch magic next. Like today. I was just not sure the behaviour observed here is expected. But when this is how it is I would like to go ahead.

kopiczko:
Code-wise it LGTM. Merge at your own responsibility.
I would personally first make sure I have a working async call code first.

xh3b4sd:
Ok when this is important to you, I am going to implement the async magic in a feature branch based on this one and then we merge along the chain.

TheoBrigitte:
I see two subjects here

First being: deployment resource takes way too much time
We aim to improve this by having async API calls #171 

Second being: simplification of deployment resource
I am fine with simplifying things, but I would like to know how you plan on having async API calls without `GetDesiredState` ?
To me it seems essential to read the current deployment status to able to perform async API calls. Sure we can do this within`EnsureCreated` but CRUD resource seems to be more suitable for such purposes.

> One weird thing we see here is that the time between these checks is not aligned with the resync period.
As we sometimes take longer than `resyncPeriod` (5mn) to process events, those events pile up and so we end up processing them one after another regardless of `resyncPeriod`.

kopiczko:
>I am fine with simplifying things, but I would like to know how you plan on having async API calls without GetDesiredState ?
>To me it seems essential to read the current deployment status to able to perform async API calls. Sure we can do this withinEnsureCreated but CRUD resource seems to be more suitable for such purposes.

I disagree with that. This is just a single `if` statement. Introducing that level of indirection for single if doesn't pay off IMO.

xh3b4sd:
I would have liked to use the CRUD resources because of their structure. Having the structure makes things explicit. Having the default resource interface is simpler though and depending on the use case you are on the edge of one of the other solution. Sometimes you are on the edge of usefulness on both sides at the same time even. This here is such a 50 50 case and I am not willing to fight for considering one approach over the other. The drawback with the default resource interface is that we need certain cascading conditionals which makes it non straight forward. The drawback with the CRUD resource interface is that the structure is a bit too much and things like the desired state do not seem to be too useful in certain edge cases of an implementation. 

Towards the asyn API calls. Azure provides something special, which we do not seem to need. See the PR which I based off this one. We can get around that and keep it simpler. With the async magic we would have to store operation IDs and that would make it way more complicated than it has to be now.

TheoBrigitte:
Ok that's fair enough, fine with me for those simplification.

xh3b4sd:
The other PR got three approvals. I merged everything in here. Going.

PR #214, opened 2018-05-28
Async

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3272. Note this PR is based off https://github.com/giantswarm/azure-operator/pull/213.

### DIFF
--- a/client/azure.go
+++ b/client/azure.go
@@ -70,6 +70,8 @@ type AzureClientSet struct {
 	DNSZonesClient *dns.ZonesClient
 	// InterfacesClient manages virtual network interfaces.
 	InterfacesClient *network.InterfacesClient
+	// VirtualNetworkClient manages virtual networks.
+	VirtualNetworkClient *network.VirtualNetworksClient
 	// VnetPeeringClient manages virtual network peerings.
 	VnetPeeringClient *network.VirtualNetworkPeeringsClient
 }
@@ -102,39 +104,39 @@ func NewAzureClientSet(config AzureConfig) (*AzureClientSet, error) {
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
 	groupsClient, err := newGroupsClient(clientConfig)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
 	dnsRecordSetsClient, err := newDNSRecordSetsClient(clientConfig)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
 	dnsZonesClient, err := newDNSZonesClient(clientConfig)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
 	interfacesClient, err := newInterfacesClient(clientConfig)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
+	virtualNetworksClient, err := newVirtualNetworkClient(clientConfig)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
 	vnetPeeringClient, err := newVnetPeeringClient(clientConfig)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
 	clientset := &AzureClientSet{
-		DeploymentsClient:   deploymentsClient,
-		GroupsClient:        groupsClient,
-		DNSRecordSetsClient: dnsRecordSetsClient,
-		DNSZonesClient:      dnsZonesClient,
-		InterfacesClient:    interfacesClient,
-		VnetPeeringClient:   vnetPeeringClient,
+		DeploymentsClient:    deploymentsClient,
+		GroupsClient:         groupsClient,
+		DNSRecordSetsClient:  dnsRecordSetsClient,
+		DNSZonesClient:       dnsZonesClient,
+		InterfacesClient:     interfacesClient,
+		VirtualNetworkClient: virtualNetworksClient,
+		VnetPeeringClient:    vnetPeeringClient,
 	}
 
 	return clientset, nil
@@ -180,6 +182,13 @@ func newDNSZonesClient(config *azureClientConfig) (*dns.ZonesClient, error) {
 
 func newInterfacesClient(config *azureClientConfig) (*network.InterfacesClient, error) {
 	client := network.NewInterfacesClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
+	client.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+
+	return &client, nil
+}
+
+func newVirtualNetworkClient(config *azureClientConfig) (*network.VirtualNetworksClient, error) {
+	client := network.NewVirtualNetworksClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	client.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
 
 	return &client, nil
--- a/service/controller/v2/key/key.go
+++ b/service/controller/v2/key/key.go
@@ -124,6 +124,20 @@ func IsDeleted(customObject providerv1alpha1.AzureConfig) bool {
 	return customObject.GetDeletionTimestamp() != nil
 }
 
+func IsFinalProvisioningState(s string) bool {
+	if s == "Succeeded" {
+		return true
+	}
+	if s == "Failed" {
+		return true
+	}
+	if s == "Canceled" {
+		return true
+	}
+
+	return false
+}
+
 // MasterSecurityGroupName returns name of the security group attached to master subnet.
 func MasterSecurityGroupName(customObject providerv1alpha1.AzureConfig) string {
 	return fmt.Sprintf("%s-%s", ClusterID(customObject), masterSecurityGroupSuffix)
--- a/service/controller/v2/resource/deployment/resource.go
+++ b/service/controller/v2/resource/deployment/resource.go
@@ -83,7 +83,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment is created")
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment")
 
 	resourceGroupName := key.ClusterID(customObject)
 	mainDeployment, err := r.newMainDeployment(customObject)
@@ -100,6 +100,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			},
 		},
 	}
+
 	d, err := deploymentsClient.Get(ctx, resourceGroupName, mainDeployment.Name)
 	if IsNotFound(err) {
 		_, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, newDeployment)
@@ -131,7 +132,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured deployment is created")
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured deployment")
 
 	return nil
 }
--- a/service/controller/v2/resource/deployment/resource.go
+++ b/service/controller/v2/resource/deployment/resource.go
@@ -2,6 +2,7 @@ package deployment
 
 import (
 	"context"
+	"fmt"
 
 	azureresource "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-02-01/resources"
 	"github.com/Azure/go-autorest/autorest/to"
@@ -87,7 +88,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	if err != nil {
 		return microerror.Mask(err)
 	}
-	d := azureresource.Deployment{
+	newDeployment := azureresource.Deployment{
 		Properties: &azureresource.DeploymentProperties{
 			Mode:       azureresource.Complete,
 			Parameters: &mainDeployment.Parameters,
@@ -97,13 +98,25 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			},
 		},
 	}
-	f, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, d)
+	d, err := deploymentsClient.Get(ctx, resourceGroupName, mainDeployment.Name)
 	if err != nil {
 		return microerror.Mask(err)
 	}
-	err = f.WaitForCompletion(ctx, deploymentsClient.Client)
-	if err != nil {
-		return microerror.Mask(err)
+	s := *d.Properties.ProvisioningState
+	fmt.Printf("\n")
+	fmt.Printf("provisioning state: %s\n", s)
+	fmt.Printf("\n")
+
+	if s == "InProgress" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment is in progress")
+
+		return nil
+
+	} else {
+		_, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, newDeployment)
+		if err != nil {
+			return microerror.Mask(err)
+		}
 	}
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured deployment is created")
--- a/service/controller/v2/resource/deployment/resource.go
+++ b/service/controller/v2/resource/deployment/resource.go
@@ -118,7 +118,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		s := *d.Properties.ProvisioningState
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", s))
 
-		if !isFinalState(s) {
+		if !key.IsFinalProvisioningState(s) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
 
 			return nil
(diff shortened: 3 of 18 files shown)


xh3b4sd:
I am looking for reviews here.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -97,16 +100,39 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	d, err := deploymentsClient.Get(ctx, resourceGroupName, mainDeployment.Name)
+	if IsNotFound(err) {
+		_, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, newDeployment)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment initiated")
+		reconciliationcanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation for custom object")
```
rossf7: <3

xh3b4sd:
Would like to get another approval on this.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -97,16 +100,39 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+		return nil
```
kopiczko: Do we need that `if` content? Can't we just fall trough?

service/controller/v2/resource/deployment/resource.go
```diff
@@ -97,16 +100,39 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	} else if err != nil {
 		return microerror.Mask(err)
+	} else {
+		s := *d.Properties.ProvisioningState
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", s))
+
+		if !isFinalState(s) {
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation for resource")
```
kopiczko: No need to cancel Resource. `return` is enough.

service/controller/v2/resource/resourcegroup/resource.go
```diff
@@ -105,21 +107,30 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
 	} else {
-		err = f.WaitForCompletion(ctx, groupsClient.Client)
-		if err != nil {
+		_, err := groupsClient.Delete(ctx, key.ClusterID(customObject))
+		if IsNotFound(err) {
+			// fall through
+		} else if err != nil {
 			return microerror.Mask(err)
+		} else {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "resource group deletion initiated")
```
kopiczko: Nitpick: I would write `resource group deletion in progress`. It could be initiated in previous reconciliation loop.

Don't we have to check `ProvisioningState` for the group as well? What will happen when you try to delete a group and deletion is in progress?

service/controller/v2/resource/deployment/resource.go
xh3b4sd: The case here is that we do not have a deployment. So this marks the very first reconciliation of it. That means we want to trigger the creation and cancel the reconciliation completely. Otherwise following resources fail because of missing Azure resources down the road. When we fall through we trigger creation but we would need to do some other weird stuff for the cancelation. It wouldn't get better I guess. Ideas welcome though.

service/controller/v2/resource/deployment/resource.go
xh3b4sd: Right. Done.

service/controller/v2/resource/resourcegroup/resource.go
xh3b4sd: I changed the logging. 

During my testing I saw the deletion being triggered several times until all resources got removed in Azure. `Delete` seems to trigger this properly without conflicts. This was actually my expectation of the API behaviour. On consecutive deletions we just run into this branch of the condition, keep the finalizer, and cancel.

xh3b4sd:
@kopiczko you want to approve or can I do something to make you happy here?

service/controller/v2/resource/deployment/resource.go
kopiczko: We need to fix other resources so they can handle that. There is not guarantee those will be created on next reconciliation loop.

service/controller/v2/resource/resourcegroup/resource.go
kopiczko: Cool.

service/controller/v2/resource/deployment/resource.go
kopiczko: https://github.com/giantswarm/giantswarm/issues/3286

xh3b4sd: I will check on this.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -97,16 +99,38 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+		if !isFinalState(s) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation for resource")
+
+			return nil
+		}
 	}
-	err = f.WaitForCompletion(ctx, deploymentsClient.Client)
+
+	_, err = deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, newDeployment)
```
TheoBrigitte: Should not we have the same logic for `ProvisioningState` here ?

Otherwise we will start reconciling other resources which might still be updating, not sure about the effect of this.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -172,6 +196,20 @@ func getDeploymentByName(list []deployment, name string) (deployment, error) {
...
 }
 
+func isFinalState(s string) bool {
+	if s == "Succeeded" {
+		return true
+	}
+	if s == "Failed" {
+		return true
+	}
+	if s == "Canceled" {
```
TheoBrigitte: `Failed` and `Canceled` should be treated as errors IMO

service/controller/v2/resource/resourcegroup/resource.go
```diff
@@ -105,21 +107,30 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
-	f, err := groupsClient.Delete(ctx, key.ClusterID(customObject))
+	_, err = groupsClient.Get(ctx, key.ClusterID(customObject))
 	if IsNotFound(err) {
 		// fall through
 	} else if err != nil {
 		return microerror.Mask(err)
 	} else {
-		err = f.WaitForCompletion(ctx, groupsClient.Client)
-		if err != nil {
+		_, err := groupsClient.Delete(ctx, key.ClusterID(customObject))
```
TheoBrigitte: What is the effect of triggering `Delete` for a resource group in deleting state ?

service/controller/v2/resource/deployment/resource.go
xh3b4sd: When we get here we have a final state and trigger the reconciliation again. This is desired IMO. Uncertain about side effects. I think we should fix problems once we have them. Right now there are no problems I am aware of with this approach. The only thing Pawel mentioned already is that other resources might have problems when cluster creation processes take more time than usual. I will check on this and we already track an issue for this.

service/controller/v2/resource/deployment/resource.go
xh3b4sd: What should then happen?

service/controller/v2/resource/resourcegroup/resource.go
xh3b4sd: This was discussed already in the other comments. The action seems to be idempotent which was what I expected in the first place.

service/controller/v2/resource/deployment/resource.go
kopiczko: Failed probably means the deployment failed. I think we should log the state in general. So we can make other decisions in the future on how to deal with that.

xh3b4sd: Any cool way of logging the state? What is important there? Feels like premature optimization. We guess into the blue and have no use case. I checked the [`DeploymentExtended`](https://godoc.org/github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-02-01/resources#DeploymentExtended). It has only pointers and such so we would need to marshal it and shit.

service/controller/v2/resource/resourcegroup/resource.go
TheoBrigitte: Cool.

Sorry for duplicate, previous comments did not showed up in diff.

service/controller/v2/resource/deployment/resource.go
TheoBrigitte: We used to have state from `_, err := DeploymentsCreateOrUpdateFuture.Done()` which was quite useful to quickly see what went wrong from the logs.

Logging state from `DeploymentExtended` seems more complex, so what about `Name`, `ID`, `Properties. CorrelationID` and `Properties.Timestamp` for now. So we can at least look up from Azure Portal what is going wrong in case of failure.

service/controller/v2/resource/deployment/resource.go
TheoBrigitte: Cool if we track that somewhere

service/controller/v2/resource/deployment/resource.go
xh3b4sd: Is it ok to track better debugging in a separate issue?

TheoBrigitte: ok

xh3b4sd: See https://github.com/giantswarm/giantswarm/issues/3290.

kopiczko: I think it's good to see in the logs that deployment failed. That can be to various reasons, e.g. be invalid json or non-existing resource id, or infinite number of other reasons. If that shows up as an error, fine. But I'm not sure about it.

client/azure.go
```diff
@@ -70,6 +70,8 @@ type AzureClientSet struct {
 	DNSZonesClient *dns.ZonesClient
 	// InterfacesClient manages virtual network interfaces.
 	InterfacesClient *network.InterfacesClient
+	// VirtualNetworkClient manages virtual networks.
+	VirtualNetworkClient *network.VirtualNetworksClient
```
xh3b4sd: To make the `vnetpeeringv2` resource resilient against timing issues due to the non blocking resource implementations we have to check for virtual networks. Thus the need for the client.

client/azure.go
```diff
@@ -102,39 +104,39 @@ func NewAzureClientSet(config AzureConfig) (*AzureClientSet, error) {
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
-
```
xh3b4sd: Just making this one block.

service/controller/v2/key/key.go
```diff
@@ -124,6 +124,20 @@ func IsDeleted(customObject providerv1alpha1.AzureConfig) bool {
 	return customObject.GetDeletionTimestamp() != nil
 }
 
+func IsFinalProvisioningState(s string) bool {
```
xh3b4sd: We use the state checks in multiple resources now.

service/controller/v2/resource/deployment/resource.go
xh3b4sd: I addressed this here now.

service/controller/v2/resource/endpoints/current.go
```diff
@@ -20,8 +20,10 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	// Deleting the K8s namespace will take care of cleaning the endpoints.
 	if key.IsDeleted(customObject) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling endpoints deletion: deleted with the namespace")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "redirecting responsibility of deletion of endpoints to namespace termination")
 		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
xh3b4sd: Making the endpoints resilient.

service/controller/v2/resource/vnetpeering/current.go
```diff
@@ -2,40 +2,86 @@ package vnetpeering
...
-	vnetPeering, err := r.getCurrentState(ctx, a)
-	if err != nil {
-		return network.VirtualNetworkPeering{}, microerror.Mask(err)
-	}
+	// In order to make vnet peering work we need a virtual network which we can
+	// use to peer. In case there is no virtual network yet we cancel the resource
+	// and try again on the next resync period. This is a classical scenario on
+	// guest cluster creation. If we would not check for the virtual network
+	// existence the client calls of CreateOrUpdate would fail with not found
+	// errors.
```
xh3b4sd: Making the vnet peering magic resilient.

service/controller/v2/resource/vnetpeering/delete.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // NewDeletePatch provide a controller.Patch holding the network.VirtualNetworkPeering to be deleted.
-func (r Resource) NewDeletePatch(ctx context.Context, azureConfig, current, desired interface{}) (*controller.Patch, error) {
+func (r *Resource) NewDeletePatch(ctx context.Context, azureConfig, current, desired interface{}) (*controller.Patch, error) {
```
xh3b4sd: This was some inconsistency I boyscouted here now.

xh3b4sd:
Please have another look. This should now also address https://github.com/giantswarm/giantswarm/issues/3286.

xh3b4sd:
I am looking for another feedback round here.

kopiczko:
I'll check after sig operator.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -97,16 +99,38 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		_, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, mainDeployment.Name, newDeployment)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment in progress")
+		reconciliationcanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation for custom object")
+
+		return nil
```
kopiczko: So again, do we need that part when other resources are ready for deployment resource to be async?

service/controller/v2/resource/endpoints/current.go
```diff
@@ -35,6 +37,9 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		manifest, err := r.k8sClient.CoreV1().Endpoints(namespace).Get(masterEndpointsName, apismetav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find the master endpoints in the Kubernetes API")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
kopiczko: If we cancel the resource when endpoints are not created, then how can we ever create them?

service/controller/v2/resource/resourcegroup/error.go
```diff
@@ -30,11 +29,8 @@ func IsNotFound(err error) bool {
...
 		dErr, ok := c.(autorest.DetailedError)
 		if ok {
-			sErr, ok := dErr.Original.(azure.ServiceError)
-			if ok {
-				if sErr.Code == "ResourceGroupNotFound" {
-					return true
-				}
+			if dErr.StatusCode == 404 {
+				return true
 			}
```
kopiczko: That looks better.

service/controller/v2/resource/vnetpeering/current.go
```diff
@@ -2,40 +2,86 @@ package vnetpeering
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
 
-func (r Resource) getCurrentState(ctx context.Context, azureConfig providerv1alpha1.AzureConfig) (network.VirtualNetworkPeering, error) {
-	vnetPeeringClient, err := r.getVnetPeeringClient()
-	if err != nil {
-		return network.VirtualNetworkPeering{}, microerror.Mask(err)
+			return network.VirtualNetworkPeering{}, nil
+		} else if err != nil {
+			return network.VirtualNetworkPeering{}, microerror.Mask(err)
+		} else {
```
kopiczko: This `else` is not needed, there are `return` in all other branches.

service/controller/v2/resource/deployment/resource.go
xh3b4sd: Uncertain. I cannot decide either way. Do you have an argument or use case or benefit for one way or the other?

service/controller/v2/resource/endpoints/current.go
xh3b4sd: I have to check on this.

service/controller/v2/resource/deployment/resource.go
kopiczko: My argument is simplification. Also if we run into timing issues we spot them earlier.

xh3b4sd: I will check on this.

service/controller/v2/resource/vnetpeering/current.go
xh3b4sd: I keep this for consistency.

service/controller/v2/resource/endpoints/current.go
xh3b4sd: Yes this was wrong. I fixed it like for the virtual networks. Now we check if the interfaces are there and if not we cancel the resource.

service/controller/v2/resource/deployment/resource.go
xh3b4sd: I removed this. Works.

xh3b4sd:
Please have another look Mr. @kopiczko. This should be fine now.

xh3b4sd:
Tested on `godsmack`. Also verified the master endpoints are set.

PR #218, opened 2018-06-04
sets informer settings explicitly to default values

xh3b4sd:
That is a left over from testing. It does not hurt but using the default values again is just fine.

### DIFF
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -1,8 +1,6 @@
 package controller
 
 import (
-	"time"
-
 	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	gsclient "github.com/giantswarm/apiextensions/pkg/clientset/versioned"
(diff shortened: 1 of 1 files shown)


service/controller/cluster.go
```diff
@@ -84,8 +82,8 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 			Logger:  config.Logger,
 			Watcher: config.G8sClient.ProviderV1alpha1().AzureConfigs(""),
 
-			//ResyncPeriod: informer.DefaultResyncPeriod,
-			ResyncPeriod: 1 * time.Minute,
+			RateWait:     informer.DefaultRateWait,
+			ResyncPeriod: informer.DefaultResyncPeriod,
```
kopiczko: Do we even need to set those?

xh3b4sd: We do not need to do it but I would like to do it to be explicit and have it easier to change them for testing.

PR #219, opened 2018-06-04
does not cancel resource on deletion when resources should be removed

xh3b4sd:
During testing I figured out there is a problem with cluster deletion. Right now the vnet peerings are not deleted. This PR fixes the problem. I tested this successfully on `godsmack`.

### DIFF
--- a/service/controller/v2/resource/vnetpeering/current.go
+++ b/service/controller/v2/resource/vnetpeering/current.go
@@ -26,7 +26,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// guest cluster creation. If we would not check for the virtual network
 	// existence the client calls of CreateOrUpdate would fail with not found
 	// errors.
-	{
+	if !key.IsDeleted(customObject) {
 		c, err := r.getVirtualNetworksClient()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


service/controller/v2/resource/vnetpeering/current.go
```diff
@@ -26,7 +26,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// guest cluster creation. If we would not check for the virtual network
 	// existence the client calls of CreateOrUpdate would fail with not found
 	// errors.
-	{
+	if !key.IsDeleted(customObject) {
```
tuommaki: Given the above comment, is correct ordering guaranteed when cluster is being deleted? More specifically: Is ordering guaranteed in the sense that "not found error" described above doesn't happen when deleting?

kopiczko: vnet can be deleted before the peering is deleted. All is async here. vnet is deleted with the resource group, peering is deleted with the sdk. So during deletion we don't need to check if vnet exists, we just need to make sure peering is deleted.

I think `Resource` interface would make more sense to make things simpler here. But for now  this is good, and very nice catch.

PR #221, opened 2018-06-05
adds new instance resource and wires it

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3272. The idea here is to have a new resource which manages the update mechanism as we want to implement it for Azure. 

> AFAIK we use the resource manager magic via the Azure API. Deployments are reconciled constantly. There is something like scale set models. Once the model is updated instances can be updated to use it. The currently used upgrade strategy is `manual` which means we have to trigger the instance update. We want to leverage this approach in order to have control over the draining. There is https://github.com/giantswarm/node-operator which we already integrated into the `aws-operator` and `kvm-operator`. During the reconciliation in the `azure-operator` we want to list all instances and check if they have to be updated. I am not sure yet how this check looks like. When we identify an instance has to be updated we drain the associated guest cluster node and once this is done we trigger the instance update in the Azure API. We repeat this process with each reconciliation loop until we upgraded all instances of the guest cluster. Later we might want to do this in batches for bigger clusters but this is some optimization aspect we can just add on top of it later on if necessary.

Also see https://gigantic.slack.com/archives/C6H0T6W1X/p1528118788000367.

### DIFF
--- a/service/controller/v2/resource/instance/create.go
+++ b/service/controller/v2/resource/instance/create.go
@@ -1,0 +1,25 @@
+package instance
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/service/controller/v2/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("EnsureCreated called for cluster ID '%s'", key.ClusterID(customObject)))
+
+	// TODO list all instances
+	// TODO find the first instance not having the latest scale set model applied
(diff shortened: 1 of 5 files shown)


xh3b4sd:
Successfully tested on `godsmack`.

xh3b4sd:
I am going with this. Please do not hesitate to provide further feedback if you want.

kopiczko:
`instance` name is misleading for me. AFAIK this resource is going to make upgrade API calls.
The resource skeleton is fine of course. We can rename it any time. I think separate resource, and maybe even controller (for different resync period) makes total sense here.

xh3b4sd:
Please suggest a better name. I think down the road we will have to go with a separate controller and yet another resource to deal with the `NodeConfig` CRs. I guess we have to garbage collect them in some general way then because there may be no state we can check that an update just happened. Will have to think about this more when the time is right.

kopiczko:
I don't want to give a name for a resource I'm not 100% clear on functionality name. If you don't more expressive ideas, that's fine. We can change it later if it makes sense.

PR #222, opened 2018-06-05
manage ingress load balancer by kubernetes

r7vme:
Task: https://github.com/giantswarm/giantswarm/issues/3268

Idea is that we are managing ingress load balancer with Kubernetes itself. So in this change:
- remove ingress load balancer creation from ARM templates
- add systemd unit that creates kubernetes service type load balancer (with external-dns annotation to create DNS A record)

### DIFF
--- a/service/controller/v2/cloudconfig/template.go
+++ b/service/controller/v2/cloudconfig/template.go
@@ -647,3 +647,42 @@ ExecStart=-/bin/bash -c "if ! findfs LABEL=$LABEL > /tmp/label.$LABEL; then wipe
 WantedBy=multi-user.target
 `
 )
+
+// TODO: This part should be removed when parameterized charts will be supported.
+const (
+	ingressLBFileName       = "/srv/k8s-ingress-loadbalancer.yaml"
+	ingressLBFileOwner      = "root:root"
+	ingressLBFilePermission = 0600
+	ingressLBFileTemplate   = `apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-loadbalancer
+  namespace: kube-system
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: ingress.{{ .ClusterDNSDomain }}.
+  labels:
+    app: ingress-loadbalancer
+spec:
+  type: LoadBalancer
+  ports:
+  - name: http
+    port: 80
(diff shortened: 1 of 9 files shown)


r7vme:
I don't know how good Azure ARM at reconciling existing templates. Will it be able to handle existing load balancer deletion or not. I would say yes this change is breaking.

kopiczko:
>  I am wondering which implications this change has for existing clusters.

From: https://msdn.microsoft.com/en-us/library/azure/dn790564.aspx

mode | Yes | String | Specifies the mode that is used to deploy resources. This value could be either Incremental or Complete. In Incremental mode, resources are deployed without deleting existing resources that are not included in the template. In complete mode resources are deployed and existing resources in the resource group not included in the template are deleted.
-- | -- | -- | --

I think we'd have to switch mode to `Complete` and test things.

service/controller/v2/cloudconfig/template.go
```diff
@@ -647,3 +647,43 @@ ExecStart=-/bin/bash -c "if ! findfs LABEL=$LABEL > /tmp/label.$LABEL; then wipe
 WantedBy=multi-user.target
 `
 )
+
+// TODO: This part should migrate to nginx-ingress-controller helm chart
+// when parametrized charts will be supported.
```
rossf7: Typo on Parameterized but also we (@giantswarm/team-batman) need to get parameters in chart-operator working so this isn't needed! 😄

r7vme:
> I think we'd have to switch mode to Complete and test things.

This can be dangerous, because in this case ARM will delete kubernetes managed resources (volumes, load balancers, etc.)

service/controller/v2/cloudconfig/template.go
r7vme: Thanks, fixed :) would love to not have this piece here too!

kopiczko:
>> I think we'd have to switch mode to Complete and test things.
>
> This can be dangerous, because in this case ARM will delete kubernetes managed resources (volumes, load balancers, etc.)

Yeah also peering, and maybe other things.

calvix:
I don't really have that much insight but if it works with updates and deletion then LGTM.

r7vme:
tested, merging

PR #224, opened 2018-06-06
implements basic update mechanism for instances

xh3b4sd:


service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +18,91 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		for _, v := range result.Values() {
+			if *v.LatestModelApplied {
+				continue
+			}
+
+			instanceID = *v.InstanceID
+			instanceName = key.ToInstanceName(customObject, *v.InstanceID)
+
+			if !key.IsFinalProvisioningState(*v.ProvisioningState) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("instance '%s' is in state '%s'", instanceName, *v.ProvisioningState))
```
kopiczko: A general thing I notice is you often do `'%s'`, there is `%q` directive which quotes a string.

service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +18,91 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		g := key.ResourceGroupName(customObject)
+		s := key.WorkerVMSSName(customObject)
+		result, err := c.List(ctx, g, s, "", "", "")
+		if IsScaleSetNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find the scale set")
+			resourcecanceledcontext.SetCanceled(ctx)
```
kopiczko: No need to cancel simple resource. Also applies to other calls.

service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +18,91 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+		if instanceID == "" {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "no instance ID found that needs to be updated")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+
+			return nil
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be updated", instanceName))
```
kopiczko: I think this log message is redundant.

service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +18,91 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring instance '%s' to be updated", instanceName))
+
+		c, err := r.getScaleSetsClient()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		g := key.ResourceGroupName(customObject)
+		s := key.WorkerVMSSName(customObject)
+		IDs := compute.VirtualMachineScaleSetVMInstanceRequiredIDs{
```
kopiczko: Nitpick: capital letter var

service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +18,91 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find the scale set")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+
+			return nil
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+
+		for _, v := range result.Values() {
```
kopiczko: Are values guaranteed to be returned in the same order? I'd sort them first by name to make sure we update only one instance at a time.

vendor/github.com/giantswarm/e2e-harness/pkg/framework/host.go
```diff
@@ -187,17 +187,17 @@ func (h *Host) InstallCertResource() error {
...
+			err := HelmCmd("registry install quay.io/giantswarm/apiextensions-cert-config-e2e-chart:stable -- -n cert-config-e2e --set commonDomain=${COMMON_DOMAIN} --set clusterName=${CLUSTER_NAME}")
 			if err != nil {
 				return microerror.Mask(err)
 			}
 
 			return nil
 		}
 		b := NewExponentialBackoff(ShortMaxWait, ShortMaxInterval)
-		n := newNotify("cert-resource-lab-chart install")
+		n := newNotify("cert-config-e2e-chart install")
```
kopiczko: When e2e-harness is updated we probably need to fix e2e tests. Because of COMMON_DOMAIN madness.

service/controller/v2/resource/instance/create.go
xh3b4sd: What problem do we have when not sorting? I do not see it.

service/controller/v2/resource/instance/create.go
xh3b4sd: Yeah this is how you name variables in go AFAIK. What is the problem?

tuommaki: I thing the acronym capitalization "rule" doesn't apply when the name starts with it, so I'd say it's totally fine to be `ids`. Also found same style from K8s repo with $RANDOM browsing: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_gc.go#L53

client/azure.go
```diff
@@ -194,6 +209,20 @@ func newVirtualNetworkClient(config *azureClientConfig) (*network.VirtualNetwork
...
+	client.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+
+	return &client, nil
+}
+
+func newVirtualMachineScaleSetVMsClient(config *azureClientConfig) (*compute.VirtualMachineScaleSetVMsClient, error) {
+	client := compute.NewVirtualMachineScaleSetVMsClient(config.subscriptionID)
+	client.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+
+	return &client, nil
```
TheoBrigitte: When there is no need to return `error` it should be dropped from the function signature. I got that this is aligned with other `new*Client` func, but this should be refactored IMO. Maybe in a different PR though.

xh3b4sd: Tracking in https://github.com/giantswarm/giantswarm/issues/3355.

service/controller/v2/key/key.go
```diff
@@ -195,6 +195,10 @@ func ToCustomObject(v interface{}) (providerv1alpha1.AzureConfig, error) {
 	return customObject, nil
 }
 
+func ToInstanceName(customObject providerv1alpha1.AzureConfig, instanceID string) string {
+	return fmt.Sprintf("%s-worker-%06s\n", ClusterID(customObject), instanceID)
```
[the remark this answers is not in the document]
xh3b4sd: Good catch. Removed.

service/controller/v2/key/key.go
```diff
@@ -195,6 +195,10 @@ func ToCustomObject(v interface{}) (providerv1alpha1.AzureConfig, error) {
...
+}
```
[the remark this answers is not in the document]
xh3b4sd: Done.

service/controller/v2/resource/instance/create.go
xh3b4sd: Well, it does not harm either. Anyways, I removed these calls. I kept the logging as it is though.

vendor/github.com/giantswarm/e2e-harness/pkg/framework/host.go
xh3b4sd: How does it have to be fixed then?

service/controller/v2/resource/instance/create.go
kopiczko: Because we'll have logs like:

```
found instance '%s' has to be updated
ensuring instance '%s' to be updated
```

It looks like the same message in other words. But I guess you want to leave it like that. Do it if it makes you happy.

service/controller/v2/resource/instance/create.go
kopiczko: So there can be more than single instance rolling at a time if rolling takes longer than resync period if I understand it correctly.

service/controller/v2/resource/instance/create.go
xh3b4sd: I would like to keep this because it expresses different semantics. It reflects we found the instance and then ensuring the instance we found is updated. These are different things.

service/controller/v2/resource/instance/create.go
xh3b4sd: In this case we see that the provisioning state is not final and cancel reconciliation.

xh3b4sd: So we never update more than one per reconciliation loop. We can batch this in the future when we want to but I would like to do this separately as soon as we have to. For now I would like to just make it work reliably.

kopiczko: I understand when `state is not final` then upgrade is in progress.

Let's say on the first loop we receive instances like

```
inst1 not_latest_model final_state
inst2 not_latest_model final_state
```

So we schedule upgrade for `inst1` because it's the first instance with not latest model in final state.

Second loop we receive list like:

```
inst1 not_latest_model not_final_state
inst2 not_latest_model final_state
```

So we don't schedule upgrade because first instance with not latest model is in not final state.

Third loop we receive list like (order not guaranteed):

```
inst2 not_latest_model final_state
inst1 not_latest_model not_final_state
```

So we schedule upgrade for `inst2` because it's the first instance with not latest model in final state.

- - -

Therefore I think we should either sort this list of check if there is *any* instance in not final state.

xh3b4sd: I added another loop for checking if any instance is in a non final provisioning state. Please have another look if this is cool for you.

xh3b4sd:
Fishing for LGTM.

service/controller/v2/resource/instance/create.go
```diff
@@ -15,11 +17,94 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+			return nil
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be updated", instanceName))
+	}
+
+	// Trigger the update for the found instance.
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring instance '%s' to be updated", instanceName))
```
kopiczko: I think I told that already there is a `%q` directive for quotation. I think it's more idiomatic to format it like `fmt.Sprintf("ensuring instance %q to be updated", instanceName)`. This is a minor.

xh3b4sd: I replied to you in the discussion above towards this.

PR #227, opened 2018-06-08
some refactorings

xh3b4sd:
Fixes https://github.com/giantswarm/giantswarm/issues/3355.


client/azure_client_set.go
```diff
@@ -4,19 +4,18 @@ import (
 	"errors"
 	"net/http"
 
-	"github.com/giantswarm/microerror"
-	"github.com/giantswarm/micrologger"
-
```
xh3b4sd: Making it one block.

client/azure_client_set.go
```diff
@@ -4,19 +4,18 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-02-01/resources"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
 )
 
-type AzureConfig struct {
+type AzureClientSetConfig struct {
```
xh3b4sd: It is called `AzureClientSet` and `NewAzureClientSet`.

client/azure_client_set.go
```diff
@@ -99,62 +98,25 @@ func NewAzureClientSet(config AzureConfig) (*AzureClientSet, error) {
...
+	clientSet := &AzureClientSet{
+		DeploymentsClient:               newDeploymentsClient(c),
+		GroupsClient:                    newGroupsClient(c),
+		DNSRecordSetsClient:             newDNSRecordSetsClient(c),
+		DNSZonesClient:                  newDNSZonesClient(c),
+		InterfacesClient:                newInterfacesClient(c),
+		VirtualNetworkClient:            newVirtualNetworkClient(c),
+		VirtualMachineScaleSetVMsClient: newVirtualMachineScaleSetVMsClient(c),
+		VirtualMachineScaleSetsClient:   newVirtualMachineScaleSetsClient(c),
+		VnetPeeringClient:               newVnetPeeringClient(c),
```
xh3b4sd: This is is to fix the reference issued.

service/controller/cluster.go
```diff
@@ -35,33 +35,11 @@ type Cluster struct {
...
-	if err := config.Azure.Validate(); err != nil {
-		return nil, microerror.Maskf(invalidConfigError, "%T.Azure.%s", config, err)
-	}
-	if err := config.AzureConfig.Validate(); err != nil {
-		return nil, microerror.Maskf(invalidConfigError, "%T.AzureConfig.%s", config, err)
-	}
-	if config.ProjectName == "" {
-		return nil, microerror.Maskf(invalidConfigError, "%T.ProjectName must not be empty", config)
-	}
-	if config.TemplateVersion == "" {
```
xh3b4sd: All these things are not used at this level but only forwarded. They are validated where they are used.

client/azure_client_set.go
tuommaki: Nitpick: I think this is a bit verbose, but the obvious suggestion of renaming `client` -> `azure` (with `azure.ClientSetConfig`) might not be that good idea either because I see this already in use from `go-autorest` library. I cannot figure better so I guess this is fine.

PR #235, opened 2018-06-13
Fix ingress dns domain for load balancer

r7vme:
I need full cluster domain for my ingress load balancer. But for some reason i've used DNSZonePrefixIngress that returns only `xxxxx.k8s`, but i need full `xxxxx.k8s.gollum.westeurope.gigantic.io`. So added new function that provides cluster doman.


r7vme:
Tested in `godsmack`.

We are more or less good with BYOC, because we don't have intermittent zone like `k8s.xxxx`. Instead we create own zone per cluster (id.k8s.xxxx) and add NS servers directly to base DNS zone (inside host cluster reosurce group).

This change adds nothing new, i just need a function that gives me cluster domain.

Issue #233, opened 2018-06-11
Azure: arm reconciliation removes all unknown resources (PVCs, load balancers) 


r7vme:
load balancer "kubernetes" that managed by Kubernetes was deleted by azure-operator (ARM deployment). Same with test PVC that i created.

See here on "incremental" and "complete" deployments

https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-deploy#incremental-and-complete-deployments

xh3b4sd:
Does that mean we have to define it in the templates somehow or what is the issue here?

r7vme:
Kubernetes creates own resources in Azure (PVCs, load balancers) and we can not predict what resources will be created, so we should not / can not define them in ARM. In other words, not all resources in resource groups will be managed by azure-operator (arm templates), some are dynamic and will be managed by Kubernetes.

Azure has two deployment modes for ARM templates: incremental and complete.

> In complete mode, Resource Manager deletes resources that exist in the resource group but are not specified in the template.
In incremental mode, Resource Manager leaves unchanged resources that exist in the resource group but are not specified in the template.

We need to switch to incremental mode.

xh3b4sd:
Is there a specific reason why we used `complete` so far? Are there any implications to this switch known? I can imagine this is a tiny change with a huge rat tail.

xh3b4sd:
Also, does this break something at the moment?

r7vme:
> Is there a specific reason why we used complete so far? 

afaik Azure ARM does not manage state (as CloudFormation does, i can be wrong), so they can not delete/replace existing resources if they are not listed in ARM template.

xh3b4sd:
During testing I saw this error in the portal. Is this related to the issue here? My question above is still open. I do not see that something obviously breaks but the portal suggests there was an issue. My cluster seems to work AFAICT though. 

```
{
    "authorization": {
        "action": "Microsoft.Compute/virtualMachineScaleSets/manualupgrade/action",
        "scope": "/subscriptions/1be3b2e6-497b-45b9-915f-eb35cae23c6a/resourceGroups/zm7fy/providers/Microsoft.Compute/virtualMachineScaleSets/zm7fy-worker"
    },
    "caller": "3ba8011d-5fc0-43f0-bd74-a50c6cfae82c",
    "channels": "Operation",
    "claims": {
        "aud": "https://management.core.windows.net/",
[...]

PR #236, opened 2018-06-13
uses incremental deployment mode to preserve k8s LB

xh3b4sd:
Towards https://github.com/giantswarm/azure-operator/issues/233.


xh3b4sd:
There are still a lot of route writes and they often fail. Uncertain if this is related here. Don't think so. SRE input please.

screen shot 2018-06-13 at 17 31 40

r7vme:
Yes, incremental helps. Standalone resources are now preserved, all good.

But as you posted we still have issue with routes. Kubernetes adds routes into Azure route table, but ARM (azure operator) deletes them. Not sure what we can do here.

r7vme:
> Can we go with this change then and sort the routes issue separately?

Definitely this what we need, even if it's not sorting all the issues.

xh3b4sd:
Ok since the issue with the routes is something else I go with this now.

Issue #231, opened 2018-06-11
Azure: No docker disks on vmss masters


r7vme:
Master looks fine, because this error silently ignored.

```
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: dev-sdd.device: Job dev-sdd.device/start timed out.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: Timed out waiting for device dev-sdd.device.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: Dependency failed for Formats the disk drive.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: Dependency failed for Mounts disk to /var/lib/docker.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: var-lib-docker.mount: Job var-lib-docker.mount/start failed with result 'dependency'.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: format-docker-disk.service: Job format-docker-disk.service/start failed with result 'dependency'.
Jun 11 10:20:22 8hpfx-master-000000 systemd[1]: dev-sdd.device: Job dev-sdd.device/start failed with result 'timeout'.
```

```
r7vme@8hpfx-master-000000 /etc/systemd/system $ lsblk 
NAME    MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
sda       8:0    0 30.4G  0 disk  
|-sda1    8:1    0  128M  0 part  /boot
|-sda2    8:2    0    2M  0 part  
|-sda3    8:3    0    1G  0 part  
| `-usr 254:0    0 1016M  1 crypt /usr
|-sda4    8:4    0    1G  0 part  
|-sda6    8:6    0  128M  0 part  /usr/share/oem
|-sda7    8:7    0   64M  0 part  
`-sda9    8:9    0 28.1G  0 part  /
sdb       8:16   0   16G  0 disk  
`-sdb1    8:17   0   16G  0 part  /mnt/resource
sdc       8:32   0   50G  0 disk  /var/lib/etcd
sr0      11:0    1  732K  0 rom   
```

xh3b4sd:
Can you please provide more information here? Does that happen during cluster creation or during updates or scaling? Do we know the root cause? Why does the master look fine when the VMSS is missing a disk?

r7vme:
[...]

PR #237, opened 2018-06-13
Add etcd disk for vmss master

r7vme:
Fixes: https://github.com/giantswarm/azure-operator/issues/231

### DIFF
--- a/service/controller/v2/arm_templates/main.json
+++ b/service/controller/v2/arm_templates/main.json
@@ -347,7 +347,10 @@
           "vmssName": {"value": "[concat(parameters('clusterID'), '-master')]"},
           "vmssVmSize": {"value": "[parameters('masterNodes')[0].vmSize]"},
           "vmssVmCount": {"value": "[length(parameters('masterNodes'))]"},
-          "vmssVmDataDisks": {"value": [{"name": "DockerDisk", "sizeGB": 50}]},
+          "vmssVmDataDisks": {"value": [
+                  {"name": "EtcdDisk", "sizeGB": 100},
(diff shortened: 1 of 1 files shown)


service/controller/v2/arm_templates/main.json
```diff
@@ -347,7 +347,10 @@
           "vmssName": {"value": "[concat(parameters('clusterID'), '-master')]"},
           "vmssVmSize": {"value": "[parameters('masterNodes')[0].vmSize]"},
           "vmssVmCount": {"value": "[length(parameters('masterNodes'))]"},
-          "vmssVmDataDisks": {"value": [{"name": "DockerDisk", "sizeGB": 50}]},
+          "vmssVmDataDisks": {"value": [
+                  {"name": "EtcdDisk", "sizeGB": 100},
```
r7vme: 100GB not because we really need 100GB of etcd data, but because Azure has linear dependency between disk size and  performance. So 100GB will be P10 type, which gives 500IOPS and 100MBps, which is fine for etcd (requres 50-500 IOPS sequental)

https://docs.microsoft.com/en-us/azure/virtual-machines/linux/premium-storage

xh3b4sd: Do we know how this behaves with reboots, VM moves and updates?

r7vme: Reboot is fine. VM move and update (vm recreation) will create new disk.

Actually it can be good idea to adress this problem (etcd data preservation) in this PR. Let me check few things.

r7vme:
@xh3b4sd this change is relevant (as we are still OK to use vmss for master) and can be merged.

Can you please approve?

PR #239, opened 2018-06-14
Prevent routes removal while reconciliation

r7vme:
Related-to: https://github.com/giantswarm/azure-operator/issues/233

Tested in `godsmack`

This change fixes issue when routes created by Kubernetes controller manager are removed while ARM reconciliation. To avoid this we adding new paramter "initialProvisioning" and deploying route table only on first run. On reconciliation "initialProvisioning" set to "No", which disables route table reconciliation.

### DIFF
--- a/service/controller/v2/arm_templates/route_table_setup.json
+++ b/service/controller/v2/arm_templates/route_table_setup.json
@@ -11,6 +11,13 @@
         "provider": "F80D01C0-7AAC-4440-98F6-5061511962AD"
       }
     },
+    "initialProvisioning": {
+      "type": "string",
+      "defaultValue": "Yes",
+      "metadata": {
+        "description": "Whether the deployment is provisioned the very first time."
+      }
+    },
     "routeTablesAPIVersion": {
       "type": "string",
       "defaultValue": "2016-09-01",
@@ -27,6 +34,7 @@
     {
       "type": "Microsoft.Network/routeTables",
       "name": "[variables('name')]",
+      "condition": "[equals(parameters('initialProvisioning'), 'Yes')]",
       "apiVersion": "[parameters('routeTablesAPIVersion')]",
       "location": "[resourceGroup().location]",
       "tags": {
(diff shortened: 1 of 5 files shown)


service/controller/v2/arm_templates/route_table_setup.json
```diff
@@ -27,6 +34,7 @@
     {
       "type": "Microsoft.Network/routeTables",
       "name": "[variables('name')]",
+      "condition": "[equals(parameters('initialProvisioning'), 'Yes')]",
```
r7vme: Disable route table creation if cluster already provisioned.

Unfortunately there is no better way to prevent this. Azure ARM templates are declarative, so if we do not define any routes in ARM template, ARM will wipe existing routes.

service/controller/v2/resource/deployment/deployment.go
```diff
@@ -70,27 +71,38 @@ func (r Resource) newMainDeployment(obj providerv1alpha1.AzureConfig) (deploymen
...
 
-// convertParameters converts the map into the structure used by the Azure API.
-func convertParameters(inputs map[string]interface{}) map[string]interface{} {
-	params := make(map[string]interface{}, len(inputs))
-	for key, val := range inputs {
-		params[key] = struct {
-			Value interface{}
-		}{
-			Value: val,
+// convertParameters merges the input maps and converts teh result into the
```
tuommaki: Nitpick: Typo `teh` -> `the`. I'm not native English speaker and not that good in grammar, but I'm wondering if `converts the result into structure used by Azure API` would be more correct. Somehow these multiple `the` close to each other doesn't sound right. I might be totally wrong here though :)

xh3b4sd: Fixed the typo and kept the articles. Based in my language understanding this makes sense.

r7vme:
@xh3b4sd thanks for adressing commnt. i'll merge, when CI will finish.

PR #240, opened 2018-06-14
adds e2e tests for scaling

xh3b4sd:


integration/test/scaling/scaling_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+	"testing"
+)
+
+func Test_Scaling_Workers(t *testing.T) {
+	err := s.Test(context.Background())
+	if err != nil {
+		t.Fatalf("%#v", err)
+	}
+}
```
rossf7: Nice! ❤️

xh3b4sd:
So I got the scaling tests running. See https://circleci.com/gh/giantswarm/azure-operator/1467. It is only that I see these tests taking a really long time because Azure is even slower than AWS. I added support for the tested version as well like we have in AWS. Right now the tests for the current version `0.1.0` would constantly fail because it does not support scaling. Since the scaling test works I removed the "ready" test since the cluster coming up is also verified by the scaling test. How should we proceed here now?

Issue #238, opened 2018-06-13
Azure: allow to preserve etcd data while upgrades


r7vme:
At the moment we use VMSS for master. In case VMSS all disks are invicible for user (impliciptly managed by VMSS)

WE have two options:
- still use vmss and use disk attachment feature https://github.com/Azure/azure-sdk-for-go/issues/944
- switch back to regular vms

r7vme:
I've tried to define vmss with pre-created etcd disk and getting error

```
                        "dataDisks": [
                            {
                                "lun": "0",
                                "caching": "ReadWrite",
                                "createOption": "Attach",
                                "managedDisk": {
                                    "storageAccountType": "Standard_LRS",
                                    "id": "[resourceId('Microsoft.Compute/disks', 'EtcdDisk')]"
                                }   
                            }   
                        ]   
```

```
Deployment failed. Correlation ID: 514cc568-9b46-4013-bd42-7d7c7ef96ff1. {
  "error": {
    "code": "InvalidParameter",
    "target": "dataDisk.managedDisk.id",
    "message": "Parameter 'dataDisk.managedDisk.id' is not allowed."
  }
}
```

This is actually aligned with API [here](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachinescalesets#VirtualMachineScaleSetDataDisk). There is only one way to attach pre-created etcd disk by issuing PUT request to particular VM as described [here](https://github.com/Azure/vm-scale-sets/tree/master/preview/disk).

It's really complicate things, so we need to wait until vmss created and then do GET (to grab existing disks) and then update with PUT.

I'm starting to evaluate option number two with regular vm.

r7vme:
With regular VM workflow can be following:
1. Create master VM with attached (externally defined) etcd disk
2. On update. Stop/Delete VM via Azure API call (depending on which operation will be faster)
3. Apply ARM template

As a trick to not depend on existing VM removal (we only need VM to be stopped to use etcd disk) we can use randomly generated master VM name.

cc: @xh3b4sd @kopiczko

xh3b4sd:
How would we manage a multi master setup at best? 

I am leaning towards trying to preserve the VMSS support for masters as well. The implications you described do not sound too complicated to me. I would like to at least try to attach a predefined volume to the master VMSS. 

Should we get Stuart into the discussion here for more MS feedback?

kopiczko:
> In case VMSS all disks are invicible for user

[...]

PR #243, opened 2018-06-15
Parametrize overprovisioning and upgrade mode

r7vme:
- parametrize upgrade mode and overprovisioning ([we can enable "Automatic" mode actually](https://github.com/giantswarm/azure-operator/issues/238#issuecomment-397593016))
- disable overprovisioning for masters

NOTE:
Potentially we want to replace "update" with "reimage" (introduced here https://github.com/giantswarm/azure-operator/commit/426fbb4e469319a430d663a545738b3334a5c8e1 ) as actual upgrade for us done, when vm "reimaged" and not when "upgdaded"


service/controller/v2/arm_templates/vmss.json
```diff
@@ -29,7 +29,7 @@
             "type": "String"
         },
         "vmssSshPublicKey": {
-            "type": "String"
+            "type": "string"
```
r7vme: boyscouting, use lower case everywhere

service/controller/v2/arm_templates/main.json
```diff
@@ -362,12 +362,14 @@
           "vmssOsImageOffer": {"value": "[parameters('masterNodes')[0].osImage.offer]"},
           "vmssOsImageSKU": {"value": "[parameters('masterNodes')[0].osImage.sku]"},
           "vmssOsImageVersion": {"value": "[parameters('masterNodes')[0].osImage.version]"},
+          "vmssOverprovision": {"value": "false"},
```
r7vme: we don't need overprovision for masters as by default it creates two vms and both can be up at the same time, messing up load balancers.

xh3b4sd:
I do not get the full picture here. What does disabling overprovisioning for masters and enabling it for workers do for us now? Is that just an intermediate step or do we fix some problem now?

r7vme:
> I do not get the full picture here. What does disabling overprovisioning for masters and enabling it for workers do for us now? 

Overprovisioning is enabled by default. We just disabling overprovisioning for masters. Imagine that two masters coming up and registered in load balancers with completely different etcd disks. We want to prevent such situations. This also makes master ID predictable, it will always be `0`. Right now we have a race between `0` and `1` and one that will faster report ready status (from waagent) wins. But as i said it's not we actually need.

Keeping overprovisioning for workers makes sense, because it's makes vmss creation with large number of VMs more reliable. Moreover if VM will be able to register in k8s at some point it will be removed by Kubernetes as soon as it will be removed by vmss (when 10/10 vms will be ready, others are removed).

From official docs
> You can overprovision scale set VMs for increased reliability and quicker deployment times. You cannot overprovision individual VMs unless you write custom code to perform this action.

r7vme:
> I do not get the full picture here. 

1. overprovisioning (see comment above) this is the final change - we want to disable over provisioning for master. This is done here.

2. upgrade mode. 

Nothing changed in actual behavious, we still on "Manual" mode, but this change makes upgrade mode configurable. Finally i expect that we will use "Automatic" mode. Is it clear why we want to use "Automatic" mode?

xh3b4sd:
> Is it clear why we want to use "Automatic" mode?

Not really. AFAIK we wanted to use "Manual" to have full control of the update process because we need to hook into it in order to drain nodes. I am not sure right now which implications "Automatic" has.

r7vme:
> Not really. AFAIK we wanted to use "Manual" to have full control of the update process because we need to hook into it in order to drain nodes. I am not sure right now which implications "Automatic" has.

Actually good point. We need to do few more tests before enabling "Automatic". I tested if custom_data changed nothing is happens. Just latest metadata applied to VM (so on next reimage new custom_data) will be used. We also need to test on what happens if vm size or CoreOS image changed.

service/controller/v2/arm_templates/main.json
```diff
@@ -397,9 +399,11 @@
           "vmssOsImageOffer": {"value": "[parameters('workerNodes')[0].osImage.offer]"},
           "vmssOsImageSKU": {"value": "[parameters('workerNodes')[0].osImage.sku]"},
           "vmssOsImageVersion": {"value": "[parameters('workerNodes')[0].osImage.version]"},
+          "vmssOverprovision": {"value": "true"},
           "vmssVmCustomData": {"value": "[parameters('workerCloudConfigData')]"},
           "vmssLbBackendPools": {"value": []},
-          "vmssVnetSubnetId": {"value": "[reference('virtual_network_setup').outputs.workerSubnetID.value]"}
+          "vmssVnetSubnetId": {"value": "[reference('virtual_network_setup').outputs.workerSubnetID.value]"},
+          "vmssUpgradePolicy": {"value": "Manual"}
```
kopiczko: Why this is parametrized?

r7vme: It's just a place holder if we will want to outsource upgrades to Azure. From latest discussions seems that we will do both "upgrade" and "reimage" in operator. So this is not really needed now.

PR #248, opened 2018-06-18
manages updating and reimaging of instances

xh3b4sd:


service/controller/v2/resource/instance/create.go
```diff
@@ -107,17 +135,82 @@ func (r *Resource) ensureDeploymentUpdate(ctx context.Context, customObject prov
...
 		s := deploymentNameFunc(customObject)
-		ids := compute.VirtualMachineScaleSetVMInstanceRequiredIDs{
+		ids := &compute.VirtualMachineScaleSetVMInstanceIDs{
 			InstanceIds: to.StringSlicePtr([]string{
-				instanceID,
+				*instanceToReimage.InstanceID,
 			}),
 		}
-		_, err = c.UpdateInstances(ctx, g, s, ids)
+		_, err = c.Reimage(ctx, g, s, ids)
```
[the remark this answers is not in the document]
r7vme: Another idea. Do by default "update" and do "reimage" only when we detect version bundle change?

Do not have separate loop for "reimage", but do it together (if necessary) with "update".
Smth like that?
```
	//     loop 1: worker 1 update (+ reimage if version change)
	//     loop 2: worker 2 update (+ reimage if version change)
	//     loop 3: worker 3 update (+ reimage if version change)
```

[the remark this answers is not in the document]
xh3b4sd: I think we cannot do the calls after each other because the update itself only to apply the latest model takes quite long. I am not sure how ARM executes or guarantees that shit happens as we want it. We would need to test this but I would also not want to rely on too much magic. When we manage the process with different reconciliation loops and iterate gradually towards the desired state the system design is way more stable and reliable IMO. We can work on some different action scheduling in the future but as we discussed the suggested algorithm would be my first attempt to move forward.

service/controller/v2/resource/deployment/deployment.go
```diff
@@ -53,22 +53,23 @@ func (r Resource) newDeployment(obj providerv1alpha1.AzureConfig, overwrites map
...
-		"workerCloudConfigData":         workerCloudConfig,
+		"masterSubnetCidr":              key.VnetMasterSubnetCIDR(obj),
+		"masterNodes":                   masterNodes,
 		"templatesBaseURI":              baseTemplateURI(r.templateVersion),
+		"versionBundleVersion":          key.VersionBundleVersion(obj),
+		"virtualNetworkCidr":            key.VnetCIDR(obj),
+		"virtualNetworkName":            key.VnetName(obj),
+		"workerCloudConfigData":         workerCloudConfig,
+		"workerNodes":                   workerNodes,
+		"workerSubnetCidr":              key.VnetWorkerSubnetCIDR(obj),
```
kopiczko: I don't see the benefit of sorting those. Previously it was nicely aligned by networking, node configuration, etc. Things in those blocks were sorted.

service/controller/v2/arm_templates/vmss.json
```diff
@@ -11,6 +11,12 @@
                 "provider": "F80D01C0-7AAC-4440-98F6-5061511962AD"
             }
         },
+        "versionBundleVersion": {
+            "type": "string",
+            "metadata": {
+              "description": "The versionbundle version used to reconcile the guest cluster."
+            }
+        },
```
kopiczko: I don't see this passed from the `main.json`.

service/controller/v2/resource/instance/create.go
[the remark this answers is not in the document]
kopiczko: Are the newest tags applied to instances after after calling `update` or after vmss ARM template is deployed?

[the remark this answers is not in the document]
xh3b4sd: > Right now the tags are not updated.

Please read what I wrote. I actually asked a question that I would like to get an answer to.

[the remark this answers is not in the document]
kopiczko: I read it a few times already. I don't understand it fully. Does `Right now` mean before calling `Reimage` or before calling `Update` or something else?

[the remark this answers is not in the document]
xh3b4sd: Well, what is the code doing? We call update, we call reimage. The tags are not updated. What is missing Pawel?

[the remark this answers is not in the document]
kopiczko: Now I understand, tags aren't applied at at all. I understand they are applied to the vmss but not to the instances themselves. Maybe you can use the SDK to make the job done. I wouldn't be surprised if ARM removed the tags after next resync period though.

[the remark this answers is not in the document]
xh3b4sd: Yes, this is the point. We have to manage the tags and I am not sure how. Need to get into this and see what we can do.

service/controller/v2/arm_templates/vmss.json
xh3b4sd: I am not sure what is missing here. How does "passing it" work or look like?

xh3b4sd: Is it just that this block has to be defined the same way in the `main.json`?

service/controller/v2/resource/instance/create.go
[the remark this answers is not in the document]
kopiczko: Can you do reimage request immediately after update request?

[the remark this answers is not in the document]
xh3b4sd: Uncertain. I would like to prevent this actually to let the system settle and have more control over it. I will test this if I have to though.

service/controller/v2/arm_templates/vmss.json
kopiczko: It has to be passed to the child template like other parameters: https://github.com/giantswarm/azure-operator/blob/fcb47d768c8c9a8b7864f91d8e46b67c2de561bc/service/controller/v1/arm_templates/main.json#L351-L365

xh3b4sd: Can you take a look again?

kopiczko: This now looks good 👍

service/controller/v3/resource/instance/create.go
```diff
@@ -2,123 +2,488 @@ package instance
...
+		workerBlobValue, err := updateVersionParameterValue(allWorkerInstances, reimagedWorkerInstance, key.VersionBundleVersion(customObject), parameters[workerBlobKey])
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		fmt.Printf("5: %#v\n", workerBlobValue)
+		params := map[string]interface{}{
+			masterBlobKey: masterBlobValue,
+			workerBlobKey: workerBlobValue,
+		}
+		computedDeployment, err := r.newDeployment(ctx, customObject, params)
```
xh3b4sd: So far I tried to track the node IDs and their version bundle versions in the parameters of the new deployment introduced here. This works for new clusters created with `v3` but does not work with clusters coming from `v2` and being upgraded to `v3`. In this case the logging above causes output like this. 

```
3
4: "{\"0\":\"1.1.0\"}"
3
5: "{\"0\":\"1.1.0\",\"1\":\"1.1.0\",\"2\":\"1.1.0\"}"
```

This is wrong because we come from `v2` so it should be `1.0.0` or at least an empty string but I cannot figure how to do that. We do not have any information if there is a CR being reconciled that got its version updated. We cannot differentiate between the situation of a cluster being created with `v3` or a cluster being updated to `v3` so there is no indicator for the version blob management to get the node IDs and their version bundle versions right. With the current implementation we miss to do the actual update because the version blob now says all is up to date. One solution would be to use a different medium to track the version blob so we do not put it into the new VMSS deployment but somewhere else. Not sure what this somewhere should be then. @kopiczko WDYT?

xh3b4sd: For the future I am thinking that it would be good to bind the version bundle version to the k8s nodes. E.g. via labels. Somebody mentioned this recently. I guess it was @teemow. When the guest cluster itself would be able to state its definite version on a node level then we would have a common ground on all providers and would not need to hack with k8s labels on pods like on KVM and hack CF stack outputs on AWS and hack whatever weird thing on Azure.

kopiczko: I didn't read the code yet, but I like the idea of having that in CR status a lot. That would make the update process visible in the CR. Also has some potential to be generic.

Also maybe we can sue NodeConfig CRD for that?

xh3b4sd: The weirdness with the status is that the operator would write and consume its own status. I am not sold that this is how it should be. I am currently favouring getting the version bound to the actual guest cluster nodes. It feels like this is how it should be. This would be the closest representation of the current state that we can get to. We have the common ground there on the interface which is the k8s guest API. So KVM, Azure and AWS would work the same way. Using the status here does not say fully "YES" to me. Should we do both? Why favouring one or the other way?

kopiczko: I don't see how different storage medium can help here. And probably I don't understand the root problem. When you call `updateVersionParameterValue`, there is no `VersionParamter` then you go to this `if` https://github.com/giantswarm/azure-operator/pull/248/files#diff-ff0780da2c476a174496018b7d37a6f7R418 and the version bundle version you provide is set. But you already know that.

Would adding some code to v2 which sets the versionBlob only, help here? I think it would.

xh3b4sd: > I don't see how different storage medium can help here.

Right now the storage medium is just something in which we write the key-value pair of node ID and version. This information is only available after some reconciliations because the templates are separated and the resources are non-blocking and the VMSS deployment has to be created in order to create nodes and so on. So we only get to know about the right key-value pairs at some point and we need indicators that tell us what kind of truth we have to record. That is why I think even the CR status would not work because it is just another storage medium in that sense and does not provide another indicator. When binding versions to the nodes themselves then the indicator is the node itself. There is nothing that can mislead the mapping between the node ID and its version then. We can just read it all the time and when the node updates its version updates and then we can read the updated information. This would also mean we are way more resilient when reconciliation breaks at some point due to whatever reason as opposed to implicitly guess based on certain information we have available during reconciliation. This is all just hacky. 

> When you call `updateVersionParameterValue`, there is no `VersionParamter` then you go to this `if` https://github.com/giantswarm/azure-operator/pull/248/files#diff-ff0780da2c476a174496018b7d37a6f7R418 and the version bundle version you provide is set. But you already know that.

I am not sure what `VersionParamter` is. I guess you mean the parameter value. The thing is that the implemented mechanism only works across multiple reconciliation loops. When we set the version blob initially the nodes are not updated because there is no information because we are just about to set it. When the version information are set we see that everything is up to date, even on an update that in fact did not get processed yet. And this is the issue. 

> Would adding some code to v2 which sets the versionBlob only, help here? I think it would.

Yes, if only the medium which we use right now is new in `v3`. In `v2` there is no separate VMSS deployment. We would need to use the specific worker and master deployments that are already there. That would work but would be a hack anyway. I would actually like to push the guest cluster node idea and have the versions bound to them in the future. I can now try to go with the dedicated deployments to just get it done and backport `v2` to make it work, but I want the node ID idea to be discussed further because right now I feel like this is the right way to go.

kopiczko: > When binding versions to the nodes themselves

What do you mean by "binding versions to the nodes"?

xh3b4sd: See https://github.com/giantswarm/giantswarm/issues/3532. We can inject the version bundle version with the user data and give it to k8s as labels or something like that.

service/controller/cluster.go
```diff
@@ -64,7 +66,7 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 			Watcher: config.G8sClient.ProviderV1alpha1().AzureConfigs(""),
 
 			RateWait:     informer.DefaultRateWait,
-			ResyncPeriod: informer.DefaultResyncPeriod,
+			ResyncPeriod: 3 * time.Minute,
```
xh3b4sd: I am not sure about this. When I test this makes a lot of sense the the cluster is updated faster. 5 minutes are sometimes super long especially on cluster creation and updates.

service/controller/v2/arm_templates/main.json
```diff
@@ -21,6 +21,12 @@
         "description": "Whether the deployment is provisioned the very first time."
       }
     },
+    "masterVersionBundleVersions":{
+      "type":"object",
+      "metadata":{
+        "description":"JSON blob of key-value pairs mapping master VMSS instance names to their applied version bundle versions."
+      }
+    },
```
xh3b4sd: This is part of the hack I have to backport so we can transport version bundle version information from `v2` to `v3`. I tested this and it looks fine. Existing clusters het the versions applied automatically and once we update them later on the update will be fine because we have the version information prepared. So the trick now is that we set the version blobs to the `master-vmss-deploy` and `worker-vmss-deploy` deployments. These exist already and also exist in `v3`. Only difference is that they are managed in another operatorkit resource now.

service/controller/v2/resource/deployment/resource.go
```diff
@@ -147,3 +194,22 @@ func (r *Resource) getDeploymentsClient() (*azureresource.DeploymentsClient, err
 
 	return azureClients.DeploymentsClient, nil
 }
+
+func (r *Resource) getVMsClient() (*compute.VirtualMachineScaleSetVMsClient, error) {
```
xh3b4sd: One idea I had was that we could have these methods being assigned to `r.azureConfig`. We have the config everywhere and the methods here are repeated all over the repo. This could be nicer.

service/controller/v3/arm_templates/main.json
```diff
@@ -60,32 +60,6 @@
...
-      "metadata": {
-        "description": "Base64-encoded cloud-config data."
-      }
-    },
-    "workerCloudConfigData": {
-      "type": "secureString",
-      "metadata": {
-        "description": "Base64-encoded cloud-config data."
-      }
-    },
```
xh3b4sd: This here is cleanup and setup of the ARM templates because they are managed differently now.

service/controller/v3/controllercontext/context.go
```diff
@@ -0,0 +1,48 @@
...
+type contextKey string
+
+const controllerKey contextKey = "controller"
+
+type Context struct {
+	APILBBackendPoolID  string
+	EtcdLBBackendPoolID string
+	MasterSubnetID      string
+	WorkerSubnetID      string
+}
```
xh3b4sd: We need a controller context now to dispatch information from the `deployment` resource to the `instance` resource.

service/controller/v3/key/key.go
```diff
@@ -200,6 +202,72 @@ func ToCustomObject(v interface{}) (providerv1alpha1.AzureConfig, error) {
 	return customObject, nil
 }
 
+func ToKeyValue(m map[string]interface{}) (interface{}, error) {
```
xh3b4sd: We need a lot of helper methods in order to work against the deployment's version blobs.

service/controller/v3/resource/deployment/deployment.go
```diff
@@ -1,108 +1,41 @@
...
-	templateContentVersion = "1.0.0.0"
-)
-
-func getDeploymentNames() []string {
-	return []string{
-		mainDeploymentName,
-	}
-}
-
-func (r Resource) newDeployment(obj providerv1alpha1.AzureConfig, overwrites map[string]interface{}) (azureresource.Deployment, error) {
```
xh3b4sd: Lots of cleanup here because the template magic is separated. Way better now IMO.

service/controller/v3/resource/deployment/resource.go
```diff
@@ -5,12 +5,13 @@ import (
...
 
 	azureresource "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-02-01/resources"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 
 	"github.com/giantswarm/azure-operator/client"
 	"github.com/giantswarm/azure-operator/service/controller/setting"
-	"github.com/giantswarm/azure-operator/service/controller/v3/cloudconfig"
+	"github.com/giantswarm/azure-operator/service/controller/v3/controllercontext"
```
xh3b4sd: The deployment does not need the `cloudconfig` service anymore because the VMSSs are managed in the `instance` resource. Here we only need the controller context because we need to dispatch some information to the `instance` resource where the VMSSs are managed.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,479 @@ import (
...
 )
 
+const (
+	masterVersionsKey  = "masterVersionBundleVersions"
+	versionsKey        = "versionBundleVersions"
+	vmssDeploymentName = "cluster-vmss-template"
+	workerVersionsKey  = "workerVersionBundleVersions"
+)
+
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
```
xh3b4sd: Here happens all the magic now. The diff is weird. Maybe check the whole file. I think here is nothing we can just split up. This is one thing that works that way.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,479 @@ import (
...
+		return &v
 	}
 
 	return nil
 }
+
+func updateVersionParameterValue(list []compute.VirtualMachineScaleSetVM, reimagedInstance *compute.VirtualMachineScaleSetVM, version string, versionValue map[string]string) (map[string]string, error) {
+	// ignore empty
+	if len(list) == 0 && versionValue == nil {
+		fmt.Printf("1\n")
```
xh3b4sd: I will remove this before merging.

service/controller/v3/resource/instance/create_test.go
```diff
@@ -0,0 +1,546 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+)
+
+func Test_Resource_Instance_findActionableInstance(t *testing.T) {
```
xh3b4sd: Here are quite some tests to ensure certain behaviour.

service/controller/v3/resource/instance/template/main.json
```diff
@@ -0,0 +1,258 @@
+{
+  "$schema":"https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion":"1.0.0.0",
```
xh3b4sd: Here are the separated templates. I ran a formatter against them so this is now consistent. I would like to do this with the other templates in a separate PR as well, because right now it is quite a mess.

service/controller/v3/resource/instance/types.go
```diff
@@ -1,4 +1,4 @@
-package deployment
+package instance
```
xh3b4sd: This file just got moved.

xh3b4sd:
@giantswarm/sre @giantswarm/sig-operator please have another look here.

service/controller/v3/resource/instance/create.go
kopiczko: Yes. That would make total sense.

kopiczko:
Would it be possible to split it to:

- Changing resync period.
- Backport version magic to v2.
- Adding the controller context and assign the necessary information to it.
- _Move VMSS template to instance resource._
- Manage the VMSSs in the instance resource.

?

xh3b4sd:
> Would it be possible to split

Well yes, I suggested that. I will do.

kopiczko:
Yes, I copied your suggestion and added this point: `Move VMSS template to instance resource.` so we don't have to see weird diffs or be distracted by large files being moved to another place.

xh3b4sd:
Ack.

xh3b4sd:
@giantswarm/sig-operator I merged 4 separate PRs into here so the diff got smaller. The rest is basically the finish line. Can you review this here now?

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
-	// Instance IDs are simple non negative integers.
-	//
-	//     4
-	//
-	var instanceID string
-	var instanceName string
+	var masterVersionsValue map[string]string
 	{
-		r.logger.LogCtx(ctx, "level", "debug", "message", "looking for the next instance to be updated")
+		d, err := deploymentsClient.Get(ctx, key.ClusterID(customObject), "master-vmss-deploy")
```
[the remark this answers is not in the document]
xh3b4sd: `key.MasterVMSSName` is for the VMSS name which contains the cluster ID. We do not have anything so far for this AFAIK.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+		if IsDeploymentNotFound(err) {
+			// fall through
+		} else if err != nil {
+			return microerror.Mask(err)
+		} else {
+			p, err := key.ToMap(d.Properties.Parameters)
```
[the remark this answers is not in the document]
xh3b4sd: This here is the `else` branch. We check for the deployment not being found above. In this branch the deployment is found. I cannot be more explicit.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+				}
+			}
+		}
+	}
 
-		c, err := r.getVMsClient()
-		if err != nil {
+	var workerVersionsValue map[string]string
+	{
+		d, err := deploymentsClient.Get(ctx, key.ClusterID(customObject), "worker-vmss-deploy")
```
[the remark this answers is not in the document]
xh3b4sd: Same as with the master. This is right IMO.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+		return nil, microerror.Mask(err)
+	}
+
+	g := key.ResourceGroupName(customObject)
+	s := deploymentNameFunc(customObject)
+	result, err := c.List(ctx, g, s, "", "", "")
+	if IsScaleSetNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the scale set '%s'", deploymentNameFunc(customObject)))
+
+		return nil, microerror.Mask(scaleSetNotFoundError)
```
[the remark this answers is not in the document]
xh3b4sd: Provide a better suggestion. When you check the logs during reconciliation this is very useful. We could move the logging away from here and log where the error is matched against but the result stays the same.

service/controller/v3/resource/instance/create.go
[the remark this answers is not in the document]
kopiczko: Ack. I missed that.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+	var err error
 
-		c, err := r.getScaleSetsClient()
+	instanceInProgress := firstInstanceInProgress(customObject, list)
+	if instanceInProgress != nil {
+		return nil, nil, nil
+	}
+
+	var instanceToUpdate *compute.VirtualMachineScaleSetVM
+	if instanceInProgress == nil {
```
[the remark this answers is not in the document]
xh3b4sd: I restructured the conditions and made it to return early.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+func firstInstanceToReimage(customObject providerv1alpha1.AzureConfig, list []compute.VirtualMachineScaleSetVM, versionValue map[string]string) (*compute.VirtualMachineScaleSetVM, error) {
+	if versionValue == nil {
+		return nil, microerror.Mask(versionBlobEmptyError)
+	}
+
+	for _, v := range list {
+		desiredVersion := key.VersionBundleVersion(customObject)
+		instanceVersion, ok := versionValue[*v.InstanceID]
+		if !ok {
+			return nil, microerror.Mask(versionBlobEmptyError)
```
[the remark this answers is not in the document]
xh3b4sd: I don't follow at all here. This is the way how it works. When we continue the logic breaks. It is also that all this is not really cool anyway and I want to ease the version magic once we have the version information in the guest cluster's k8s nodes. This will be refactored anyway and I would just not invest more time here now.

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+
+		return m, nil
+	}
+
+	// remove missing
+	if len(versionValue) != 0 {
+		fmt.Printf("3\n")
+		m := map[string]string{}
+		for k, v := range versionValue {
+			if !containsInstanceVersion(list, k) {
```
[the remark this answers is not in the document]
xh3b4sd: Right, this was just wrong. `k` is the ID and not the version. I changed the function name and the parameter name.

service/controller/v3/resource/instance/create_test.go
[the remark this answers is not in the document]
xh3b4sd: Please be more specific. I follow a lot of things there. What do you miss?

service/controller/v3/resource/instance/create.go
```diff
@@ -9,116 +9,491 @@ import (
...
+		}
+	}
 
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("instance '%s' is in state '%s'", instanceName, *v.ProvisioningState))
-			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment")
 
-			return nil
+		masterVersionsValue, err := updateVersionParameterValue(allMasterInstances, reimagedMasterInstance, key.VersionBundleVersion(customObject), masterVersionsValue)
```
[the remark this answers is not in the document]
xh3b4sd: I changed the function name as you suggested.

service/controller/v3/resource/instance/create.go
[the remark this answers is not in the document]
kopiczko: > We could move the logging away from here and log where the error is matched against but the result stays the same.

This doesn't smell at all.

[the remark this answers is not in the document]
xh3b4sd: Done.

service/controller/v3/resource/instance/template.go
```diff
@@ -0,0 +1,17 @@
...
+	mainTemplate = "main.json"
+)
+
+func templateURI(version, template string) string {
+	return fmt.Sprintf(templateURIFmt, version, template)
+}
+
+func baseTemplateURI(version string) string {
+	return templateURI(version, "")
+}
```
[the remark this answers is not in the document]
xh3b4sd: I will provide this with the next PRs.

PR #249, opened 2018-06-19
Remove lb backend pool from workers

r7vme:


service/controller/v2/arm_templates/vmss_worker.json
```diff
@@ -0,0 +1,282 @@
...
+                                "name": "[concat(parameters('vmssName'), '-nic')]",
+                                "properties": {
+                                    "enableIPForwarding": true,
+                                    "primary": "true",
+                                    "ipConfigurations": [
+                                        {
+                                            "name": "[concat(parameters('vmssName'), '-ipconfig')]",
+                                            "properties": {
+                                                "subnet": {
+                                                    "id": "[parameters('vmssVnetSubnetId')]"
```
r7vme: Only removed `loadBalancerBackendAddressPools` from vmss_worker.json, except that just a copy of initial vmss.json

service/controller/v2/arm_templates/main.json
```diff
@@ -402,7 +407,6 @@
           "vmssOsImageVersion": {"value": "[parameters('workerNodes')[0].osImage.version]"},
           "vmssOverprovision": {"value": "true"},
           "vmssVmCustomData": {"value": "[parameters('workerCloudConfigData')]"},
-          "vmssLbBackendPools": {"value": []},
```
r7vme: By default all load balancing managed by Kubernetes, so do not pass any backend configuration to workers

r7vme:
Noop, this does not help. I have to wait additional 15 minutes and all Kubernetes changes were wiped. We need alternative solution.

xh3b4sd:
What about the initial provisioning condition?

r7vme:
> What about the initial provisioning condition?

This means that vmss won't be updated at all. This is more or less fine with route table (as i hardly imagine we do many changes there). Not really awesome with security group (because we can change some rules). And totally bad with vmss. We want to update vmss.

I'm thinking on bringing back ingress load balancer, but with name kubernetes. But looks like it won't solve the issue, because load balancer also will be overwritten every time.

service/controller/v2/arm_templates/vmss_worker.json
kopiczko: What happens when you set it to empty string instead?

kopiczko: There are also options to make it conditional, we can talk if necessary. I'm against having another "vmss" json.

r7vme:
@xh3b4sd @kopiczko please review new approach.

Pros:
- it fixes the issue and allows us to reconcile worker vmss

Cons:
- it creates "dummy" resources (dummy public ip and empty frontend configuration)

service/controller/v2/arm_templates/kubernetes_load_balancer_setup.json
```diff
@@ -0,0 +1,46 @@
...
+            "properties": {
+              "publicIPAddress": {
+                "id": "[resourceId('Microsoft.Network/publicIPAddresses', 'dummy-pip')]"
+              }
+            }
+          }
+        ],
+        "backendAddressPools": [
+          {
+            "name": "kubernetes"
```
r7vme: kubernetes expects backend name "kubernetes"

service/controller/v2/arm_templates/kubernetes_load_balancer_setup.json
```diff
@@ -0,0 +1,46 @@
...
+    {
+      "apiVersion": "2017-10-01",
+      "type": "Microsoft.Network/publicIPAddresses",
+      "name": "dummy-pip",
+      "location": "[resourceGroup().location]",
+      "properties": {}
+    },
+    {
+      "apiVersion": "2017-10-01",
+      "name": "kubernetes",
```
r7vme: pre create empty "kubernetes" lb to associate with our workers.

service/controller/v2/arm_templates/kubernetes_load_balancer_setup.json
```diff
@@ -0,0 +1,46 @@
...
+      "type": "Microsoft.Network/loadBalancers",
+      "condition": "[equals(parameters('initialProvisioning'), 'Yes')]",
+      "location": "[resourceGroup().location]",
+      "properties": {
+        "frontendIPConfigurations": [
```
r7vme: frontend configuration is required if we define backend configuration.

service/controller/v2/arm_templates/kubernetes_load_balancer_setup.json
```diff
@@ -0,0 +1,46 @@
...
+      "type": "Microsoft.Network/publicIPAddresses",
+      "name": "dummy-pip",
+      "location": "[resourceGroup().location]",
+      "properties": {}
+    },
+    {
+      "apiVersion": "2017-10-01",
+      "name": "kubernetes",
+      "type": "Microsoft.Network/loadBalancers",
+      "condition": "[equals(parameters('initialProvisioning'), 'Yes')]",
```
r7vme: do not LB reconcile this after initial installation as it will be managed by kubernetes

PR #251, opened 2018-06-19
Fix kubernetes lb dependency

r7vme:
forgot to add dependency. It fails sometimes without.

Initial change https://github.com/giantswarm/azure-operator/pull/249


kopiczko:
Merging because of failing builds.

PR #252, opened 2018-06-20
Changelog

r7vme:


service/controller/v2/arm_templates/vmss.json
```diff
@@ -189,9 +189,7 @@
                 "tier": "Standard",
                 "capacity": "[int(parameters('vmssVmCount'))]"
             },
-            "identity": {
-                "type": "systemAssigned"
-            },
+            "identity": "[if(parameters('vmssMSIEnabled'), json('{\"type\":\"systemAssigned\"}'), json('{\"type\":\"None\"}'))]",
```
kopiczko: In this case I'd like to simplify it to:

```json
"identity": {
    "type": "[if(parameters('vmssMSIEnabled'), 'systemAssigned', 'None')]"
},
```

service/controller/v2/arm_templates/vmss.json
```diff
@@ -283,7 +281,7 @@
             },
             "properties": {
                 "roleDefinitionId": "[variables('contributorRoleDefinitionId')]",
-                "principalId": "[reference(concat(resourceId('Microsoft.Compute/virtualMachineScaleSets/', parameters('vmssName')),'/providers/Microsoft.ManagedIdentity/Identities/default'),'2015-08-31-PREVIEW').principalId]",
+                "principalId": "[if(parameters('vmssMSIEnabled'), reference(concat(resourceId('Microsoft.Compute/virtualMachineScaleSets/', parameters('vmssName')),'/providers/Microsoft.ManagedIdentity/Identities/default'),'2015-08-31-PREVIEW').principalId, '')]",
```
kopiczko: Is it really needed? We have `"condition"` here, so it shouldn't be executed at all.

service/controller/v2/arm_templates/vmss.json
r7vme: Looks better, right

service/controller/v2/arm_templates/vmss.json
r7vme: Oh yeeeah, it still tries to execute `reference` even is resource is disabled :/

PR #261, opened 2018-06-25
tmp

r7vme:


r7vme:
Seems issue is not actual anymore. Closing.

PR #264, opened 2018-06-25
BYOC add service/controller/v3/context

TheoBrigitte:


service/controller/v2/context/context.go
```diff
@@ -0,0 +1,32 @@
+package context
```
kopiczko: This package was renamed to `controllercontext` in aws-operator to avoid ambiguity: https://github.com/giantswarm/aws-operator/tree/master/service/controller/v14/controllercontext

service/controller/v2/context/error.go
```diff
@@ -0,0 +1,12 @@
+package context
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
+var serviceNotFound = microerror.New("service not found")
```
kopiczko: Also error there is `var notFoundError = microerror.New("not found")`:  https://github.com/giantswarm/aws-operator/blob/b80db2bdc3ea8533f4ede4f9df0a639ca1650214/service/controller/v14/controllercontext/error.go#L7

TheoBrigitte:
all good?

PR #266, opened 2018-06-25
BYOC add credential pkg

TheoBrigitte:

### DIFF
--- a/service/controller/v2/credential/credential.go
+++ b/service/controller/v2/credential/credential.go
@@ -1,0 +1,82 @@
+package credential
+
+import (
+	"github.com/giantswarm/microerror"
+	"k8s.io/api/core/v1"
+	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/v2/key"
+)
+
+const (
+	ClientIDKey       = "clientID"
+	ClientSecretKey   = "clientSecret"
+	SubscriptionIDKey = "subscriptionID"
+	TenantIDKey       = "tenantID"
+)
+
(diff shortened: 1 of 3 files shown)


service/controller/v2/credential/credential.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/v2/key"
+)
+
+const (
+	ClientIDKey       = "clientID"
+	ClientSecretKey   = "clientSecret"
+	SubscriptionIDKey = "subscriptionID"
+	TenantIDKey       = "tenantID"
```
kopiczko: I would not export them if it isn't necessary.

kopiczko: Shouldn't they be `azure.azureoperator.clientid`, etc. to match with what we have in aws-operator?

TheoBrigitte:
all good?

service/controller/v3/credential/credential.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+const (
+	ClientIDKey       = "azure.azureoperator.clientid"
+	ClientSecretKey   = "azure.azureoperator.clientsecret"
+	SubscriptionIDKey = "azure.azureoperator.subscriptionid"
+	TenantIDKey       = "azure.azureoperator.tenantid"
```
kopiczko: I still think that should be private, but not a blocker.
Good the values are aligned.

TheoBrigitte: I need them later on in e2e test in order to create the test credential-default
see https://github.com/giantswarm/azure-operator/pull/267/files/88bf173c35a5c1ff30852a12648ff6bf749dd516#diff-e325bfe85d45301943ab7fe56c606b9fR128

kopiczko: LoL I forgot for the second time :)

PR #267, opened 2018-06-25
BYOC update e2e

TheoBrigitte:


integration/template/azure_operator_charts_values.go
```diff
@@ -24,6 +24,7 @@ var AzureOperatorChartValues = `Installation:
           CIDR: "10.0.0.0/16"
           ResourceGroup: "godsmack"
           VirtualNetwork: "godsmack"
+          VirtualNetworkGateway: "GatewayTest"
```
TheoBrigitte: TODO: align with godsmack vpn gateway

helm/azure-operator-resource-chart/templates/cluster.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       cidr: "{{ .Values.azure.cidr }}"
       masterSubnetCIDR: "{{ .Values.azure.masterSubnetCIDR }}"
       workerSubnetCIDR: "{{ .Values.azure.workerSubnetCIDR }}"
+      vpnSubnetCidr: "{{ .Values.azure.vpnSubnetCidr }}"
```
kopiczko: I think we should just remove `examples` directory and `helm/azure-operator-resource-chart` directory.

integration/setup/setup.go
```diff
@@ -92,3 +105,45 @@ func Resources(c *client.AzureClientSet, g *framework.Guest, h *framework.Host)
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: credentialName,
+			},
+
+			// TODO: change to guest subscription credential.
+			Data: map[string][]byte{
+				credential.ClientIDKey:       []byte(c.ClientID),
+				credential.ClientSecretKey:   []byte(c.ClientSecret),
+				credential.SubscriptionIDKey: []byte(c.SubscriptionID),
+				credential.TenantIDKey:       []byte(c.TenantID),
```
kopiczko: I get why they are exported now.

integration/setup/setup.go
```diff
@@ -92,3 +105,45 @@ func Resources(c *client.AzureClientSet, g *framework.Guest, h *framework.Host)
...
 	return nil
 }
+
+func installCredential(h *framework.Host) error {
+	o := func() error {
+		k8sClient := h.K8sClient()
+
+		k8sClient.CoreV1().Secrets(credentialNamespace).Delete(credentialName, &metav1.DeleteOptions{})
+
+		c := env.AzureGuestConfig()
```
kopiczko: Why do we need to create a `client.AzureClientSetConfig ` here? I think this is confusing. It suggests creating a client. I'd just take raw values for the `Data map[string][]byte` from `os.Getenv`.

helm/azure-operator-resource-chart/templates/cluster.yaml
TheoBrigitte: Some cleaning would be needed there, let's do that in a different PR.

integration/template/azure_operator_charts_values.go
TheoBrigitte: @r7vme do you know how will the host cluster VPN Gateway will be named ?

r7vme: https://github.com/giantswarm/giantnetes-terraform/pull/111/files#diff-22469f75baefa6503494b40c9b8b860bR11

godsmack-vpn-gateway

TheoBrigitte:
all good?

integration/setup/setup.go
```diff
@@ -92,3 +105,50 @@ func Resources(c *client.AzureClientSet, g *framework.Guest, h *framework.Host)
...
+		k8sClient := h.K8sClient()
+
+		k8sClient.CoreV1().Secrets(credentialNamespace).Delete(credentialName, &metav1.DeleteOptions{})
+
+		c := client.AzureClientSetConfig{
+			ClientID:       env.AzureGuestClientID(),
+			ClientSecret:   env.AzureGuestClientSecret(),
+			SubscriptionID: env.AzureGuestSubscriptionID(),
+			TenantID:       env.AzureGuestTenantID(),
+		}
```
kopiczko: I just don't follow why should we use `AzureClientSetConfig` here.

You could just do:

```
+				credential.ClientIDKey:       []byte(env.AzureGuestClientID()),
+				credential.ClientSecretKey:   []byte(env.AzureGuestClientSecret()),
+				credential.SubscriptionIDKey: []byte(env.AzureGuestSubscriptionID()),
+				credential.TenantIDKey:       []byte(env.AzureGuestTenantID()),
```

But being in scope of a single method it's now easier to realise that `AzureClientSetConfig` isn't use to create actual client, so it isn't a blocker.

TheoBrigitte: you are right, no need for that. removing this

TheoBrigitte:
Re-opening as this was merged to the wrong branch

PR #268, opened 2018-06-27
Remove examples lab chart

TheoBrigitte:
Follow up on: https://github.com/giantswarm/azure-operator/pull/267#discussion_r197830666

Removing `examples/azure-*lab-chart*` has they get little attention and are therefore mostly outdated.
Along this I also plan to update the examples/README.md to get an up to date guide on how to run azure-operator locally. It would be great if we could share some practices we have there.

@kopiczko you also mentioned removing  `helm/azure-operator-resource-chart` but aren't those still in use somehow ? Beside I do use them for local development.


kopiczko:
@TheoBrigitte I think you should use https://github.com/giantswarm/apiextensions/tree/master/helm/apiextensions-azure-config-e2e-chart instead.

e2e tests use that chart.

xh3b4sd:
We had an issue in master. Got sorted. You need to merged master here so CI can be green. Not doing it for you now due to conflicts. <3

TheoBrigitte:
No progress here since a long time.

PR #273, opened 2018-06-27
adds controller context magic

xh3b4sd:
Towards https://github.com/giantswarm/azure-operator/pull/248.


xh3b4sd:
Fishing for LGTM.

service/controller/v3/key/key.go
```diff
@@ -210,6 +212,72 @@ func ToCustomObject(v interface{}) (providerv1alpha1.AzureConfig, error) {
 	return customObject, nil
 }
 
+func ToKeyValue(m map[string]interface{}) (interface{}, error) {
+	v, ok := m["value"]
+	if !ok {
+		return "", microerror.Maskf(missingOutputValueError, "value")
```
MarcelMue: This reads a bit weird, wouldn't the error be something like "missing output value" and then the variable be "value"? Wouldn't `microerror.Mask` be enough here?

xh3b4sd: Ok cool. Only returning the error then.

PR #274, opened 2018-06-28
uses helper methods from key package when computing deployment

xh3b4sd:


service/controller/v3/resource/deployment/deployment.go
```diff
@@ -21,53 +21,53 @@ func getDeploymentNames() []string {
 	}
 }
 
-func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureConfig, overwrites map[string]interface{}) (azureresource.Deployment, error) {
+func (r Resource) newDeployment(ctx context.Context, customObject providerv1alpha1.AzureConfig, overwrites map[string]interface{}) (azureresource.Deployment, error) {
```
xh3b4sd: We call it `customObject` like everywhere so there is no reason to do it differently here.

service/controller/v3/resource/deployment/deployment.go
```diff
@@ -21,53 +21,53 @@ func getDeploymentNames() []string {
...
+		"clusterID":                     key.ClusterID(customObject),
+		"virtualNetworkName":            key.VnetName(customObject),
+		"virtualNetworkCidr":            key.VnetCIDR(customObject),
+		"calicoSubnetCidr":              key.VnetCalicoSubnetCIDR(customObject),
+		"masterSubnetCidr":              key.VnetMasterSubnetCIDR(customObject),
+		"workerSubnetCidr":              key.VnetWorkerSubnetCIDR(customObject),
 		"masterNodes":                   masterNodes,
 		"workerNodes":                   workerNodes,
-		"dnsZones":                      obj.Spec.Azure.DNSZones,
+		"dnsZones":                      key.DNSZones(customObject),
```
xh3b4sd: There are no helper methods for these things so I just added them.

PR #277, opened 2018-06-28
BYOC subnet computation

TheoBrigitte:

Move network subnets computation in azure-operator to no longer use subnets CIDR from CR.
Using `VirtualNetwork.CIDR` from CR we compute all different subnets we need (calico, master, worker and vpn later on)
So we avoid hijacking into kubernetesd/cluster-operator every time we need a new network.


service/controller/v3/network/network.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/giantswarm/azure-operator/service/controller/setting"
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+const (
+	ipv4Mask = 32
+)
+
+// Compute network subnets within network from CR.
```
xh3b4sd: Proper godoc by convention starts with the thing it explains.

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	vnetCIDR := key.VnetCIDR(azureConfig)
+	_, vnet, err := net.ParseCIDR(vnetCIDR)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return Compute(*vnet, networkSetting)
```
xh3b4sd: I said this already a couple of times. It is really helpful when we do not do these shortcuts because we do error masking and when the stack hides certain locations the stack does not match reality and the mental model of the one who has to debug something breaks and all gets just more complicated.

service/controller/v3/network/subnets.go
```diff
@@ -0,0 +1,29 @@
...
+	return a.Calico.IP.Equal(b.Calico.IP) &&
+		bytes.Equal(a.Calico.Mask, b.Calico.Mask) &&
+		a.Master.IP.Equal(b.Master.IP) &&
+		bytes.Equal(a.Master.Mask, b.Master.Mask) &&
+		a.Parent.IP.Equal(b.Parent.IP) &&
+		bytes.Equal(a.Parent.Mask, b.Parent.Mask) &&
+		a.VPN.IP.Equal(b.VPN.IP) &&
+		bytes.Equal(a.VPN.Mask, b.VPN.Mask) &&
+		a.Worker.IP.Equal(b.Worker.IP) &&
+		bytes.Equal(a.Worker.Mask, b.Worker.Mask)
```
xh3b4sd: I would just write down the conditions. My eyes cannot really parse this.

main.go
```diff
@@ -125,6 +125,10 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.ResourceGroup, "", "Host cluster resource group name.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.VirtualNetwork, "", "Host cluster virtual network name.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Azure.MSI.Enabled, true, "Whether to enabled Managed Service Identity (MSI).")
+	daemonCommand.PersistentFlags().Int(f.Service.Azure.Network.CalicoSubnetMask, 27, "calico network subnet mask.")
```
kopiczko: I think you mean 17 here

main.go
```diff
@@ -125,6 +125,10 @@ func mainError() error {
...
+	daemonCommand.PersistentFlags().Int(f.Service.Azure.Network.MasterSubnetMask, 24, "master network subnet mask.")
+	daemonCommand.PersistentFlags().Int(f.Service.Azure.Network.VPNSubnetMask, 24, "vpn network subnet mask.")
+	daemonCommand.PersistentFlags().Int(f.Service.Azure.Network.WorkerSubnetMask, 24, "worker network subnet mask.")
```
kopiczko: In general I don't like the design with the flags.

We have now following information in the CR

```
    virtualNetwork:
      calicoSubnetCIDR: 10.1.128.0/17
      cidr: 10.1.0.0/16
      masterSubnetCIDR: 10.1.0.0/24
      workerSubnetCIDR: 10.1.1.0/24
```

After this is merged we should pass only:

```
    virtualNetwork:
      cidr: 10.1.0.0/16
```

We shouldn't rely on fixed mask numbers here.

We always should take second half for calico (as kubernetesd does now).
Then we should assert that mask for `virtualNetwork.cidr` is > 22. And assign `/24` subnets for master, worker and VPN.

Does that make sense? If not we can have a call.

kopiczko:
@xh3b4sd it computes only the subnets. Network allocation happens outside. I think it's up to operator what it does with the allocated network, but I think this information is useful to have in the status.

TheoBrigitte:
You are right `aws-operator` is not computing network ranges on its own, but it should.

IMO it does not feels right to have a different service computing the subnets used by guest clusters.
This should belong to the operator as it is the entity which defines what are the different components required to run the guest clusters and how they operate all together; network is also part of this.
I am fine with having cluster-operator/kubernetesd defining the guest cluster network, but knowledge about how this network is then split into subnets should belong to operator.

It also decouples things and make easier to add new subnets in the future, as we would only need to make changes within operator. Currently for every new subnet we add we also need to make those changes in cluster-operator/kubernetesd. I get that we do not do that every day, but its a little improvement.

xh3b4sd:
> but knowledge about how this network is then split into subnets should belong to operator.

That makes sense. I am just not into how the network is split up even further and what it is actually used for but when this is how it is then this makes sense. This should then be the same in the end in AWS, if there is even some similar need.

xh3b4sd:
> but I think this information is useful to have in the status.

This is a good candidate for status I guess.

TheoBrigitte:
>> but I think this information is useful to have in the status.

> This is a good candidate for status I guess.

I guess `network` package should be converted to a resource and store the different subnets in status then

main.go
TheoBrigitte: so instead of calico /17, we should use `ipam.Half`

for the rest my intuition would tell me, let's ipam do the hard stuff and fails if it has too, there not much we can do there if the provided network is too small.

kopiczko: Yes is mask is <22 this is unrecoverable.

service/controller/v3/network/network.go
TheoBrigitte: fixed, sorry I was not 100% sure about this one.

service/controller/v3/network/subnets.go
TheoBrigitte: done

service/controller/v3/network/network.go
TheoBrigitte: done

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/giantswarm/ipam"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/service/controller/setting"
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+const (
+	ipv4MaskSize = 32
```
kopiczko: I think this should be 24

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	vpnSubnetMask := net.CIDRMask(subnetsSetting.VPNSubnetMask, ipv4MaskSize)
+	subnets.VPN, err = ipam.Free(network, vpnSubnetMask, []net.IPNet{subnets.Master, subnets.Worker})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	calicoSubnetMask := net.CIDRMask(subnetsSetting.CalicoSubnetMask, ipv4MaskSize)
+	subnets.Calico, err = ipam.Free(network, calicoSubnetMask, []net.IPNet{subnets.Master, subnets.Worker, subnets.VPN})
```
kopiczko: I think this should be second half (`ipam.Half`)

service/controller/setting/types.go
```diff
@@ -53,3 +54,10 @@ type OIDC struct {
 	UsernameClaim string
 	GroupsClaim   string
 }
+
+type AzureNetwork struct {
+	CalicoSubnetMask int
+	MasterSubnetMask int
+	VPNSubnetMask    int
+	WorkerSubnetMask int
+}
```
kopiczko: I don't think it's needed anymore.

service/controller/v3/network/network.go
TheoBrigitte: No, this is the length in bytes of the CIDRMask which is 4*8 for IPv4

service/controller/setting/types.go
TheoBrigitte: Why ? This is actually used by `network.Compute` to determine the size of subnets

kopiczko: For the record we discussed those changes in a call. Changes incoming.

service/controller/v3/network/network.go
TheoBrigitte: done

service/controller/setting/types.go
TheoBrigitte: done

dropped flags and using constants now

TheoBrigitte:
all good ?

service/controller/v3/resource_set.go
```diff
@@ -299,6 +300,11 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	}
 
 	initCtxFunc := func(ctx context.Context, obj interface{}) (context.Context, error) {
+		subnets, err := network.ComputeFromCR(ctx, obj)
```
kopiczko: I'd preffer asserting `obj` type here before calling `network.ComputeFromCR` and making `network.ComputeFromCR` argument strongly typed.

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	return subnets, nil
+}
+
+// Compute computes subnets within network.
+//
+// subnets computation rely on ipam.Free and use ipv4MaskSize as IPMask length.
+func Compute(network net.IPNet) (subnets *Subnets, err error) {
```
kopiczko: We don't need that public, do we?

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	masterSubnetMask = 24
+	workerSubnetMask = 24
+	vpnSubnetMask    = 24
+
+	ipv4MaskSize = 32
+)
+
+// ComputeFromCR computes subnets using network found in CR.
+func ComputeFromCR(ctx context.Context, obj interface{}) (*Subnets, error) {
```
kopiczko: I would just call it `FromCR`. Package name indicates it will be `network`

service/controller/v3/network/network.go
```diff
@@ -0,0 +1,74 @@
...
+	subnets.Worker, err = ipam.Free(network, workerCIDRMask, []net.IPNet{subnets.Calico, subnets.Master})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	vpnCIDRMask := net.CIDRMask(vpnSubnetMask, ipv4MaskSize)
+	subnets.VPN, err = ipam.Free(network, vpnCIDRMask, []net.IPNet{subnets.Calico, subnets.Master, subnets.Worker})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
kopiczko: This is very nice now.

integration/network/network.go
```diff
@@ -4,61 +4,18 @@ import (
...
-	MasterSubnetCIDR string
-	WorkerSubnetCIDR string
-	CalicoSubnetCIDR string
-}
-
-func ComputeCIDR(buildNumber uint) (*azureCIDR, error) {
-	cidrs := new(azureCIDR)
-
+func ComputeSubnets(buildNumber uint) (*network.Subnets, error) {
 	azureNetwork := determineSubnet(e2eNetwork, e2eSubnetQuantity, buildNumber)
```
kopiczko: I'd say we don't need to extract `determineSubnet` anymore :)

service/controller/v3/resource_set.go
TheoBrigitte: done

service/controller/v3/network/network.go
TheoBrigitte: This is used in e2e test

https://github.com/giantswarm/azure-operator/pull/277/files#diff-8864153a0b634472800453cb4d1ad300L61

service/controller/v3/network/network.go
TheoBrigitte: Any idea to rename `Compute` then ?

service/controller/v3/network/network.go
kopiczko: Ack.

service/controller/v3/network/network.go
kopiczko: I see. Given https://github.com/giantswarm/azure-operator/pull/277/files#r199463805 this name makes sense.

integration/network/network.go
TheoBrigitte: We still need this to get e2e test network

service/controller/v3/network/subnets.go
xh3b4sd: This is not what I meant. I meant this. 

```
if a.Calico.IP  != b.Calico.IP {
  return false
}

if a.Master.IP  != b.Master.IP {
  return false
}

...

return true
```

PR #278, opened 2018-06-28
BYOC vpngateway

TheoBrigitte:

Replace vnet peering with vpn gateway, in order to connect host and guest cluster.
This is a requirement for BYOC since vnet peering does not work across different subscriptions owned by different tenants.


service/controller/v3/resource/vpngateway/update.go
```diff
@@ -0,0 +1,172 @@
...
+// needUpdate determine if current needs to be updated in order to comply with
+// desired. Following properties are examined:
+//
+//     Name
+//     VirtualNetworkGatewayConnectionPropertiesFormat.VirtualNetworkGateway1.ID
+//     VirtualNetworkGatewayConnectionPropertiesFormat.VirtualNetworkGateway2.ID
+//     VirtualNetworkGatewayConnectionPropertiesFormat.ConnectionType
+//     VirtualNetworkGatewayConnectionPropertiesFormat.ConnectionStatus
+//
+func needUpdate(current, desired network.VirtualNetworkGatewayConnection) bool {
```
xh3b4sd: This looks like it needs proper unit testing.

service/controller/v3/resource/vpngateway/update.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return nil
+}
+
+func (r *Resource) applyUpdateChange(ctx context.Context, azureConfig providerv1alpha1.AzureConfig, change connections) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensure vpn gateway connections")
+
+	if change.isEmpty() {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensure vpn gateway connections: already ensured")
```
xh3b4sd: As I mentioned this already a couple of times this message format is not what we use everywhere. It got introduced in the azure operator at some point for whatever reason but it is not really helping us IMO. I would like to stick to the messages as we have them everywhere else.

TheoBrigitte:
> I was wondering how we can ensure that this change does not break anything. Do we already cover the important bits and pieces with the integration tests we have or do we still have to write more?

Current e2e test cover important parts here, since it ensures guest cluster is ready which now includes:
* setup vpn gateway
* establishing vpn connections

I believe this is enough when Azure says the connection has been established.

> What I also do not like here is that this PR is again super big.

I do not see how I could have split it into meaningful smaller PRs. Any suggestions on this ?

service/controller/v3/resource/vpngateway/update.go
TheoBrigitte: done

service/controller/v3/resource/vpngateway/update.go
TheoBrigitte: done

https://github.com/giantswarm/azure-operator/pull/278/commits/ddbf286226a3146b429bfe48ffcb7ea5c341e7cb

service/controller/v3/resource/vpngateway/update_test.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+func TestNeedUpdate(t *testing.T) {
+	testCases := []struct {
+		description      string
+		current, desired network.VirtualNetworkGatewayConnection
+		expected         bool
+	}{
+		{
+			"desired state empty",
```
xh3b4sd: Please just initialize the struct with the keys being visible. This is by no means useful. Lets not be lazy on the wrong place. You cannot easily maintain nor extend this. You cannot read it. Tomorrow somebody else has to work with this and has no idea what is going on and when the order of the struct properties changes all this here is immediately fucked and we all just waste a lot of time.

xh3b4sd:
> I do not see how I could have split it into meaningful smaller PRs. Any suggestions on this ?

Something like this for sure. 

- flags
- e2e tests
- removal of vnetpeering resource
- vpngateway resource structure
- vpngateway resource functionality

xh3b4sd:
> Current e2e test cover important parts here

Ok, but we have nothing covered for updates in general. So we still have to manually ensure that `1.0.0` clusters are upgradable to `1.1.0` and I am curious how this will work out. When we change the networking the whole cluster gets fucked somehow IMO and when there is no graceful and automated migration we have a breaking change and then this has to be discussed and communicated with the teams and SIG customer. So is this seamless or a breaking change?

service/controller/v3/resource/vpngateway/update_test.go
TheoBrigitte: done

TheoBrigitte:
IMO upgrade from `1.0.0` to `1.1.0` should work, new vpn gateway and connections will be created.
But vnet peering would be left there as we deploy arm template in incremental mode. Would this be considered as breaking change ?

xh3b4sd:
> IMO upgrade from `1.0.0` to `1.1.0` should work

Lets please simply test this manually. Lets just make sure what we push doesn't break anything. 

> But vnet peering would be left there as we deploy arm template in incremental mode. Would this be considered as breaking change ?

When the network transition works automatically then this is fine with me but I would like to get some @giantswarm/sre feedback here as well because I cannot imagine this is just all fine. The resources we leave behind have to automatically cleaned up though so we do not create orphaned resources.

r7vme:
@TheoBrigitte please make sure to test upgrade path from "peered" cluster to "vpn" cluster.

Another thing: Feeling bad that we are getting rid of peering. But i've raised this already few weeks ago.

TheoBrigitte:
I successfully migrated from `1.0.0` to `1.1.0`.
* vpn gateway was created
* vpn connections were established
* ssh from host cluster to guest works (with vnet peering removed beforehand)

vnet peering resource remained there, as expected.

When all tests passes I would like to go ahead and proceed with merging this PR and tackle removal of the vnet peering resource in a different issue.

xh3b4sd:
Can we write an e2e test to ensure automatically and continuously what you manually verified? Would that make sense and help us down the road?

TheoBrigitte:
Yes it does make sense to have a migration e2e, which ensure we can actually migrate from the previous version to the current wip.

xh3b4sd:
I rather thought about a test that is useful in general and not only for the migration. When the migration is tested now then we do not need a test again IMO. What we need is that the general network availability and reliability is ensured by e2e tests and based on what happened here in terms of code changes. I am just not sure what we additionally have to test. And since we have more issues with QA in general I just want to be more pushy here and rather ask a couple of times too often than once not often enough.

TheoBrigitte:
I though migration test could be useful here so we always test that migration is correctly handled from one version to the other. It would make our life easier by catching omissions early.
Regarding network there is not much to be tested there. Azure arm templates guarantee us that after deployment succeeded we have a working vnet and vpn gateway.
Do you mean here testing that the connection works and we can reach guest cluster from host ?

r7vme:
> Do you mean here testing that the connection works and we can reach guest cluster from host ?

It will be tricky to test, because we are running code from CircleCI and in general don't have access to hsot cluster (ghost in that case). Alternatively we can fetch the status of Azure VPN connection. 

with cli it can be done like that
```
 az network vpn-connection show --resource-group ghost --name ghost-vpn-connection | jq .connectionStatus
"Connected"
```

TheoBrigitte:
Yes, testing the ssh connection would be indeed tricky and does not make so much sense.
What we want here is to ensure the connection is established and that's what the vpngateway resource does here:

https://github.com/giantswarm/azure-operator/pull/278/files#diff-a1b9f4769170217186a28a0021707417R109

It looks simple, but here are the possible connection status and action to take:
* `Connected`: all good nothing to do
* `Connecting`: transition state to `Connected` or `NotConnected`
* `NotConnected`: connection is re-issued
* `Unknown`: same as `Connecting`

https://godoc.org/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network#VirtualNetworkGatewayConnectionStatus

service/controller/v3/arm_templates/virtual_network_gateway.json
```diff
@@ -0,0 +1,103 @@
...
+        "description": "API version used by the Microsoft.Network/virtualNetworkGateways resource."
+      }
+    },
+    "publicIPAddressesAPIVersion": {
+      "type": "string",
+      "defaultValue": "2017-10-01",
+      "metadata": {
+        "description": "API version used by the Microsoft.Network/publicIPAddresses resource."
+      }
+    }
```
kopiczko: `*APIVersion` should be variables instaed of parameters. I'd say the same for the `vpnSubnetName`.

service/controller/v3/arm_templates/virtual_network_setup.json
```diff
@@ -44,7 +47,9 @@
     "masterSubnetID": "[concat(variables('virtualNetworkID'), '/subnets/', variables('masterSubnetName'))]",
 
     "workerSubnetName": "[concat(parameters('virtualNetworkName'), '-WorkerSubnet')]",
-    "workerSubnetID": "[concat(variables('virtualNetworkID'), '/subnets/', variables('workerSubnetName'))]"
+    "workerSubnetID": "[concat(variables('virtualNetworkID'), '/subnets/', variables('workerSubnetName'))]",
+
+    "vpnSubnetName": "GatewaySubnet",
```
kopiczko: I think we should call it `vnetGatewaySubnetName`. `vpn` only happens to be the type of the subnet if I understand correctly.

service/controller/v3/key/key.go
```diff
@@ -312,6 +313,15 @@ func VNetID(customObject providerv1alpha1.AzureConfig, subscriptionID string) st
 	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s", subscriptionID, ResourceGroupName(customObject), VnetName(customObject))
 }
 
+// VPNGatewayName returns name of the virtual network gateway.
+func VPNGatewayName(customObject providerv1alpha1.AzureConfig) string {
+	return fmt.Sprintf("%s-%s", ClusterID(customObject), vpnGatewaySuffix)
+}
+
+func VPNGatewayID(subscriptionID, resourceGroupName, resourceName string) string {
```
kopiczko: Is it used anywhere?

service/controller/v3/resource/vpngateway/delete.go
```diff
@@ -61,27 +62,34 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, azureConfig, change in
...
+	connectionName := *change.Host.Name
+
+	respFuture, err := hostGatewayConnectionClient.Delete(ctx, resourceGroup, connectionName)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	res, err := vnetPeeringClient.DeleteResponder(respFuture.Response())
+	res, err := hostGatewayConnectionClient.DeleteResponder(respFuture.Response())
 	if client.ResponseWasNotFound(res) {
```
kopiczko: Why is it actually checked? I mean if there is a response when we don't care if there isn't.

service/controller/v3/resource/vpngateway/desired.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			if provisioningState := *hostVPNGateway.ProvisioningState; provisioningState != "Succeeded" {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("host vpn gateway is in state '%s'", provisioningState))
+				resourcecanceledcontext.SetCanceled(ctx)
+				r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+				return connections{}, nil
+			}
+		}
+	}
```
kopiczko: Isn't GetCurrentState better place for that? We shouldn't exercise the Azure API anymore if those two guys aren't ready yet.

service/controller/v3/resource/vpngateway/helper.go
```diff
@@ -0,0 +1,33 @@
...
+
+	return vpnGatewayConnections, nil
+}
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890"
+
+func randStringBytes(n int) string {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
```
kopiczko: Are you sure it's initialized? https://golang.org/pkg/math/rand/#Seed

Please grep the code and check. If not lets do it in some meaningful file like main.go

service/controller/v3/resource/vpngateway/update.go
kopiczko: Why do we even use `needUpdate` when we can call `CreateOrUpdate` and it does the heavy lifting for us.

service/controller/v3/resource/vpngateway/desired.go
```diff
@@ -0,0 +1,125 @@
+package vpngateway
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
```
kopiczko: There is `services/network/mgmt/2018-05-01/network` I think we should use the newest.

service/controller/v3/arm_templates/virtual_network_gateway.json
TheoBrigitte: done

service/controller/v3/arm_templates/virtual_network_setup.json
TheoBrigitte: yes correct

service/controller/v3/key/key.go
TheoBrigitte: no it isnt. good catch

service/controller/v3/resource/vpngateway/delete.go
TheoBrigitte: done

service/controller/v3/resource/vpngateway/desired.go
TheoBrigitte: Yes that was my initial design. But connection need reference to gateway objects, so it makes more sense to have them here in the end.

service/controller/v3/resource/vpngateway/helper.go
TheoBrigitte: done

service/controller/v3/resource/vpngateway/update.go
TheoBrigitte: The connection shared key is different on every resync period, I am not sure how this will behave when updating the shared key on every resync period. But IMO it is less hassle not to update when not needed.

service/controller/v3/resource/vpngateway/desired.go
TheoBrigitte: This imply updating it in all different controller version as they are all bound together by `client` package. There is also some minor changes on peering that need to be addressed.

We can do this in a different PR

service/controller/v3/resource/vpngateway/helper.go
TheoBrigitte: https://github.com/giantswarm/azure-operator/pull/278/commits/a4f92645e87f4c41b59536fe66a3a9cffba62b26

main.go
```diff
@@ -37,6 +39,8 @@ func main() {
 func mainError() error {
 	var err error
 
+	rand.Seed(time.Now().UTC().UnixNano())
```
kopiczko: I think `func init()` is a better place for such stuff. I would search for it there to make sure if rand was inited.

service/controller/v3/arm_templates/virtual_network_gateway.json
```diff
@@ -0,0 +1,89 @@
...
+    },
+    "vpnGatewayName": {
+      "type": "string"
+    }
+  },
+  "variables": {
+    "publicIPName": "[concat(parameters('clusterID'), '-VPNGateway-PublicIP')]",
+    "publicIPID": "[resourceId('Microsoft.Network/publicIPAddresses', variables('publicIPName'))]",
+
+    "vnetGatewaySubnetName": "GatewaySubnet",
```
kopiczko: We rely on this name in two places. I think we should make it a variable in main.json and pass it to both virtual_network_gateway.json and virtual_network_setup.json. Then we are DRY.

main.go
TheoBrigitte: did not found any other `rand.Seed` so I moved it to `init` func

service/controller/v3/arm_templates/virtual_network_gateway.json
TheoBrigitte: done

TheoBrigitte:
Fishing for more green lights

PR #281, opened 2018-07-03
draining

xh3b4sd:


helm/azure-operator-chart/templates/rbac.yaml
```diff
@@ -9,6 +9,12 @@ rules:
       - customresourcedefinitions
     verbs:
       - "*"
+  - apiGroups:
+      - core.giantswarm.io
+    resources:
+      - drainerconfigs
+    verbs:
+      - "*"
```
xh3b4sd: We drain guest cluster nodes on updates now. So we have to allow the management of drainer configs.

integration/setup/setup.go
```diff
@@ -75,12 +75,15 @@ func Resources(c *client.AzureClientSet, g *framework.Guest, h *framework.Host)
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
 		err = h.InstallStableOperator("cert-operator", "certconfig", e2etemplates.CertOperatorChartValues)
 		if err != nil {
 			return microerror.Mask(err)
 		}
-
-		err = h.InstallBranchOperator("azure-operator", "azureconfig", template.AzureOperatorChartValues)
+		err = h.InstallStableOperator("node-operator", "drainerconfig", e2etemplates.NodeOperatorChartValues)
```
xh3b4sd: The `node-operator` has to be involved now, even though we do not have it integrated for scaling.

integration/teardown/teardown.go
```diff
@@ -32,6 +32,10 @@ func Teardown(c *client.AzureClientSet, g *framework.Guest, h *framework.Host) e
 		if err != nil {
 			return microerror.Mask(err)
 		}
+		err = framework.HelmCmd("delete node-operator --purge")
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
xh3b4sd: We have to refactor all of this still to not use the shell out magic anymore. We have to boyscout this separately.

service/controller/v3/resource/deployment/resource.go
```diff
@@ -97,10 +97,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	} else {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", *d.Properties.ProvisioningState))
 
-		if !key.IsFinalProvisioningState(*d.Properties.ProvisioningState) {
+		if !key.IsSucceededProvisioningState(*d.Properties.ProvisioningState) {
 			r.debugger.LogFailedDeployment(ctx, d)
+		}
+		if !key.IsFinalProvisioningState(*d.Properties.ProvisioningState) {
```
xh3b4sd: Improving the debugging here.

service/controller/v3/resource/instance/error.go
```diff
@@ -38,6 +38,13 @@ func IsDeploymentNotFound(err error) bool {
 	return false
 }
 
+var missingLabelError = microerror.New("missing label")
```
xh3b4sd: When draining is sorted I go after the error migration in this repo.

service/controller/v3/resource/instance/template/main.json
```diff
@@ -41,12 +41,6 @@
         "description":"Output value of the master subnet ID as referenced from the virtual network setup."
       }
     },
-    "masterVersionBundleVersions":{
-      "type":"object",
-      "metadata":{
-        "description":"JSON blob of key-value pairs mapping master VMSS instance names to their applied version bundle versions."
-      }
-    },
```
xh3b4sd: Since we rely on the guest API information we do not need to carry these information here anymore.

integration/setup/setup.go
kopiczko: What is `"drainerconfig"` or `"azureconfig"`?

xh3b4sd: This is the thing identifying the resource we expect and wait for when setting up the operator. Like the CRD it ensures on boot. Nothing new though.

xh3b4sd:
Can you ask more specific questions? This here is to provide draining for updates in order to make updates on Azure graceful. When you follow the issue above you get that information. I am not sure what I should add here now from the top of my head.

integration/setup/setup.go
kopiczko: Ack. Yes nothing new, but I just noticed and ti looked weird. This will probably have to change when there is an operator that reconciles two or more CRs. But unrelated.

kopiczko:
I would expect something like:

```
prepare draining foundation

- Inject guestCluster to the instance resource.
- Install node-operator in integration test setup.
```

Just to make it easier to review. Those changes seem to be a little bit unrelated and difficult to connect w/o having full picture.

kopiczko:
> I usually comment on my own PRs within the code to explain a lot

Yes I notice and appreciate that. But this one here is an exception.

service/controller/v3/resource/instance/create.go
```diff
@@ -248,58 +246,141 @@ func (r *Resource) allInstances(ctx context.Context, customObject providerv1alph
...
 		if instanceToReimage != nil {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be reimaged", instanceNameFunc(customObject, *instanceToReimage.InstanceID)))
 		}
 	}
 
-	return instanceToUpdate, instanceToReimage, nil
+	return instanceToUpdate, instanceToDrain, instanceToReimage, nil
 }
 
 func (r *Resource) reimageInstance(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance *compute.VirtualMachineScaleSetVM, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) string) error {
```
kopiczko: Unrelated: Isn't it just simpler to pass `deploymentName string` instead of `deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string`?

service/controller/v3/resource/instance/create.go
```diff
@@ -372,34 +453,38 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
 
-	instanceInProgress := firstInstanceInProgress(customObject, list)
+	instanceInProgress := firstInstanceInProgress(customObject, instances)
 	if instanceInProgress != nil {
-		return nil, nil, nil
+		return nil, nil, nil, nil
 	}
 
 	var instanceToUpdate *compute.VirtualMachineScaleSetVM
 	if instanceInProgress == nil {
```
kopiczko: Looks like this is always true here.

service/controller/v3/resource/instance/create.go
```diff
@@ -372,34 +453,38 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
-		instanceToUpdate = firstInstanceToUpdate(customObject, list)
+		instanceToUpdate = firstInstanceToUpdate(customObject, instances)
 		if instanceToUpdate != nil {
-			return instanceToUpdate, nil, nil
+			return instanceToUpdate, nil, nil, nil
 		}
 	}
 
 	var instanceToReimage *compute.VirtualMachineScaleSetVM
 	if instanceToUpdate == nil {
```
kopiczko: Looks like this is always true here.

service/controller/v3/resource/instance/create.go
```diff
@@ -61,71 +49,75 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			return microerror.Mask(err)
+		}
+
+		nodeConfigs = list.Items
+	}
+
 	var allMasterInstances []compute.VirtualMachineScaleSetVM
+	var drainedMasterInstance *compute.VirtualMachineScaleSetVM
 	var reimagedMasterInstance *compute.VirtualMachineScaleSetVM
 	var updatedMasterInstance *compute.VirtualMachineScaleSetVM
```
kopiczko: There is no need to declare those vars outside the block. This gives notion they will be used further.

Also I think those names should be `masterInstanceToDrain` etc.

Same applies to worker.

service/controller/v3/resource/instance/create.go
```diff
@@ -6,18 +6,20 @@ import (
```
kopiczko: Github decided to hide this file diff. It basically went unreviewed. This is why I was asking what this PR is about. After you messaged in #news-dev I checked again and found that.

It's shame.

I reviewed it again. Nothing but minor comments left. Safe to ignore.

service/controller/v3/resource/instance/create.go
xh3b4sd: Maybe. I decided against it because the calling side would be more bloated and ugly.

service/controller/v3/resource/instance/create.go
xh3b4sd: True. Will provide a PR.

service/controller/v3/resource/instance/create.go
xh3b4sd: This derailed a bit from the prototyping face. This was all a bit different in the first place and now it is not needed that way anymore. Will provide a PR.

service/controller/v3/resource/instance/create.go
xh3b4sd: Wow. :D

PR #287, opened 2018-07-11
reconcile services

calvix:
As the annotation is now needed for guest cluster monitoring we need to be able to update old cluster and the that means enable reconciliation of the service to desired state

Also since the etcd port is different across providers and it needs to be part of annotation


service/controller/v3/resource/service/resource.go
```diff
@@ -44,6 +45,32 @@ func (r *Resource) Name() string {
...
+	for _, l := range list {
+		if l.Name == name {
+			return l, nil
+		}
+	}
+
+	return nil, microerror.Mask(notFoundError)
+}
+
+func isServiceModified(a, b *corev1.Service) bool {
```
xh3b4sd: We definitely need unit tests for this.

service/controller/v3/resource/service/update.go
```diff
@@ -3,11 +3,29 @@ package service
...
+
+		_, err := r.k8sClient.CoreV1().Services(serviceToUpdate.Namespace).Update(serviceToUpdate)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "updated services")
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "no need to update services")
+	}
```
xh3b4sd: Well, why wasn't this already implemented? Anything else we do not properly reconcile? @kopiczko @TheoBrigitte

kopiczko: It was some copy paste by mr @JosephSalisbury

service/controller/v3/resource/service/resource.go
calvix: ack

service/controller/v3/resource/service/update.go
calvix: yeah its same in all operators, i already have prepared PRs there a well

calvix:
RFR, tested  and it is looking fine

service/controller/v3/resource/service/resource_test.go
```diff
@@ -0,0 +1,445 @@
...
+							Protocol:   corev1.ProtocolTCP,
+							Port:       httpsPort,
+							TargetPort: intstr.FromInt(httpsPort),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "case 1: wrong type",
```
kopiczko: Please add this is because a pointer is expected. It took mea while to realize.

PR #289, opened 2018-07-12
enable byoc

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/2652

Enable BYOC by making use of guest cluster credentials

`credential-default` secret are in places in every azure installation (godsmack, gollum, ghost, tarantula, victory) and in e2e tests.

### DIFF
--- a/service/controller/v3/resource_set.go
+++ b/service/controller/v3/resource_set.go
@@ -19,6 +19,7 @@ import (
 	"github.com/giantswarm/azure-operator/service/controller/setting"
 	"github.com/giantswarm/azure-operator/service/controller/v3/cloudconfig"
 	"github.com/giantswarm/azure-operator/service/controller/v3/controllercontext"
+	"github.com/giantswarm/azure-operator/service/controller/v3/credential"
 	"github.com/giantswarm/azure-operator/service/controller/v3/debugger"
 	"github.com/giantswarm/azure-operator/service/controller/v3/key"
 	"github.com/giantswarm/azure-operator/service/controller/v3/network"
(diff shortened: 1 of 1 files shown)


tuommaki:
> Enable BYOC by making use of guest cluster credentials

For my curiosity: When this release gets active, does this mean that on some installations we need to occasionally manually manage DNS entries like in AWS, or is that only in AWS?

kopiczko:
AWS only. Currently I'm working towards making AWS DNS design to be on par with the one we have in Azure.

service/controller/v3/resource_set.go
```diff
@@ -324,7 +325,12 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return nil, microerror.Mask(err)
 		}
 
-		azureClients, err := client.NewAzureClientSet(config.HostAzureConfig)
+		guestAzureConfig, err := credential.GetAzureConfig(config.K8sClient, obj)
```
xh3b4sd: Isn't the name here a bit misleading? Why is this a guest config? This is not about config of an guest cluster, but about some other Azure subscription, right?

TheoBrigitte: Name is somehow misleading but this is aligned with `HostAzureConfig`, those are in fact credential for the Azure API.

I would happily tackle this in a different PR, but I need good naming first. How about `guestAzureCredential` and `hostAzureCredential` ?

xh3b4sd: Ah yeah, naming man. No idea.

kopiczko: The returned type is `*client.AzureClientSetConfig`. Pretty clear IMO. If the name was `guestAzureClientSetConfig` I think it would be pretty self explanatory.

TheoBrigitte: you said it naming ... ok I'll do a bunch of renaming in a different PR then

PR #291, opened 2018-07-12
add guest cluster role definition

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/3675

We decided to store roles definitions in plain json in order to be able to build tooling to easily deploy them.
This also come in handy when we later want to update the policies needed by the operator.


policies/guest.json
```diff
@@ -0,0 +1,13 @@
...
+  "Name": "azure-operator",
+  "Description": "Role for github.com/giantswarm/azure-operator",
+  "Actions": [
+    "*"
+  ],
+  "NotActions": [
+    "Microsoft.Authorization/elevateAccess/Action"
+  ],
+  "AssignableScopes": [
+    "/subscriptions/${SUBSCRIPTION_ID}"
```
TheoBrigitte: Planning to use `os.ExpandEnv` to expand those variables later on.
But this also make this file not usable by azure cli as their is no way to pass the variable. Some shell magic would be needed there.

JosephSalisbury: I think that's fine.

xh3b4sd: I do not understand how this is used but in general I think this should be a go template then. Fucking with the environment is only a pain in the ass most of the time and later we have to refactor it anyway.

kopiczko: I like the go template idea. This can be regular file still. Not necessary a `.go` file. You can name it `guest.tmpl.json` or something.

xh3b4sd: When it is effectively a template then it should be `guest.json.tmpl`, but yeah.

kopiczko: `guest.json.tmpl` is impractical, because of editors highlighting.

xh3b4sd: Your editor should be configurable and detect that properly. The thing is that it is semantically more correct to express the file type with its extension. Go templates aren't valid JSON files by nature. So this would raise quite wrong expectations. You have to render the template in order to get a valid JSON as it should be.

kopiczko: Let's say I can configure my editor which is unecessary.
But I can't configure github.
And this is a common practice. Check for example files in this directory: https://github.com/giantswarm/azure-operator/tree/master/helm/azure-operator-chart/templates
We do that pretty much in every single project.

policy.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
+
+set -eu
+
+
+main() {
```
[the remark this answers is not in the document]
TheoBrigitte: Sorry that was a mistake, I just removed this.

PR #295, opened 2018-07-18
service-reconcilation-minor-bugfix

calvix:
found a small race condition with reconciliation fo service resource
If the `current service` is read from the early stage then it has no `spec.clusterIP` value and then if we tried to update the service with no `spec.clusterIP` it will throw errors as that value is immutable.


calvix:
The error wasn't critical as the next reconciliation loop it was fixed, but it is there to remove the amount fo errors from the operator.

PR #296, opened 2018-07-18
updates status resource

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918.


Gopkg.toml
```diff
@@ -86,8 +86,8 @@ required = [
   branch = "master"
   name = "github.com/giantswarm/operatorkit"
 
-[[constraint]]
-  branch = "master"
+[[override]]
+  branch = "nodes"
```
xh3b4sd: See https://github.com/giantswarm/statusresource/pull/4.

Gopkg.toml
```diff
@@ -38,8 +38,8 @@ required = [
   branch = "master"
   name = "github.com/docker/distribution"
 
-[[constraint]]
-  branch = "master"
+[[override]]
+  branch = "nodes"
```
xh3b4sd: See https://github.com/giantswarm/apiextensions/pull/100.

service/controller/v3/key/key.go
```diff
@@ -326,7 +326,7 @@ func ToVersionBundleVersion(v interface{}) (string, error) {
 		return "", microerror.Mask(err)
 	}
 
-	return customObject.Spec.VersionBundle.Version, nil
+	return VersionBundleVersion(customObject), nil
```
xh3b4sd: Boyscouting.

service/controller/v2/key/key.go
```diff
@@ -209,6 +209,17 @@ func ToCustomObject(v interface{}) (providerv1alpha1.AzureConfig, error) {
 	return customObject, nil
 }
 
+func ToNodeCount(v interface{}) (int, error) {
```
kopiczko: This should be rather `func NodeCount`

xh3b4sd: Why? We have the `To` prefix as a pretty common thing for everything that takes an interface value and returns something concrete.

kopiczko: Other functions in key package don't have `To` prefix. This prefix indicates for me the function is on conversion which it isn't.

xh3b4sd: One of the very first functions and a reason why we invented the key package at all was `ToCustombject`.

kopiczko: In fact we fist had extractors and the decided to move `ToCustombject` to the `key` package.

But I understand what is weird here. Normally it would look like:

```
func NodeCount(customObject providerv1alpha1.AzureConfig) int {
	nodeCount := len(customObject.Spec.Cluster.Masters) + len(customObject.Spec.Cluster.Workers)

	return nodeCount
}
```

That's what confused me and I didn't realize. Why this one does both conversion and extraction?

xh3b4sd: Because we use it to configure the `statusresource`, which is provider independent. See https://github.com/giantswarm/statusresource/blob/4193e00bf9e1bfcf2fa4a38480589bdf0e7fb6b5/resource.go#L17.

kopiczko: Ack 👍

PR #298, opened 2018-07-24
Disable Ingress Controller Service

rossf7:
Towards giantswarm/giantswarm/issues/3710

Uses v_3_5_0 of k8scloudconfig and sets new `DisableIngressControllerService` param to true. Ingress Controller on Azure is already migrated to `chart-operator`. So there is no change to the cloudconfigs that are rendered.


rossf7:
Tested on ghost.

Gopkg.toml
```diff
@@ -63,7 +63,7 @@ required = [
   name = "github.com/giantswarm/e2etests"
 
 [[constraint]]
-  branch = "master"
+  branch = "ingress-service-param"
```
r7vme: do not forget to change to master

rossf7: Thanks! Will do.

rossf7:
Tested on `godsmack` and CI green. Going with this.

Issue #284, opened 2018-07-05
Clean up orphaned resources


TheoBrigitte:
In https://github.com/giantswarm/azure-operator/pull/278 vnet peering resource was replaced by vpn gateway in order to connect both host and guest clusters.

Currently the new controller version (`1.1.0`) will ensure that vpn gateway is created and connected, but vnet peering created by previous version (`1.0.0`) will remain there.
We need to ensure that orphaned vnet peering are cleaned up when migrating from `1.0.0` to `1.1.0`.

How to reproduce:
* create cluster using version `1.0.0`
* update cluster to `1.1.0`
* updated cluster will have both vpn gateway and vnet peering

TheoBrigitte:
There is two parts in this task:
* delete vnetpeering from the host cluster network (covered in #302)
* delete vnetpeering from the guest cluster network

For the guest cluster part, resources are handled via arm_templates. So the easiest solution there would be to switch from `Incremental` to `Complete` deployment mode, which would take care of cleaning up any resources not specified in the arm_template.

Doing some research I stumbled #236 where arm_template deployment mode was switched from `Complete` to `Incremental`. This does not seems to be an option then. Will proceed with sdk as for host cluster.

TheoBrigitte:
There is now a new resource on called `vnetpeeringcleaner` which is in charge of removing both host and guest clusters virtual network peering resource.

PR #302, opened 2018-08-01
Add vnetpeering cleaner resource

TheoBrigitte:
Towards: https://github.com/giantswarm/azure-operator/issues/284

Implement cleaning of left over vnetpeering resource created by controller v2 and removed in v3.

Note: this only clean up the vnetpeering on the host cluster side. I will adresse cleaning up of the vnetpeering on the guest cluster in a following PR.


service/controller/v3/resource/vnetpeeringcleaner/create.go
```diff
@@ -0,0 +1,43 @@
...
+
+	vnetPeeringClient, err := r.getVnetPeeringClient()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	resourceGroupName := r.azure.HostCluster.ResourceGroup
+	vnetName := r.azure.HostCluster.ResourceGroup
+	peeringName := key.ResourceGroupName(azureConfig)
+	respFuture, err := vnetPeeringClient.Delete(ctx, resourceGroupName, vnetName, peeringName)
```
xh3b4sd: I am pretty sure deleting something that is already deleted causes an error, which is not handled here. Check the `resourcegroup` resource. There is a `IsNotFound` error matcher implementation.

service/controller/v3/resource/vnetpeeringcleaner/create.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// DeleteResponder ensure that response body is closed.
+	_, err = vnetPeeringClient.DeleteResponder(respFuture.Response())
```
xh3b4sd: Can you explain why we need this? I do not understand it. IMO it is not necessary.

service/controller/v3/resource/vnetpeeringcleaner/create.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+// EnsureCreated ensure that vnetpeering resource are deleted,
+// since they are no longer in use in this version.
+func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring delete host vnetpeering")
```
xh3b4sd: I would say it would be better and more correct in one of these two ways. 

- `ensuring deletion of host vnetpeering`
- `deleting host vnetpeering`

service/controller/v3/resource/vnetpeeringcleaner/create.go
TheoBrigitte: It is to ensure response body is closed, in order to avoid memory leaks.

TheoBrigitte: According to azure godoc: "DeleteResponder handles the response to the Delete request`"

We can then handle result with helpers like `client.ResponseWasNotFound`
https://github.com/giantswarm/azure-operator/blob/0e88f6eefa1ddcdd9d080fa32858ce641909b1d0/client/azure_client_set.go#L123-L129

service/controller/v3/resource/vnetpeeringcleaner/create.go
TheoBrigitte: I am not sure what is the right way to do things here.

- using `IsNotFound` like in `resourcegroup`, which result in less boilerplate
- using `DeleteResponder` and `client.ResponseWasNotFound`, which handle some magic like response body closing.

service/controller/v3/resource/vnetpeeringcleaner/create.go
xh3b4sd: If this is true we have these memory leaks in several places. I am not into the details of the SDK implementation but I would guess there is not leaking anything that you do not make use of.

service/controller/v3/resource/vnetpeeringcleaner/create.go
xh3b4sd: With respect to the leaking issue, which I do not see, I would just go with the error handling.

xh3b4sd: I mean, maybe I am wrong. Then lets make it right, but lets maybe check and verify based on data and not assume magic which is none.

TheoBrigitte: I used a simple program in combination with memory profiling in order to attest the evidence of memory leaks when not using `*Responder` functions. You can find it here https://gist.github.com/TheoBrigitte/212fc1ff17b9f954833e37f4d078125e

I used the list instead of the delete function as it is more convenient for such test, but the underlying logic is the same. All those functions follow the same pattern `Preparer`, `Sender`, and `Responder`.

It does list peering 1000x, once with the responder and once without it.
Here is the top memory consumption for both run:

Without responder:
```
Showing nodes accounting for 54.24MB, 97.31% of 55.74MB total
Showing top 20 nodes out of 104
      flat  flat%   sum%        cum   cum%
   10.56MB 18.94% 18.94%    10.56MB 18.94%  crypto/tls.(*block).reserve
    8.55MB 15.34% 34.28%     8.55MB 15.34%  bytes.makeSlice
    5.02MB  9.00% 43.28%     5.02MB  9.00%  bufio.NewReaderSize (inline)
    4.02MB  7.20% 50.49%     4.02MB  7.20%  bufio.NewWriterSize (inline)
    4.01MB  7.20% 57.68%    19.61MB 35.18%  crypto/tls.(*Conn).readHandshake
    3.50MB  6.29% 63.97%    11.01MB 19.74%  crypto/x509.parseCertificate
    3.50MB  6.28% 70.25%     3.50MB  6.28%  reflect.unsafe_NewArray
       3MB  5.38% 75.64%        3MB  5.38%  math/big.nat.make (inline)
       2MB  3.59% 79.22%        2MB  3.59%  crypto/x509/pkix.(*Name).FillFromRDNSequence
    1.50MB  2.69% 81.92%     1.50MB  2.69%  crypto/aes.(*aesCipherGCM).NewGCM
    1.50MB  2.69% 84.61%    12.54MB 22.49%  net/http.(*Transport).dialConn
    1.50MB  2.69% 87.30%     1.50MB  2.69%  crypto/tls.(*halfConn).newBlock
       1MB  1.80% 89.09%        1MB  1.80%  fmt.Sprintf
       1MB  1.80% 90.89%        1MB  1.80%  crypto/tls.Client (inline)
       1MB  1.79% 92.68%        1MB  1.79%  runtime.malg
    0.58MB  1.04% 93.72%     0.58MB  1.04%  crypto/elliptic.initTable
    0.50MB   0.9% 94.62%     0.50MB   0.9%  crypto/tls.(*Config).Clone
    0.50MB   0.9% 95.51%     0.50MB   0.9%  crypto/aes.newCipher
    0.50MB   0.9% 96.41%     0.50MB   0.9%  net.newFD (inline)
    0.50MB   0.9% 97.31%     0.50MB   0.9%  net.cgoLookupIP
```

With responder:
```
Showing nodes accounting for 2049.64kB, 100% of 2049.64kB total
Showing top 20 nodes out of 21
      flat  flat%   sum%        cum   cum%
 1537.48kB 75.01% 75.01%  2049.64kB   100%  crypto/x509.parseCertificate
  512.16kB 24.99%   100%   512.16kB 24.99%  math/big.nat.make (inline)
         0     0%   100%  2049.64kB   100%  crypto/tls.(*Conn).Handshake
         0     0%   100%  2049.64kB   100%  crypto/tls.(*Conn).clientHandshake
         0     0%   100%  2049.64kB   100%  crypto/tls.(*clientHandshakeState).doFullHandshake
         0     0%   100%  2049.64kB   100%  crypto/tls.(*clientHandshakeState).handshake
         0     0%   100%  2049.64kB   100%  crypto/x509.(*CertPool).AppendCertsFromPEM
         0     0%   100%  2049.64kB   100%  crypto/x509.(*Certificate).Verify
         0     0%   100%  2049.64kB   100%  crypto/x509.ParseCertificate
         0     0%   100%  2049.64kB   100%  crypto/x509.initSystemRoots
         0     0%   100%  2049.64kB   100%  crypto/x509.loadSystemRoots
         0     0%   100%   512.16kB 24.99%  crypto/x509.parsePublicKey
         0     0%   100%  2049.64kB   100%  crypto/x509.systemRootsPool
         0     0%   100%   512.16kB 24.99%  encoding/asn1.Unmarshal
         0     0%   100%   512.16kB 24.99%  encoding/asn1.UnmarshalWithParams
         0     0%   100%   512.16kB 24.99%  encoding/asn1.parseBigInt
         0     0%   100%   512.16kB 24.99%  encoding/asn1.parseField
         0     0%   100%   512.16kB 24.99%  math/big.(*Int).SetBytes
         0     0%   100%   512.16kB 24.99%  math/big.nat.setBytes
         0     0%   100%  2049.64kB   100%  net/http.(*persistConn).addTLS.func2
```

We can clearly see the memory accumulation when not using the responder here.

service/controller/v3/resource/vnetpeeringcleaner/create.go
```diff
@@ -0,0 +1,46 @@
...
+	vnetName := r.azure.HostCluster.ResourceGroup
+	peeringName := key.ResourceGroupName(azureConfig)
+	respFuture, err := vnetPeeringClient.Delete(ctx, resourceGroupName, vnetName, peeringName)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// DeleteResponder ensure that response body is closed.
+	res, err := vnetPeeringClient.DeleteResponder(respFuture.Response())
+	if client.ResponseWasNotFound(res) {
```
kopiczko: I think the discussed error is handled here.

service/controller/v3/resource/vnetpeeringcleaner/create.go
xh3b4sd: Cool. Good to know. I wasn't aware of this. Can you please create an issue so we check the whole operator for this? There are several places in which we then should use the futures to close the response bodies. Can you also engage with the Azure folks in our Slack channel and give feedback about this? I feel like this is not obvious and the godoc of the SDK doesn't inform properly about the implications. IMO not using the responder should not result in such implicit behaviour. Like this is build to trap developers.

service/controller/v3/resource/vnetpeeringcleaner/create.go
xh3b4sd: Now it is, yes.

service/controller/v3/resource/vnetpeeringcleaner/create.go
```diff
@@ -0,0 +1,46 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensured delete host vnetpeering")
+	} else if err != nil {
+		return microerror.Mask(err)
+	} else {
```
xh3b4sd: We could omit the last `else` and just fall through with the not found error then we do not have to duplicate the log line.

service/controller/v3/resource/vnetpeeringcleaner/create.go
TheoBrigitte: Yeah good point, I'll engage with Azure folks about this, as I Responder is sometime called and sometime it is up to us to call it like here.

I'll create the issue to ensure we always close response bodies.

PR #305, opened 2018-08-02
re-use vnetpeering resource

TheoBrigitte:
Towards: #284 

Re-use the previously gone vnetpeering resource as a base for cleaning up orphaned vnetpeering.

### DIFF
--- a/service/controller/v3/resource/vnetpeering/create.go
+++ b/service/controller/v3/resource/vnetpeering/create.go
@@ -1,0 +1,10 @@
+package vnetpeering
+
+import (
+	"context"
+)
+
+// ApplyCreateChange is noop. Creation goes through ApplyUpdateChange.
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	return nil
+}
--- a/service/controller/v3/resource/vnetpeering/current.go
+++ b/service/controller/v3/resource/vnetpeering/current.go
@@ -1,0 +1,87 @@
+package vnetpeering
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+// GetCurrentState retrieve the current host cluster virtual network peering
+// resource from azure.
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return network.VirtualNetworkPeering{}, microerror.Mask(err)
+	}
+
+	// In order to make vnet peering work we need a virtual network which we can
+	// use to peer. In case there is no virtual network yet we cancel the resource
+	// and try again on the next resync period. This is a classical scenario on
+	// guest cluster creation. If we would not check for the virtual network
+	// existence the client calls of CreateOrUpdate would fail with not found
+	// errors.
+	if !key.IsDeleted(customObject) {
+		c, err := r.getVirtualNetworksClient()
+		if err != nil {
+			return network.VirtualNetworkPeering{}, microerror.Mask(err)
+		}
+
+		g := key.ResourceGroupName(customObject)
+		n := key.VnetName(customObject)
+		e := ""
+		v, err := c.Get(ctx, g, n, e)
+		if IsVirtualNetworkNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find the virtual network in the Azure API")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+
+			return network.VirtualNetworkPeering{}, nil
+		} else if err != nil {
+			return network.VirtualNetworkPeering{}, microerror.Mask(err)
+		} else {
+			s := *v.ProvisioningState
+
+			if !key.IsFinalProvisioningState(s) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("virtual network is in state '%s'", s))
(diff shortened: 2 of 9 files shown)


service/controller/v3/resource/vnetpeering/create.go
```diff
@@ -0,0 +1,10 @@
+package vnetpeering
+
+import (
+	"context"
+)
+
+// ApplyCreateChange is noop. Creation goes through ApplyUpdateChange.
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
```
xh3b4sd: I am wondering why we not just use the normal `Resource` interface. The logic we want here does not really require the CRUD guidance with current and desired state.

service/controller/v3/resource/vnetpeering/current.go
```diff
@@ -0,0 +1,87 @@
...
+		return network.VirtualNetworkPeering{}, microerror.Mask(err)
+	}
+
+	// In order to make vnet peering work we need a virtual network which we can
+	// use to peer. In case there is no virtual network yet we cancel the resource
+	// and try again on the next resync period. This is a classical scenario on
+	// guest cluster creation. If we would not check for the virtual network
+	// existence the client calls of CreateOrUpdate would fail with not found
+	// errors.
+	if !key.IsDeleted(customObject) {
```
xh3b4sd: I think these checks should be the exception and they are not really necessary here, right?

service/controller/v3/resource/vnetpeering/current.go
```diff
@@ -0,0 +1,87 @@
...
+			return network.VirtualNetworkPeering{}, nil
+		} else if err != nil {
+			return network.VirtualNetworkPeering{}, microerror.Mask(err)
+		} else {
+			s := *v.ProvisioningState
+
+			if !key.IsFinalProvisioningState(s) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("virtual network is in state '%s'", s))
+				resourcecanceledcontext.SetCanceled(ctx)
+				r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
xh3b4sd: Cancelation should actually just used to prevent the resource from doing something wrong or doing something it cannot do because some system state is not lined up. Here we only want to delete something and I am pretty sure no cancelation is required here.

service/controller/v3/resource/vnetpeering/delete.go
```diff
@@ -0,0 +1,87 @@
...
+		return microerror.Mask(err)
+	}
+
+	respFuture, err := vnetPeeringClient.Delete(ctx, r.azure.HostCluster.ResourceGroup, r.azure.HostCluster.ResourceGroup, *change.Name)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	res, err := vnetPeeringClient.DeleteResponder(respFuture.Response())
+	if client.ResponseWasNotFound(res) {
```
xh3b4sd: What is the responder good for here?

service/controller/v3/resource/vnetpeering/error.go
```diff
@@ -0,0 +1,46 @@
+package vnetpeering
+
+import (
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/giantswarm/microerror"
+)
+
+var invalidConfigError = microerror.New("invalid config")
```
xh3b4sd: Not sure if you missed the 100 PRs about rewriting errors but we want to do this differently now. See https://github.com/giantswarm/fmt/blob/master/go/errors.md.

service/controller/v3/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller"
+
+	"github.com/giantswarm/azure-operator/service/controller/v3/key"
+)
+
+// NewUpdatePatch provide a controller.Patch holding the needed network.VirtualNetworkPeering update to have current comply with desired.
+func (r *Resource) NewUpdatePatch(ctx context.Context, azureConfig, current, desired interface{}) (*controller.Patch, error) {
```
xh3b4sd: I am lost now. I thought this here is just to delete the orphaned peering. Why do we need this update magic now?

TheoBrigitte:
It was probably not clear from my side but this is just the base, where I start from in order to implement the vnetpeering cleaner.

I thought bringing this as a separate PR would ease up the review later on. But maybe that's just unneeded overhead and I should just proceed and submit the final resource.

service/controller/v3/resource/vnetpeering/create.go
kopiczko: I also think with normal `Resource` it would easier to review. I'm not very emotional when it comes to this particular resource because it's going to be removed in the next version, but it shouldn't be much work to rewrite it.

service/controller/v3/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,133 @@
...
+
+	patch.SetUpdateChange(change)
+
+	return patch, nil
+}
+
+func (r *Resource) newUpdateChange(ctx context.Context, azureConfig providerv1alpha1.AzureConfig, current, desired network.VirtualNetworkPeering) (network.VirtualNetworkPeering, error) {
+	var change network.VirtualNetworkPeering
+
+	if needUpdate(current, desired) {
```
kopiczko: Should be needUpdate check? I think we should always try to delete.

service/controller/v3/resource/vnetpeering/update.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if isVNetPeeringEmpty(change) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensure host vnet peering: already ensured")
+		return nil
+	}
+
+	_, err = vnetPeeringClient.CreateOrUpdate(ctx, r.azure.HostCluster.ResourceGroup, r.azure.HostCluster.VirtualNetwork, *change.Name, change)
```
kopiczko: Shouldn't that be `Delete`?

Issue #113, opened 2018-01-25
Disks on master are not mounted sometimes


r7vme:
etcd and docker disks are expected to be `sdc` and `sdd` respectevely.

master vm messed up disk naming once:
- OS disk became `sdc` 
- etcd disk became `sda`
- docker disk became `sdb`
- ephemeral (by default in Azure) became `sdd`

Node had following scsi attachments:
```
[1:0:0:0]    cd/dvd  Msft     Virtual CD/ROM   1.0   /dev/sr0 
[3:0:0:0]    disk    Msft     Virtual Disk     1.0   /dev/sdc 
[4:0:0:0]    disk    Msft     Virtual Disk     1.0   /dev/sda 
[4:0:0:1]    disk    Msft     Virtual Disk     1.0   /dev/sdb 
[5:0:1:0]    disk    Msft     Virtual Disk     1.0   /dev/sdd 
```

Node with right disk naming has `lscsi` output:
```
[1:0:0:0]    cd/dvd  Msft     Virtual CD/ROM   1.0   /dev/sr0 
[2:0:0:0]    disk    Msft     Virtual Disk     1.0   /dev/sda 
[3:0:1:0]    disk    Msft     Virtual Disk     1.0   /dev/sdb 
[5:0:0:0]    disk    Msft     Virtual Disk     1.0   /dev/sdc 
[5:0:0:1]    disk    Msft     Virtual Disk     1.0   /dev/sdd 
```

We probably need to find better way, to detect disks. Right now we rely on kernel disk names (e.g. `sdc`)

r7vme:
hmm, acs-engine also hardcodes [disk name](https://github.com/Azure/acs-engine/blob/master/parts/k8s/kubernetes_mountetcd.sh#L5). In general, we can not identify which disk for what, because block device is empty and does not contain any labels. Only proper way if Azure can put device labels, but this are only my dreams.

Ok, so this issue can not really be solved properly.

What we should do is to trigger immediate alert.

@JosephSalisbury if .mount unit failed will node-exporter detect that in guest cluster?

r7vme:
@kopiczko please report here if you'll see .mount units failed in guest clusters. Thanks

JosephSalisbury:
See https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus-chart/prometheus-rules/systemd.rules#L1

We get alerts on failed systemd units.

JosephSalisbury:
[...]

PR #308, opened 2018-08-03
Fix Azure disk mounts

r7vme:
Fixes: https://github.com/giantswarm/giantswarm/issues/3847
Fixes: https://github.com/giantswarm/azure-operator/issues/113

In Azure you can set LUN ID, when attaching disk to VM. CoreOS supplied with special udev rules for Azure that create symlinks in `/dev/disk/azure/scsi1/lunX`, where `X` is the lun that was specified while attaching.

### DIFF
--- a/service/controller/v3/cloudconfig/template.go
+++ b/service/controller/v3/cloudconfig/template.go
@@ -589,7 +589,7 @@ After=format-etcd-disk.service
 Before=etcd3.service
 
 [Mount]
-What=/dev/{{ .DiskName }}
+What=/dev/disk/by-label/etcd
 Where=/var/lib/etcd
 Type=ext4
 
@@ -599,14 +599,14 @@ WantedBy=multi-user.target
 	etcdDiskFormatUnitName     = "format-etcd-disk.service"
 	etcdDiskFormatUnitTemplate = `[Unit]
 Description=Formats the disk drive
-Requires=dev-{{ .DiskName }}.device
-After=dev-{{ .DiskName }}.device
+Requires=dev-disk-azure-scsi1-lun{{ .LUNID }}.device
+After=dev-disk-azure-scsi1-lun{{ .LUNID }}.device
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 Environment="LABEL=etcd"
-Environment="DEV=/dev/{{ .DiskName }}"
+Environment="DEV=/dev/disk/azure/scsi1/lun{{ .LUNID }}"
 # Do not wipe the disk if it's already being used, so the etcd data is persistent across reboot.
 ExecStart=-/bin/bash -c "if ! findfs LABEL=$LABEL > /tmp/label.$LABEL; then wipefs -a -f $DEV && mkfs.ext4 -T news -F -L $LABEL $DEV && echo wiped; fi"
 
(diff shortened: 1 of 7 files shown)


service/controller/v3/cloudconfig/template.go
```diff
@@ -631,16 +631,15 @@ WantedBy=multi-user.target
...
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 Environment="LABEL=docker"
-Environment="DEV=/dev/{{ .DiskName }}"
-# Do not wipe the disk if it's already being used.
-ExecStart=-/bin/bash -c "if ! findfs LABEL=$LABEL > /tmp/label.$LABEL; then wipefs -a -f $DEV && mkfs.xfs -L $LABEL $DEV && echo formatted; fi"
+Environment="DEV=/dev/disk/azure/scsi1/lun{{ .LUNID }}"
+ExecStart=-/bin/bash -c "wipefs -a -f $DEV && mkfs.xfs -L $LABEL $DEV"
```
kopiczko: I think we can reuse templates for docker and etcd. It would be just enough to add `WipeOut bool` and single `if`.

r7vme: Let's not do that.

1. they are different in 50% of lines
2. we soon will be moving to ignition and will get rid of them.

r7vme:
@xh3b4sd need your advice with upgrade. I see that it tries to upgrade master already for a second time (now goes for third one).

See `kubectl get nodes` history.

[kubectl-get-node.txt](https://github.com/giantswarm/azure-operator/files/2257201/kubectl-get-node.txt)

r7vme:
It's cluster `ugw0r` in ghost, i'll leave it for debugging.

xh3b4sd:
I will check after lunch. Thanks for reporting.

xh3b4sd:
So I can confirm there is an issue with the selection of the instance actions. We do in fact update the master twice. See https://github.com/giantswarm/giantswarm/issues/3908. Anyways this is not a serious issue AFAICT. The update succeeds properly. It just takes a bit of time. Verified this with several clusters. 

The problem here might be that the deployments failed. The operator logs suggest this as well. I reckon it may be related to the changes of the PR. 

```
deployment is in state 'Failed'
```

```
{"caller":"github.com/giantswarm/azure-operator/service/controller/v3/resource/instance/create.go:44","event":"update","level":"debug","message":"deployment is in state 'Failed'","object":"/apis/provider.giantswarm.io/v1alpha1/namespaces/default/azureconfigs/ugw0r","resource":"instancev3","time":"2018-08-03T14:51:40.392947+00:00"}
```

screen shot 2018-08-03 at 16 56 51

Let me know when I can help further.

r7vme:
Looking deeper in the logs of "kubernetes" load balancer i see the only error.

```
"statusMessage": "{\"status\":\"Failed\",\"error\":{\"code\":\"ResourceOperationFailure\",\"message\":\"The resource operation completed with terminal provisioning state 'Failed'.\",\"details\":[{\"code\":\"InternalExecutionError\",\"message\":\"An internal execution error occurred.\",\"details\":[]}]}}",
```

Retesting.

r7vme:
Tested upgrade and works fine.

r7vme:
wip e2e tests taking very long time due to [this](https://github.com/giantswarm/azure-operator/issues/311).

r7vme:
please review

- e2e tests finally passed (seems flapping was the reason)
- upgrade was tested on `ghost`. Issue above was not the case.

PR #310, opened 2018-08-03
use status versions

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3822.


service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +31,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	{
-		d, err := deploymentsClient.Get(ctx, key.ClusterID(customObject), vmssDeploymentName)
-		if IsDeploymentNotFound(err) {
-			// fall through
+	if !resourceStatusExists(customObject, "Stage") {
```
xh3b4sd: The diff is shite. There are 3 stages the `instance` resource ensures sequentially. These stages where executed in a rather uncontrolled way which is the reason for the issue this PR addresses and fixes. These are the three stages. 

- Uploading ARM templates, described by stage `DeploymentInitialized`.
- Waiting for ARM templates to be applied, described by stage `ProvisioningSuccessful`.
- Draining and upgrading the VMSS instances, described by stage `InstancesUpgrading`.

The stages are executed one after another and the instance cycles through them reliably until all necessary upgrade magic is sorted.

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +31,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deployment")
+
+		computedDeployment, err := r.newDeployment(ctx, customObject, nil)
+		if controllercontext.IsInvalidContext(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "missing dispatched output values in controller context")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not ensure deployment")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```
xh3b4sd: Improved the logging a bit here and there.

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +31,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			if err != nil {
+				return microerror.Mask(err)
 			}
-		}
-	}
 
-	// TODO this is a hack we have to fix as soon as the statusresource works and
-	// is enabled.
-	//
-	// Issue: https://github.com/giantswarm/giantswarm/issues/3822
```
xh3b4sd: We are using the node versions provided with the CR status so this part of the TODO is sorted.

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +31,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 		} else {
-			_, err = deploymentsClient.CreateOrUpdate(ctx, key.ClusterID(customObject), vmssDeploymentName, computedDeployment)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "not processing worker VMSSs due to master VMSSs processing")
+		}
+
+		if !masterUpgraded && !workerUpgraded {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "neither masters nor workers upgraded")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "removing resource status 'Stage/InstancesUpgrading'")
+
+			err := r.deleteResourceStatus(customObject, "Stage", "InstancesUpgrading")
```
xh3b4sd: The cycle ends here and with removal of the last stage status the next cycle starts with the next resync period.

service/controller/v3/resource/instance/create_test.go
```diff
@@ -544,6 +544,138 @@ func Test_Resource_Instance_findActionableInstance(t *testing.T) {
 	}
 }
 
+func Test_Resource_Instance_computeForDeleteResourceStatus(t *testing.T) {
```
xh3b4sd: Had to write some unit tests to ensure the status structures are cleaned properly and do not leave behind empty garbage structures.

xh3b4sd:
Successfully tested on `godsmack`.

xh3b4sd:
@giantswarm/sig-operator I am fishing for LGTM here.

integration/teardown/teardown.go
```diff
@@ -15,7 +15,8 @@ func Teardown(c *client.AzureClientSet, g *framework.Guest, h *framework.Host) e
 	var err error
 
 	{
-		h.DeleteGuestCluster("azure-operator", "azureconfig", "removing finalizer from runtime object")
+		// TODO the deletion detection should rather happen based on the cluster status or even based on the CR being gone
```
rossf7: If this is a long term TODO should there be an issue? If its short term I'd still like to see a sentence with a `.`.

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +31,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-	{
-		var k8sClient kubernetes.Interface
-		{
-			i := key.ClusterID(customObject)
-			e := key.ClusterAPIEndpoint(customObject)
-			k8sClient, err = r.guestCluster.NewK8sClient(ctx, i, e)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "ensured deployment")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "setting resource status to 'Stage/DeploymentInitialized'")
+
+			err = r.setResourceStatus(customObject, "Stage", "DeploymentInitialized")
```
rossf7: Nitpick but `State` and `DeploymentInitialized ` strings are used a lot. Could they be extracted to constants?

Are they just for information to users or is there code checking for these values?

service/controller/v3/resource/instance/create.go
rossf7: Thanks for the explanation that makes sense.

integration/teardown/teardown.go
xh3b4sd: Good catch. I was lazy here. The issue is https://github.com/giantswarm/giantswarm/issues/3839. Added it to the comment in the code.

service/controller/v3/resource/instance/create.go
xh3b4sd: Should I add it to the godoc for `EnsureCreated`?

TheoBrigitte: Yes that would help

rossf7: +1 that would be useful.

xh3b4sd: I added godoc.

service/controller/v3/resource/instance/create.go
xh3b4sd: Good shout. I added constants for them.

service/controller/v3/resource/instance/create.go
TheoBrigitte: This feels like a lot of lines, and IMO it is overwhelming when we have to read logs.
Could we somehow reduce this to one line ? Maybe with new key value pairs, to provide more information on a single line.

e.g.
`r.logger.LogCtx(ctx, "level", "debug", "message", "missing dispatched output values in controller context", "action", "canceling resource")`

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +49,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-			if err != nil {
-				return microerror.Mask(err)
-			}
-			err = r.createDrainerConfig(ctx, customObject, workerInstanceToDrain, key.WorkerInstanceName)
-			if err != nil {
-				return microerror.Mask(err)
-			}
-			err = r.reimageInstance(ctx, customObject, workerInstanceToReimage, key.WorkerVMSSName, key.WorkerInstanceName)
-			if err != nil {
+		var masterUpgraded bool
```
TheoBrigitte: The whole block about upgrading master is repeated for workers, it would be great to move this into a function to avoid repetition.

service/controller/v3/resource/instance/create.go
```diff
@@ -33,187 +49,205 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-		if guest.IsAPINotAvailable(err) {
-			// fall through
-			r.logger.LogCtx(ctx, "level", "debug", "message", "not fetching node versions")
-			r.logger.LogCtx(ctx, "level", "debug", "message", "guest API is not available")
+	if hasResourceStatus(customObject, Stage, DeploymentInitialized) {
+		d, err := deploymentsClient.Get(ctx, key.ClusterID(customObject), vmssDeploymentName)
+		if IsDeploymentNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "deployment not found")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "waiting for creation")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```
TheoBrigitte: Resource is canceled without using `resourcecanceledcontext` is that intended ?

xh3b4sd: This is documented in https://github.com/giantswarm/operatorkit/blob/master/docs/control_flow_primitives.md#cancel-resources. This here is the normal `Resource` implementation. No need for context cancelation. Simple `return` does the trick.

service/controller/v3/resource/instance/create.go
xh3b4sd: You have a good point but I am not addressing it in this PR because it was like this already and the PR already derailed more than I anticipated. For sure something to boyscout down the road.

service/controller/v3/resource/instance/create.go
xh3b4sd: IMO this aligns with https://github.com/giantswarm/fmt/blob/master/go/log_messages.md. I agree it is quite expressive and maybe intense but I like it a lot. These logs make perfect sense when you follow them in production or during development. You know what happens and this is what they are good for.

PR #316, opened 2018-08-07
adds e2e test for updates

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918.


.circleci/config.yml
```diff
@@ -91,6 +93,22 @@ jobs:
 
 
 
+  # Note we only have update tests for the current version because it anyway
+  # tests one version transition. Having the same tests for WIP only runs the
+  # same test twice and we can simply safe resources by not doing so.
```
rossf7: 🇬🇧 nit, save resources.

integration/env/common.go
```diff
@@ -51,6 +51,7 @@ var (
 	circleSHA            string
 	clusterID            string
 	commonDomain         string
+	githubToken          string
```
rossf7: <3

service/controller/v1/resource_set.go
```diff
@@ -264,8 +259,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 
 	{
 		c := retryresource.WrapConfig{
-			BackOffFactory: func() backoff.BackOff { return backoff.WithMaxTries(backoff.NewExponentialBackOff(), ResourceRetries) },
-			Logger:         config.Logger,
+			Logger: config.Logger,
```
rossf7: So this backoff factory is only used in v1?

xh3b4sd: Due to the migration to our backoff lib there is no `NewExponentialBackOff` anymore. In fact the `retryresource` in operatorkit defaults to a backoff factory. It is just that we never made use of the defaulting because we anticipated the `v1` package goes away anyway. Now I just had to make it compile so I simply removed it because the defaulting covers it after all.

rossf7: Cool thanks for the explanation.

.circleci/config.yml
xh3b4sd: Done.

xh3b4sd:
e2e tests are green. First e2e tests for updates are in place. Rolling.

PR #322, opened 2018-08-16
v4: copy v3

TheoBrigitte:
Create controller v4

PR #325, opened 2018-08-16
wire controller/v4

TheoBrigitte:
Wire new controller v4

Previous PRs: #322 #323


service/controller/v4/version_bundle.go
```diff
@@ -55,6 +55,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "azure-operator",
-		Version: "1.1.0",
+		Version: "1.2.0",
```
rossf7: The existing changelogs should be replaced with something like "Add your changes here.".

TheoBrigitte: done

PR #334, opened 2018-08-27
Fix missing vmssMSIEnabled conditionals

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/4024

I created a cluster with this on `tarantula` and it came up properly. Cluster is currently still running as `85off`.


MarcelMue:
Do we need a patch release for this? I believe that azure-operator should be fine when we push to master, the template doesn't change for any clusters we have running.

Opinions please.

xh3b4sd:
I am fine either way. Backport or patch is fine. We just have to do the leg work and update to the patches actively. Do what you think is right.

xh3b4sd:
The only thing that we have to check with the backporting approach is that existing clusters are properly reconciled once the backport is deployed. The e2e tests do not cover this so we have to do this manually then. But the same applies for new patch releases. Do your do.

MarcelMue:
I did the dance of 
- create cluster on godsmack with azure-operator from master
- wait for cluster to come up
- deploy this branch of azure-operator

It seems to be fine, cluster is healthy as far as I can tell. I'll go with backport if the e2e tests are fine.

xh3b4sd:
We had an issue in master. Got sorted. I merged master here so CI can be green. <3

MarcelMue:
CI is green after latest merge from master, going with this.

PR #337, opened 2018-08-28
Rename guest.tmpl.json to tenant.tmpl.json

marians:
Towards https://github.com/giantswarm/giantswarm/issues/3991


TheoBrigitte:
Do not forget to update the link from the documentation https://github.com/giantswarm/docs/blob/master/src/content/guides/prepare-azure-subscription-for-tenant-clusters/index.md#2-role-definition

marians:
`build_and_e2eTest` is on hold. What does that mean?

TheoBrigitte:
> build_and_e2eTest is on hold. What does that mean?

This is because we do not want to always run e2e test for every push, it would be way to expensive in terms of CI resources.

If think you need to run e2e tests, click the `Details` button, which is taking you to the Circle CI workflow page, then click `hold` and `Approve`. 

Since you are only renaming a file with very low impact here, there is no need to run e2e tests

PR #338, opened 2018-08-28
add credential-default to helm chart

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/3815

Add `credential-default` which is now required by azure-operator to operate tenant cluster.


xh3b4sd:
We had an issue in master. Got sorted. I merged master here so CI can be green. <3

PR #336, opened 2018-08-28
adds operator collector to kickoff better observability

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4026.

PR #339, opened 2018-08-28
refactors network lib to make it version independent

xh3b4sd:
We can decouple some functionality a bit from all the version magic. It is not necessary to bind these things to a certain resource package version. With this we can then also use the refactored libs transparently for the operator's collector. Preparations can be found in https://github.com/giantswarm/azure-operator/pull/336.


xh3b4sd:
I have no idea to be honest. I am just working on my postmortem right now and want the code to be better than before. We can totally come back here in case there are ideas, but I have none at this point.

PR #349, opened 2018-09-04
Add worker docker volume size 

MarcelMue:


MarcelMue:
Out of sync here, going to pick this up again when we get to it.

PR #366, opened 2018-09-27
Disable k8scc CoreDNS templates

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902


fgimenez:
Tested on godsmack.

PR #370, opened 2018-09-28
wire v5, bump version and initialize changelog

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/4287


fgimenez:
Good catch! It is missing, I'll add it shortly.

PR #371, opened 2018-09-28
Include memleak patch

fgimenez:
Towards giantswarm/giantswarm#4287

Using k8scc `v_3_6_1` (which includes the 1.11 mem leak fix) in both `v4` and `v5`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -279,9 +279,10 @@
   packages = [
     "v_3_3_4",
     "v_3_5_0",
-    "v_3_6_0"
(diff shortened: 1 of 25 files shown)


fgimenez:
We had consistent e2e test failures for `2.0.0`, for instance https://circleci.com/gh/giantswarm/azure-operator/3490. The operator logs look ok to me, not sure how to interpret these errors. I tested manually on godsmack and all went good with `2.0.0`, including the 1.11 image with the memleak patch, so going with this.

PR #374, opened 2018-10-10
removes ready e2e test and enables others

xh3b4sd:

PR #375, opened 2018-10-10
removes not used azure client set

xh3b4sd:
Note this branch is currently based on https://github.com/giantswarm/azure-operator/pull/374. I will merge the base first and base this here then off of `master`. I just prepare some work now so I can easily merge away later once approved.


integration/setup/setup.go
```diff
@@ -30,10 +29,10 @@ const (
 )
 
 // WrapTestMain setup and teardown e2e testing environment.
-func WrapTestMain(c *client.AzureClientSet, g *framework.Guest, h *framework.Host, m *testing.M) {
+func WrapTestMain(g *framework.Guest, h *framework.Host, m *testing.M) {
```
xh3b4sd: I have no idea what this is about and why it is that way but it is apparently not used for anything. So bye bye. Licky boom boom down.

PR #377, opened 2018-10-11
improves setup by leveraging config structure

xh3b4sd:


integration/setup/teardown.go
```diff
@@ -1,6 +1,6 @@
 // +build k8srequired
 
-package teardown
+package setup
```
xh3b4sd: I do not like this. Something is weird here. But lets iterate. Having setup and teardown in a `setup` package feels super wrong. The package name should maybe just change. `wrap` would work for me. Uncertain.

MarcelMue: I think the naming is just all around wrong. Setup was never a pure setup package but wraps around the test itself. This has been bugging me for a while as well, we will need to iterate here.

kopiczko: My idea was you create and then tear down the e2e setup.

PR #378, opened 2018-10-11
Added calico limits and update to 3.2.3 bugfix release

r7vme:
Part of: https://github.com/giantswarm/giantswarm/issues/4159

### DIFF
--- a/service/controller/v4/version_bundle.go
+++ b/service/controller/v4/version_bundle.go
@@ -31,7 +31,7 @@ func VersionBundle() versionbundle.Bundle {
 		Components: []versionbundle.Component{
 			{
 				Name:    "calico",
-				Version: "3.0.8",
+				Version: "3.2.0",
 			},
 			{
 				Name:    "containerlinux",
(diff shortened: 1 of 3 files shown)


service/controller/v4/version_bundle.go
```diff
@@ -31,7 +31,7 @@ func VersionBundle() versionbundle.Bundle {
 		Components: []versionbundle.Component{
 			{
 				Name:    "calico",
-				Version: "3.0.8",
+				Version: "3.2.0",
```
r7vme: Align with reality

service/controller/v5/cloudconfig/template.go
```diff
@@ -195,6 +196,8 @@ spec:
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      nodeSelector:
```
r7vme: From 3.2.3 manifest, just to be aligned

service/controller/v4/version_bundle.go
pipo02mix: 3.2.3?

r7vme: 2.0.0 bundle has 3.2.0 calico, but we forgot to update this.

PR #381, opened 2018-10-12
installs the azure-operator using the release package

xh3b4sd:


integration/setup/setup.go
```diff
@@ -69,36 +73,76 @@ func Setup(c Config) error {
...
+			},
+		}
+
+		values, err := chartvalues.NewAzureOperator(c)
 		if err != nil {
 			return microerror.Mask(err)
 		}
-		err = c.Host.InstallStableOperator("cert-operator", "certconfig", e2etemplates.CertOperatorChartValues)
+
+		err = config.Release.InstallOperator(context.Background(), "azure-operator", release.NewVersion(env.CircleSHA()), values, providerv1alpha1.NewAzureConfigCRD())
```
kopiczko: Let's please pass the context as an argument from top-level `Setup` call <3

xh3b4sd: I will do this later. One step at a time. I do not yet have a picture where it should come from but we will get to this point. No worries.

PR #386, opened 2018-10-17
separates provider and common setup into files

xh3b4sd:


integration/setup/common.go
```diff
@@ -0,0 +1,60 @@
...
+
+	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/e2e-harness/pkg/release"
+	"github.com/giantswarm/e2etemplates/pkg/chartvalues"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/integration/env"
+)
+
+func Common(config Config) error {
```
TheoBrigitte: A comment on what does this function would be nice.

integration/setup/provider.go
```diff
@@ -0,0 +1,93 @@
...
+
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/e2e-harness/pkg/release"
+	"github.com/giantswarm/e2etemplates/pkg/chartvalues"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/integration/env"
+)
+
+func Provider(config Config) error {
```
TheoBrigitte: A comment on what does this function would be nice.

integration/setup/common.go
xh3b4sd: Then please provide a suggestion. This is so obvious to me personally that I do not know what to write here as everything I could come up with would be rather useless. Please also suggest for the one below.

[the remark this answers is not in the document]
xh3b4sd: Done.

xh3b4sd:
Going with this. @TheoBrigitte feel free to suggest godoc I will add it then later.

integration/setup/common.go
TheoBrigitte: common installs components required to run the operator.

provider installs the operator and tenant cluster CR.

PR #387, opened 2018-10-17
Add blob container for ignition

corest:
With next PR I'll add storage client and sas tokens for retrieving ignition


service/controller/v5/arm_templates/virtual_network_setup.json
```diff
@@ -10,7 +10,7 @@
     },
     "virtualNetworksAPIVersion":{
       "type":"string",
-      "defaultValue":"2016-09-01",
+      "defaultValue":"2016-12-01",
```
xh3b4sd: Why does this change now? Does that have any implications for the client libs we use in the SDK? They are versioned in a similar way as well.

service/controller/v5/arm_templates/main.json
```diff
@@ -68,6 +68,10 @@
       "type":"string",
       "defaultValue":"dns_cname_setup.json"
     },
+    "ignitionContainerSetupTemplateFile":{
+      "type":"string",
+      "defaultValue":"ignition_container_setup.json"
+    },
```
xh3b4sd: I need some more explanation on why we need this container, what it does and how all this works. Do we have specs or docs or some diagrams for that?

service/controller/v5/arm_templates/virtual_network_setup.json
kopiczko: AFAIK this doesn't have to be in sync with SDK. Different API versions can deal with the same resources.

service/controller/v5/arm_templates/ignition_container_setup.json
```diff
@@ -0,0 +1,69 @@
...
+            "type": "string"
+        },
+        "kind": {
+            "defaultValue": "BlobStorage",
+            "type": "string"
+        },
+        "supportsHttpsTrafficOnly": {
+            "defaultValue": true,
+            "type": "bool"
+        },
```
kopiczko: Can parameters above become variables? I find it easier to reason about as I know those aren't configurable outside the template.

service/controller/v5/arm_templates/ignition_container_setup.json
```diff
@@ -0,0 +1,69 @@
...
+            "type": "string"
+        },
+        "accountType": {
+            "defaultValue": "Standard_RAGRS",
+            "type": "string"
+        },
+        "blobContainer": {
+            "defaultValue": "ignition",
+            "type": "string"
+        },
```
kopiczko: I think this should be more generic. Something like `container_setup.json` would make sense for me. `blobContainer` (which I would call `blobContainerName`) should come from the `key` package as it will be also used by one of the operator's resources to store the actual ignition.

service/controller/v5/arm_templates/virtual_network_setup.json
corest: It was changed because I was going to use service endpoint for Storage. And old version of API didn't support it. But as doesn't work with azure-operator, I switched that back

service/controller/v5/arm_templates/main.json
corest: https://github.com/giantswarm/giantswarm/issues/2816#issuecomment-430650737

service/controller/v5/arm_templates/ignition_container_setup.json
corest: The can, but variables seems to be explicitely mutable for me. And parameters with default values no

kopiczko: > variables seem to be explicitly mutable

I don't understand. They can't be mutated once initialized. And it's enforced they can't be set from outside the template.

PR #390, opened 2018-10-25
Add resource limits for install-cni container and make it initContainer.

r7vme:
Part of: https://github.com/giantswarm/giantswarm/issues/4159

Makes calico QoS policy guaranteed.

### DIFF
--- a/service/controller/v5/cloudconfig/template.go
+++ b/service/controller/v5/cloudconfig/template.go
@@ -6,6 +6,7 @@ const (
 	calicoAzureFilePermission = 0600
 	calicoAzureFileTemplate   = `# Extra changes:
 #  - Added "nodename_file_optional" set to true (can be removed on the next upgrade).
+#    Tracked here: https://github.com/giantswarm/giantswarm/issues/4113
 #  - Added resource limits to calico-node.
 #  - Added resource limits to install-cni.
 #  - Made install-cni initContainer.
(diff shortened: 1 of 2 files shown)


service/controller/v5/cloudconfig/template.go
```diff
@@ -5,8 +5,10 @@ const (
 	calicoAzureFileOwner      = "root:root"
 	calicoAzureFilePermission = 0600
 	calicoAzureFileTemplate   = `# Extra changes:
-#  - added "nodename_file_optional" set to true (can be removed on the next upgrade)
-#  - added resource limits
+#  - Added "nodename_file_optional" set to true (can be removed on the next upgrade).
```
xh3b4sd: Do we track the removal? I recall we had an issue for this. Then we maybe should align with the TODO issue link thingy here as well.

r7vme: Yes we have a task and we tracking this in team magic board. I'm taking care in few months after calico 3.2.x will become active to make sure all clusters have 3.2 (and kubelet with proper mount.)

https://github.com/giantswarm/giantswarm/issues/4113


i'll add

r7vme:
> Are there any implications we have to be aware off?

I hope not, tested this already and calico migrates just fine. We also have this in host clusters (will be rolled tomorrow morning), so if there will be something non-trivial we can have time to put a fix.

PR #393, opened 2018-10-27
update v5 to k8s 1.12.3

teemow:
* update cloudconfig to 3.7.0
* update container linux amis to 1855.5.0
* edit changelog in versionbundle

PR #394, opened 2018-10-27
fix new go formating

teemow:


teemow:
my build in #393 failed because of this.

PR #395, opened 2018-10-30
Add blobobject resource

corest:

This is gonna create cloudconfig objects in the storage account/container. There are no tests for now, because I need to check a general approach and then will work on tests. But I'd like to get some feedback for the current state.


service/controller/v5/resource/blobobject/resource.go
```diff
@@ -0,0 +1,197 @@
...
+
+type Resource struct {
+	hostAzureClientSetConfig client.AzureClientSetConfig
+	logger                   micrologger.Logger
+}
+
+type StorageClient struct {
+	accountsClient *storage.AccountsClient
+	containerURL   azblob.ContainerURL
+}
```
kopiczko: Can this be moved to `storage_client.go`? It would be easier to distinguish what is where. Now all is intertwined.

service/controller/v5/resource/blobobject/resource.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return r, nil
+}
+
+// EnsureDeleted is a noop since the deletion of blob is part is redirected to
+// the deletion of container deployment.
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	return nil
+}
```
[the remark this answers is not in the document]
corest: Oh, I've added it before finding CRUD resource as a time-saver :D

corest:
Tested manually, ignition resources were created without issues.

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+	resourceNotFoundError = "NotFound"
+	maxRetriesRequests    = 3
+)
+
+type BlobClient struct {
+	storageAccountsClient *storage.AccountsClient
+	containerURL          azblob.ContainerURL
+}
+
+func (c *BlobClient) BlobExists(ctx context.Context, blobName string) (bool, error) {
```
kopiczko: This is exported and not even used?

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+		return false, nil
+	}
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
+func (c *BlobClient) createBlockBlob(ctx context.Context, blobName string, payload string) (azblob.BlockBlobURL, error) {
```
kopiczko: This is used inside this package. But let's please think as it was used outside. We should have exported methods which are used outside this object and private methods which are used only internally. If you are afraid this object can be used outside package which I think it's legit just make it private by renaming it to `blobClient`.

service/controller/v5/resource/blobobject/spec.go
```diff
@@ -0,0 +1,13 @@
...
+	prefixMaster = "master"
+	prefixWorker = "worker"
+)
+
+type ContainerObjectState struct {
+	ContainerName      string
+	Body               string
+	Key                string
+	StorageAccountName string
+}
```
kopiczko: Let's please put that in `types.go`. It isn't any specification.

service/controller/v5/resource/blobobject/desired.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	cc, err := controllercontext.FromContext(ctx)
```
kopiczko:
```suggestion
	ctlCtx, err := controllercontext.FromContext(ctx)
```

service/controller/v5/resource/blobobject/desired.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	storageAccountName := key.StorageAccountName(customObject)
+	containerName := key.BlobContainerName()
+
+	output := map[string]ContainerObjectState{}
+
+	{
+		b, err := cc.CloudConfig.NewMasterCloudConfig(customObject)
```
kopiczko:
```suggestion
		b, err := ctlCtx.CloudConfig.NewMasterCloudConfig(customObject)
```

service/controller/v5/resource/blobobject/desired.go
```diff
@@ -0,0 +1,55 @@
...
+		output[k] = ContainerObjectState{
+			Body:               b,
+			ContainerName:      containerName,
+			Key:                k,
+			StorageAccountName: storageAccountName,
+		}
+	}
+
+	{
+		b, err := cc.CloudConfig.NewWorkerCloudConfig(customObject)
```
kopiczko:
```suggestion
		b, err := ctlCtx.CloudConfig.NewWorkerCloudConfig(customObject)
```

service/controller/v5/resource/blobobject/spec.go
kopiczko: Also I think it would be cooler to have real state type. In this case it would be `type State map[string] ContainerObjectState`.

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "looking for container objects")
```
kopiczko:
```suggestion
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	groupName := key.ClusterID(customObject)
+	storageAccountName := key.StorageAccountName(customObject)
+	containerName := key.BlobContainerName()
+
+	// if there is no storage account - return and wait for deployment to finish storage account operation.
+	_, err = sc.storageAccountsClient.GetProperties(ctx, groupName, storageAccountName)
+	if IsStorageAccountNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "blob object's storage account not found, no current objects present")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find container objects")
		r.logger.LogCtx(ctx, "level", "debug", "message", "blob object's storage account not found")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	// if there is no container account - return and wait for deployment to finish container operation.
+	containerURL, err := sc.getContainerURL(ctx, storageAccountName, groupName, containerName)
+	if IsContainerNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "blob object's container not found, no current objects present")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find blob object's container")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+	// if there is no storage account - return and wait for deployment to finish storage account operation.
+	_, err = sc.storageAccountsClient.GetProperties(ctx, groupName, storageAccountName)
+	if IsStorageAccountNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "blob object's storage account not found, no current objects present")
+		return nil, nil
+	}
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
```
kopiczko:
```suggestion

	r.logger.LogCtx(ctx, "level", "debug", "message", "finding blob object's container")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+	if IsContainerNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "blob object's container not found, no current objects present")
+		return nil, nil
+	}
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	sc.containerURL = containerURL
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "found the blob container")
```
kopiczko:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", "found blob object's container")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+
```
kopiczko:
```suggestion

	 r.logger.LogCtx(ctx, "level", "debug", "message", "finding container objects")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		output[object.Name] = ContainerObjectState{
+			Body:               string(body),
+			ContainerName:      containerName,
+			Key:                object.Name,
+			StorageAccountName: storageAccountName,
+		}
+	}
+
```
kopiczko:
```suggestion

	 r.logger.LogCtx(ctx, "level", "debug", "message", "found container objects")
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,76 @@
...
+	return output, nil
+
```
kopiczko:
```suggestion
```

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+	defer response.Body(retryReaderOptions).Close()
+	blobData, err := ioutil.ReadAll(response.Body(retryReaderOptions))
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return blobData, nil
+}
+
+func (c *BlobClient) getContainerURL(ctx context.Context, accountName, accountGroupName, containerName string) (azblob.ContainerURL, error) {
```
kopiczko: I think we should rename this to `func (c *BlobClient) Boot(ctx context.Context) error` and pass `accountName, accountGroupName, containerName` with `Config`. That would be the way we do such things usually.

client/azure_client_set.go
```diff
@@ -55,6 +56,8 @@ func (c AzureClientSetConfig) Validate() error {
 
 // AzureClientSet is the collection of Azure API clients.
 type AzureClientSet struct {
+	//StorageAccountsClient manages blobs in storage containers.
+	StorageAccountsClient *storage.AccountsClient
```
xh3b4sd: Lets please get the order straight here. Or at least, do not make it worse.

service/controller/v5/key/key.go
```diff
@@ -13,7 +13,9 @@ const (
 	installationTagName = "GiantSwarmInstallation"
 	organizationTagName = "GiantSwarmOrganization"
 
-	blobContainerName         = "ignition"
+	blobContainerName = "ignition"
+	// cloudConfigVersion used in blob object ignition name
```
xh3b4sd:
```suggestion
	// cloudConfigVersion is used in blob object ignition name.
```

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return blob, err
+}
+
+func (c *BlobClient) getAccountPrimaryKey(ctx context.Context, accountName, accountGroupName string) (string, error) {
+	keys, err := c.storageAccountsClient.ListKeys(ctx, accountGroupName, accountName)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+	return *(((*keys.Keys)[0]).Value), nil
```
xh3b4sd: That looks like we should have more error handling for the list length check and type assertion.

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+	p := azblob.NewPipeline(sc, azblob.PipelineOptions{})
+	u, _ := url.Parse(fmt.Sprintf(blobFormatString, accountName))
+	service := azblob.NewServiceURL(*u, p)
+	containerURL = service.NewContainerURL(containerName)
+	_, err = containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if IsContainerNotFound(err) {
+		return containerURL, nil
+	}
+
+	return containerURL, err
```
xh3b4sd:
```suggestion
	return containerURL, nil
```

service/controller/v5/resource/blobobject/blob_client.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (c *BlobClient) listBlobs(ctx context.Context, containerName string) (*azblob.ListBlobsFlatSegmentResponse, error) {
+	return c.containerURL.ListBlobsFlatSegment(
```
xh3b4sd: Let's please have dedicated error handling here.

service/controller/v5/resource/blobobject/blob_client.go
corest: It is exported as I expect this to be used in `instance` resource so that we can stop reconciling instance resource if blob with ignition is not ready yet

service/controller/v5/resource/blobobject/blob_client.go
corest: > But let's please think as it was used outside

But why? Also I don't want to make `BlobClient` private as it will be used within `instance` resource. How we can do that in a different way?

kopiczko: Then put it in a different package. We don't have dependencies between resources. That would lead to a spaghetti code and circular dependencies. My proposition is `v5/azureclient/`.

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,124 @@
...
+	sc := &BlobClient{
+		storageAccountsClient: storageAccountsClient,
+	}
+
+	groupName := key.ClusterID(customObject)
+	storageAccountName := key.StorageAccountName(customObject)
+	containerName := key.BlobContainerName()
+
+	// if there is no storage account - return and wait for deployment to finish storage account operation.
+	_, err = sc.storageAccountsClient.GetProperties(ctx, groupName, storageAccountName)
```
xh3b4sd: That is already done in `GetCurrentState`. What we should do there is canceling the reconciliation for the resource so the rest of the resource is not being executed on such a condition.

service/controller/v5/resource/blobobject/blob_client.go
corest: I could do that in that way. But every `BlobClient` is related to the specific blobObject as it operates over blob objec with `ContainerURL`. If that could be possible to a setup unique client for all objects - I would implement that in the `azure-operator/clients` package

service/controller/v5/resource_set.go
```diff
@@ -166,6 +167,25 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		}
 	}
 
+	var blobobjectResource controller.Resource
```
xh3b4sd:
```suggestion
	var blobObjectResource controller.Resource
```

service/controller/v5/resource_set.go
```diff
@@ -166,6 +167,25 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+
+			HostAzureClientSetConfig: config.HostAzureClientSetConfig,
+		}
+
+		ops, err := blobobject.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		blobobjectResource, err = toCRUDResource(config.Logger, ops)
```
xh3b4sd:
```suggestion
		blobObjectResource, err = toCRUDResource(config.Logger, ops)
```

service/controller/v5/resource_set.go
```diff
@@ -314,6 +334,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		namespaceResource,
 		serviceResource,
 		resourceGroupResource,
+		blobobjectResource,
```
xh3b4sd:
```suggestion
		blobObjectResource,
```

service/controller/v5/resource/blobobject/blob_client.go
corest: Hm, I took a look into code again and the move to the `v5/blobclient` might work

service/controller/v5/resource/blobobject/update.go
corest: How can I cancel reconciliation in the `GetCurrentState`?

service/controller/v5/resource/blobobject/blob_client.go
corest: API by spec returns two keys. There is no state you can delete key or so. So there are always two keys returned if no error

xh3b4sd: I understand that. We try to have our code to be defensive though and deal with more weirdnesses than we anticipate. This makes the platform more reliable.

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+		return microerror.Mask(err)
+	}
+
+	sc, err := azblob.NewSharedKeyCredential(c.storageAccountName, key)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	p := azblob.NewPipeline(sc, azblob.PipelineOptions{})
+	u, _ := url.Parse(fmt.Sprintf(blobFormatString, c.storageAccountName))
```
kopiczko: I suggest removing this const.

```suggestion
	u, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", c.storageAccountName))
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+	service := azblob.NewServiceURL(*u, p)
```
kopiczko:
```suggestion
	serviceURL := azblob.NewServiceURL(*u, p)
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+	containerURL = service.NewContainerURL(c.containerName)
```
kopiczko:
```suggestion
	containerURL = serviceURL.NewContainerURL(c.containerName)
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+	_, err = containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if IsContainerNotFound(err) {
+		return nil
```
kopiczko: What happens when this code returns `nil` and we call `c.containerURL.NewBlockBlobURL(blobName)`?

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+		return false, nil
+	}
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
+func (c *BlobClient) CreateBlockBlob(ctx context.Context, blobName string, payload string) (azblob.BlockBlobURL, error) {
```
kopiczko: Will this method override blob if it already exists? If so I suggest changing name to Put.

```suggestion
func (c *BlobClient) PutBlockBlob(ctx context.Context, blobName string, payload string) (azblob.BlockBlobURL, error) {
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if err != nil {
+		return listBlobs, microerror.Mask(err)
+	}
+
+	return listBlobs, nil
+}
+
+func (c *BlobClient) StorageAccountExists(ctx context.Context) (bool, error) {
+
```
kopiczko:
```suggestion
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
+func (c *BlobClient) getAccountPrimaryKey(ctx context.Context) (string, error) {
+	keys, err := c.storageAccountsClient.ListKeys(ctx, c.groupName, c.storageAccountName)
+	if len(*(keys.Keys)) == 0 {
```
kopiczko: Should this be checked before checking the error?

service/controller/v5/blobclient/blob_client.go
corest: It will fail as containerURL will be nil at that moment. But we don't do that, we just stop reconciling of the resource and waiting for container to be ready

corest:
Ready for next round :D

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,184 @@
...
+	return true, nil
+}
+
+func (c *BlobClient) getAccountPrimaryKey(ctx context.Context) (string, error) {
+	keys, err := c.storageAccountsClient.ListKeys(ctx, c.groupName, c.storageAccountName)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+	if len(*(keys.Keys)) == 0 {
+		return "", microerror.Maskf(err, "storage account key's list is empty")
```
kopiczko: The `err` here is `nil`.

```suggestion
		return "", microerror.Maskf(executionFailedError, "storage account key's list is empty")
```

service/controller/v5/blobclient/blob_client.go
corest: But in this case, we can just drop this additional check as on the next line we have the same masked error return. So It will be handled here https://github.com/giantswarm/azure-operator/pull/395/files#diff-0b6d094c4f73237335bfcd04bfb277c5R53

kopiczko: Yes, that's fine as well.

corest:
@giantswarm/sig-operator getting back to this. PTAL

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (c *BlobClient) BlobExists(ctx context.Context, blobName string) (bool, error) {
+	blobURL := c.containerURL.NewBlockBlobURL(blobName)
+
+	_, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{})
+	if IsBlobNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
```
xh3b4sd: I see this in a couple of places. Let's stick to `fmt`. 

```suggestion
	} else if err != nil {
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,181 @@
...
+		return microerror.Mask(err)
+	}
+
+	c.containerURL = containerURL
+
+	return nil
+}
+
+func New(config Config) (*BlobClient, error) {
```
xh3b4sd: Let's please move the constructor up. Methods should come after it.

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,181 @@
...
+	groupName             string
+	storageAccountName    string
+	storageAccountsClient *storage.AccountsClient
+
+	// containerURL is configured separately from the default
+	// parameters.
+	containerURL azblob.ContainerURL
+}
+
+func (c *BlobClient) Boot(ctx context.Context) error {
```
xh3b4sd: `Boot` is now called in `GetCurrentState`, `ApplyCreateChange` and `ApplyUpdateChange`. Basically twice in each loop, causing the resource to block longer than necessary and creating more pressure on the system. Let's make it to be called once. I would also question why we have to do this over and over again at runtime.

service/controller/v5/blobclient/error.go
```diff
@@ -0,0 +1,86 @@
...
+// IsBlobNotFound asserts blob not found error from upstream's API message.
+func IsBlobNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(microerror.Cause(err).Error(), "ServiceCode=BlobNotFound")
+}
+
+// IsContainerNotFound asserts container not found error from upstream's API message.
+func IsContainerNotFound(err error) bool {
```
xh3b4sd: There are matchers that do not have an associated error type. Let's please stay consistent and have them.

service/controller/v5/resource/blobobject/create.go
```diff
@@ -0,0 +1,97 @@
...
+			return microerror.Mask(err)
+		}
+	}
+
+	err = blobClient.Boot(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for key, containerObject := range containerObjectToCreate {
```
xh3b4sd: It is weird that you can iterate over something that is called "object".

service/controller/v5/resource/blobobject/create.go
```diff
@@ -0,0 +1,97 @@
...
+			StorageAccountsClient: storageAccountsClient,
+		}
+
+		blobClient, err = blobclient.New(c)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	err = blobClient.Boot(ctx)
```
xh3b4sd: At this point it is not really clear the client is needed at all. In general, let's make sure the work we do in resources is always useful. In case the managed storage thingy is already created, we boot the client and what not even if this is not necessary. Maybe will be addressed anyway when the other concern above is addressed, but just wanted to emphazise that again.

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "found blob object's container")
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding container objects")
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding container objects")
```
xh3b4sd: The logging is duplicated. Glitch in the matrix.

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,92 @@
...
+		output[object.Name] = ContainerObjectState{
+			Body:               string(body),
+			ContainerName:      key.BlobContainerName(),
+			Key:                object.Name,
+			StorageAccountName: key.StorageAccountName(customObject),
+		}
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "found container objects")
+	r.logger.LogCtx(ctx, "level", "debug", "message", "found container objects")
```
xh3b4sd: Same.

service/controller/v5/resource/blobobject/resource.go
```diff
@@ -0,0 +1,71 @@
+package blobobject
+
+import (
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2018-07-01/storage"
+	"github.com/giantswarm/azure-operator/client"
```
xh3b4sd: Let's make sure we stick to `fmt`. Please also check other places where this is not aligned. 

```suggestion

	"github.com/giantswarm/azure-operator/client"
```

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,117 @@
...
+
+			_, err := blobClient.PutBlockBlob(ctx, key, containerObject.Body)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created container object %#q", key))
+		} else {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not create container object %#q", key))
+			r.logger.LogCtx(ctx, "level", "debug", "message", "container object already exists")
```
xh3b4sd: I would drop this line.

service/controller/v5/blobclient/error.go
corest: By staying consistent do you mean just having them without usage? Those are an error which in my case do not go into the stack but just wait for next loop iteration, because of resources itself are created with arm templates

kopiczko: We had this discussion already. IMO having a not used error variable is confusing.

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, microerror.Mask(err)
+	}
+	desiredContainerObject, err := toContainerObjectState(desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the container objects should be updated")
+
+	updateState := map[string]ContainerObjectState{}
```
kopiczko: I don't like this is a map. There is an assumption that `ContainerObjectState.Key` is equal to the map key. I would rather use slice and search for the object based on its `Key` property. Performance isn't really crucial here.

service/controller/v5/blobclient/error.go
xh3b4sd: The scheme you scan without needing to think is usually error, matcher, error, matcher, and so on. When this changes you have to start to think. That takes time. That takes energy. It is the same for alphabetical order. We are as efficient as we are because we have a couple of consistency standards and all of them make it pretty easy and fast to work with a project even if you are not used to it on a daily basis. Errors and matchers come in pairs. Let's stick to the guidelines we give ourselves. When you want to change the guideline, get consensus on them first.

kopiczko: I didn't see a guideline saying having an error variable for the matcher is mandatory. So I don't think there is any consensus on that. IMO having a not used error variable is confusing. I won't fight for that though. It's a minor thing for me.

xh3b4sd: All the examples show always both. See https://github.com/giantswarm/fmt/blob/master/go/errors.md.

service/controller/v5/resource/blobobject/update.go
corest: Frankly speaking I didn't get this :/

kopiczko: I think updateState type should be `[]ContainerObjectState` instead of the map or `Key` should be removed from `ContainerObjectState`. I like the first more approach more.

corest:
I've refactored a bit boot of the blobclient. It looks even uglier now but doesn't call `Boot` multiple times. Feel free to drop a feedback on what is wrong with that approach.

xh3b4sd:
It is cool that we do not call something unnecessarily multiple times. What I still find unfortunate is that the resource has a client configured which is mutated internally in each reconciliation loop. This is a side effect caused by an implementation detail. The interface would be safer and cleaner if we would gather the URL once, either in the init function or separate resource and put it into the controller context. The URL can then be used where needed and given to the client in each call. We have this pattern already applied in a couple of places.

corest:
@xh3b4sd I agree with you. The thing is containerURL requires group/container names, which are constructed from ID. So to change this - I can pass into `blobobject` functions like `key.ToClusterID`, and then do casting of the azureconfig object every time I need that info. Is that a better approach? Otherwise I don't know how refactor this and can consider this as blocked for me because of lack knowledge. Also, discussion on theory doesn't really help as I just don't know how to implement this without example

service/controller/v5/resource/blobobject/update.go
corest: Ok, I got this. But with the lists, we then need nested loops for comparing current/desired states. Is that a nicer approach?

xh3b4sd:
Let's hang on Monday and we sort it out together. <3

corest:
I've updated PR with `containerurl` resource usage. Looks a bit clearer now. PTAL

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"io/ioutil"
+	"strings"
+
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	resourceNotFoundError = "NotFound"
```
xh3b4sd: That is not used for anything as it seems. 

```suggestion
```

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx,
+		strings.NewReader(payload),
+		azblob.BlobHTTPHeaders{
+			ContentType: "text/plain",
+		},
+		azblob.Metadata{},
+		azblob.BlobAccessConditions{},
+	)
+
+	return blob, err
```
xh3b4sd: Let's please have explicit error handling and then also masking for all errors.

service/controller/v5/resource/blobobject/create.go
```diff
@@ -0,0 +1,69 @@
...
+	containerObjectToCreate, err := toContainerObjectState(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, containerObject := range containerObjectToCreate {
+		if containerObject.Key != "" {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating container object %#q", containerObject.Key))
+
+			_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, containerURL)
```
xh3b4sd:
```suggestion
			_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, cc.ContainerURL)
```

service/controller/v5/resource/blobobject/create.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/giantswarm/azure-operator/service/controller/v5/blobclient"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	containerURL := cc.ContainerURL
```
xh3b4sd:
```suggestion
```

service/controller/v5/resource/blobobject/create.go
```diff
@@ -0,0 +1,69 @@
...
+
```
xh3b4sd:
```suggestion
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if cc.ContainerURL == nil {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "containerurl resource is not ready")
+		resourcecanceledcontext.SetCanceled(ctx)
```
xh3b4sd: Let's please have logging for the cancelation.

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,80 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding blob object's container")
+	// if here is no container account - return and wait for deployment to finish container operation.
+	containerExists, err := blobclient.ContainerExists(ctx, containerURL)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	if !containerExists {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find blob object's container")
+		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
xh3b4sd:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,87 @@
...
+	containerObjectToUpdate, err := toContainerObjectState(updateChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, containerObject := range containerObjectToUpdate {
+		if containerObject.Key != "" {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating container object %#q", containerObject.Key))
+
+			_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, containerURL)
```
xh3b4sd:
```suggestion
			_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, cc.ContainerURL)
```

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/giantswarm/operatorkit/controller"
+)
+
+func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	containerURL := cc.ContainerURL
```
xh3b4sd:
```suggestion
```

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,87 @@
...
+
```
xh3b4sd:
```suggestion
```

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,87 @@
...
+	containerObjectToUpdate, err := toContainerObjectState(updateChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, containerObject := range containerObjectToUpdate {
+		if containerObject.Key != "" {
```
xh3b4sd: I am wondering why we have to check this when we have something to iterate over. Is there any case where we append the item to the list while the key is empty?

service/controller/v5/resource/blobobject/update.go
```diff
@@ -0,0 +1,87 @@
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating container object %#q", containerObject.Key))
+
+			_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, containerURL)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created container object %#q", containerObject.Key))
```
xh3b4sd: Since this is about an update event, the wording should reflect that.

corest:
Everyone is happy ?)

service/controller/v5/blobclient/blob_client.go
```diff
@@ -0,0 +1,96 @@
...
+		ctx,
+		azblob.Marker{},
+		azblob.ListBlobsSegmentOptions{
+			Details: azblob.BlobListingDetails{
+				Snapshots: false,
+			},
+		})
+
+	if err != nil {
+		return listBlobs, microerror.Mask(err)
```
xh3b4sd: Let's please always either return the solid zero value or an error. It is too easy to fuck up the API and behaviour of the function otherwise. 

```suggestion
		return nil, microerror.Mask(err)
```

service/controller/v5/resource/blobobject/current.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	output := []ContainerObjectState{}
+	for _, object := range listBlobs.Segment.BlobItems {
+		body, err := blobclient.GetBlockBlob(ctx, object.Name, containerURL)
+
+		if err != nil {
+			return output, microerror.Mask(err)
```
xh3b4sd:
```suggestion
			return nil, microerror.Mask(err)
```

corest:
Merging as tested manually

PR #396, opened 2018-11-07
removes old cluster name config

xh3b4sd:
Basically vendor update and little modification to keep the code compiling. The update is necessary due to e2e test setup changes.


Gopkg.toml
```diff
@@ -59,7 +59,7 @@ required = [
   name = "github.com/giantswarm/e2esetup"
 
 [[constraint]]
-  branch = "master"
+  branch = "optional-label-selector"
   name = "github.com/giantswarm/e2etemplates"
```
xh3b4sd: See https://github.com/giantswarm/e2etemplates/pull/44.

integration/setup/common.go
```diff
@@ -41,7 +41,6 @@ func common(config Config) error {
 
 	{
 		c := chartvalues.CertOperatorConfig{
-			ClusterName:        env.ClusterID(),
```
xh3b4sd: That property is gone.

PR #397, opened 2018-11-08
add-calico-priority-class

calvix:
also adding back calico priorityClass as result of this discussions https://github.com/giantswarm/k8scloudconfig/pull/455#discussion_r231562647


service/controller/v4/cloudconfig/template.go
```diff
@@ -118,6 +119,7 @@ spec:
       # Since Calico can't network a pod until Typha is up, we need to run Typha itself
       # as a host-networked pod.
       serviceAccountName: calico-node
+      priorityClassName: system-cluster-critical
```
r7vme: Why we change active release? If it's smth urgently critical, let's create pattch release and go thru all clusters upgrade, if not then let's follow regular process - only changing WIP

PR #401, opened 2018-11-09
env: improve environment setting setup

kopiczko:
🏗 This is still WIP 🏗

I'd like to gather some feedback here.

🏗 This is still WIP 🏗

I was trying to incorporate `releaseindex` package from e2esetup but
fetching build information here with the error handling felt really
awkward. We don't really have a pattern established for initializing
complex structs. We agreed to not pass `context.Context` to the `New`
methods. `Boot` didn't really make sense so I came up with the
`Builder`.

It solves following problems:

- Dependencies between variables and order of `init()` calls.
- Passing context.Context.
- Error handling.


integration/env/new_azure_go_file_to_improve_the_diff_should_rplace_azure_go.go
```diff
@@ -0,0 +1,125 @@
+package env
```
kopiczko: This is a new file improve the diff. This is meant to replace `azure.go` in the end.

integration/env/new_azure_go_file_to_improve_the_diff_should_rplace_azure_go.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"os"
+	"strconv"
+
+	"github.com/giantswarm/azure-operator/integration/network"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+)
+
+var (
```
kopiczko: I'll create issues for the TODOs below.

integration/env/env.go
```diff
@@ -0,0 +1,72 @@
...
+	{
+		// ...
+	}
+
+	var azure Azure
+	{
+		c := azureBuilderConfig{
+			Logger: b.logger,
+
+			CircleBuildNumber: common.Circle.BuildNumber,
```
kopiczko: Here is an example on how to pass a value from common which wasn't really possible with `init()` functions.

integration/env/new_azure_go_file_to_improve_the_diff_should_rplace_azure_go.go
```diff
@@ -0,0 +1,125 @@
...
+	azureSubscriptionID = mustGetEnv("AZURE_SUBSCRIPTIONID")
+	azureTenantID       = mustGetEnv("AZURE_TENANTID")
+
+	azureGuestClientID       = mustGetEnv("AZURE_GUEST_CLIENTID")
+	azureGuestClientSecret   = mustGetEnv("AZURE_GUEST_CLIENTSECRET")
+	azureGuestSubscriptionID = mustGetEnv("AZURE_GUEST_SUBSCRIPTIONID")
+	azureGuestTenantID       = mustGetEnv("AZURE_GUEST_TENANTID")
+
+	// TODO this should be prefixed with AZURE_.
+	commonDomainResourceGroup = mustGetEnv("COMMON_DOMAIN_RESOURCE_GROUP")
```
kopiczko: This could be done in `Build()` but I think it's good to have that on the top of the file for visibility. This gives nice overview what variables are there which of them are required and which optional.

integration/env/env.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	b := &Bilder{
+		logger: config.Logger,
+	}
+
+	return b, nil
+}
+
+func (b *Builder) Build(ctx context.Context) (Env, error) {
```
kopiczko: We get context which can be used to call other APIs while constructing data types.

I also thought about `e2eSetup` where we cold provide `BuildAWS`, `BuildAzure` and `BuildKVM`. With with `init()` functions we'd either have a lot of code repetition or we'd have set e.g. set kvm variables in azure test. None of these approaches is good.

integration/env/new_azure_go_file_to_improve_the_diff_should_rplace_azure_go.go
kopiczko: I think I should put that comment in the code.

kopiczko: Comment  added.

integration/env/helper.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+func getEnv(name string) string {
+	return os.Getenv(name)
+}
+
+func mustGetEnv(name string) string {
```
xh3b4sd: Since we refactor this and make it nicer we can return an error to the caller and only panic in a couple of exclusive cases at the very core of the program initialization. This would then maybe better called something like `getEnvOrError`. Or better, simply `getEnv` and return a specific error type when the value is empty, which we can ignore if we want to, or deal with if we should.

kopiczko: I thought about this as well. In that case would get like 5 extra lines for each env var and we loose immediate visibility which of them are optional. And what can really we do if the required value isn't initalized? Yes we can exit with 1 and print error message which is bit more elegant but for me it isn't enough to pay the price of code explosion.

kopiczko: But in general it makes sense. Otherwise `BuildAWS`, `BuildAzure`, `BuildKVM` won't really work as they will panic because of variables missing for other provider. I'll change that.

integration/env/new_azure_go_file_to_improve_the_diff_should_rplace_azure_go.go
```diff
@@ -0,0 +1,165 @@
...
+		getEnvRequired(EnvVarAzureGuestClientID, &azureGuestClientID),
+		getEnvRequired(EnvVarAzureGuestClientSecret, &azureGuestClientSecret),
+		getEnvRequired(EnvVarAzureGuestSubscriptionID, &azureGuestSubscriptionID),
+		getEnvRequired(EnvVarAzureGuestTenantID, &azureGuestTenantID),
+
+		getEnvRequired(EnvVarCommonDomainResourceGroup, commonDomainResourceGroup),
+	)
+	if err != nil {
+		return Azure{}, microerror.Mask(err)
+	}
```
kopiczko: The alternative code which I really don't like is below. I could do stuff like 

```go
azure := Azure{}

getEnvOptional(EnvVarAzureCIDR, &azure.CIDR)
```

But there are too many cases where we will use environment variables values only in `Build` methods. 

The alternative code below. It would look much worse if optional and required envs were intertwined.

```go
        azureCIDR := getEnvOptional(EnvVarAzureCIDR)
        azureCalicoSubnetCIDR := getEnvOptional(EnvVarAzureCalicoSubnetCIDR )
        azureMasterSubnetCIDR := getEnvOptional(EnvVarAzureMasterSubnetCIDR)
        azureVPNSubnetCIDR := getEnvOptional(EnvVarAzureVPNSubnetCIDR)
        azureWorkerSubnetCIDR := getEnvVarOptional(EnvVarAzureWorkerSubnetCIDR)

        azureClientID, err := getEnvRequired(EnvVarAzureClientID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureClientSecret, err := getEnvRequired(EnvVarAzureClientSecret)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureLocation, err := getEnvRequired(EnvVarAzureLocation)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureSubscriptionID, err := getEnvRequired(EnvVarAzureSubscriptionID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureTenantID, err := getEnvRequired(EnvVarAzureTenantID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }

        azureGuestClientID, err := getEnvRequired(EnvVarAzureGuestClientID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureGuestClientSecret, err := getEnvRequired(EnvVarAzureGuestClientSecret)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureGuestSubscriptionID, err := getEnvRequired(EnvVarAzureGuestSubscriptionID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
        azureGuestTenantID, err := getEnvRequired(EnvVarAzureGuestTenantID)
        if err != nil {
                return Azure{}, microerror.Mask(err)
        }
```

kopiczko: @xh3b4sd PTALA

xh3b4sd: I would really like to not go with pointers. I hate it. Anything can happen with this. Instead, lets please have a clear interface. Input. Output. Easy.

kopiczko: I don't follow. What can go wrong here? This is the same approach as in any command line flag library in Go.

xh3b4sd: I am not a fan of the pattern in general. In this specific example, not so much can go wrong. I personally simply do not want to apply the reference pattern in general when I do not have to, because it forces me to think more and adds complexity in general. The value can be assigned at runtime at any given point in time and only because the function returns does not mean the value does not change anymore. This is already really painful, confusing and dangerous with the command line flag handling. I do not mind the extra error handling when this means the code is simpler and more straight forward. There will also be alternatives in further go versions. When you want to do it that way, this is fine. I am done here.

kopiczko: I'm not so afraid of pointers and I encourage you to embrace them as well. I think simpler code means less errors. Even if it uses some potentially dangerous features. I've seen a code when dealing a lot of copies gives similar headaches as dealing with pointers. You need to track both almost the same way.

But I played with it and it really doesn't give much gain. So I wen with error checking in place. I just made `getEnvVarOptional` return error which is always nil just to visually align things.

kopiczko:
I abandon the idea for now due to lack of time. It still needs some thinking.

PR #402, opened 2018-11-12
v3, v3patch1, v4, v5: fix vmss instances paging

kopiczko:
This effectively blocked us from making upgrades on clusters with more
that 10 workers.

It also fixes a problem with instance names that have ID > 10. Apparently base 36 is used to create those names. For more details see https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name.

### DIFF
--- a/service/controller/v3patch1/key/key.go
+++ b/service/controller/v3patch1/key/key.go
@@ -2,6 +2,7 @@ package key
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/Azure/go-autorest/autorest/to"
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
@@ -217,7 +218,8 @@ func WorkerSubnetName(customObject providerv1alpha1.AzureConfig) string {
 }
 
 func MasterInstanceName(customObject providerv1alpha1.AzureConfig, instanceID string) string {
-	return fmt.Sprintf("%s-master-%06s", ClusterID(customObject), instanceID)
+	idB36 := vmssInstanceIDBase36(instanceID)
+	return fmt.Sprintf("%s-master-%06s", ClusterID(customObject), idB36)
 }
 
 // MasterNICName returns name of the master NIC.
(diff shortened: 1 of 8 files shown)


xh3b4sd:
We are not able to update clusters affected by that without the backport. We have to solve a problem. This is the right fix for it IMO.

service/controller/v3patch1/key/key.go
```diff
@@ -389,9 +391,27 @@ func VPNGatewayName(customObject providerv1alpha1.AzureConfig) string {
...
+	if err != nil {
+		// This must be an int according to the documentation linked below.
+		//
+		// We are panicking here to make the API nice. If this is not
+		// an int there is nothing we can really do and we need to
+		// redesign `instance` resource.
+		//
+		//	https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name
+		//
+		panic(fmt.Sprintf("expected VMSS instanceID to be a positive integer number but got %#q", instanceID))
```
kopiczko: Just saying. There is a `panic` in this code. I'm not going to address it ever and if this PR is blocked only because of this panic I'll still merge it. I want to resolve issue we have now. Also I don't see any good handling so I don't wan to pollute the code. Feel free to change it if it hurts you.

xh3b4sd: Please create a TODO issue which we track in the board and go.

kopiczko:
Verified to work on victory. Note the instance name `2krpk-worker-00000d` ends with a letter.

```
D 11/12 17:57:11 default/azureconfigs/2krpk instancev3patch1 found instance '2krpk-worker-00000d' has to be drained | azure-operator/service/controller/v3patch1/resource/instance/create.go:435 | event=update | version=39422417
```

service/controller/v5/key/key.go
```diff
@@ -400,9 +402,27 @@ func VPNGatewayName(customObject providerv1alpha1.AzureConfig) string {
...
+	return fmt.Sprintf("%s-worker-%06s", ClusterID(customObject), idB36)
 }
 
 func WorkerVMSSName(customObject providerv1alpha1.AzureConfig) string {
 	return fmt.Sprintf("%s-worker", ClusterID(customObject))
 }
+
+func vmssInstanceIDBase36(instanceID string) string {
+	i, err := strconv.ParseUint(instanceID, 10, 64)
+	if err != nil {
```
kopiczko:
```suggestion
	if err != nil {
		// TODO Avoid panic call if feasible.
		//
		//	See https://github.com/giantswarm/giantswarm/issues/4674
		//
```

kopiczko:
CI is green in the previous run.

PR #405, opened 2018-11-13
fixes deployment updates when deployments got deleted in the azure console

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4363.


xh3b4sd:
We are backporting the fix to a couple of versions here. What I did during my testing was to delete the deployment `virtual_network_setup`. The operator tried to catch up but couldn't. The fix here enables the operator to catch up again and fix the deleted deployment. Meanwhile I verified nothing happens to the workloads. The tenant cluster just keeps operating.

xh3b4sd:
Scaling test passed. Good enough for me. Rolling.

PR #406, opened 2018-11-15
moves status magic to separate files

xh3b4sd:
Just some boyscouting. In the next PR I want to address some improvements for the status management.

### DIFF
--- a/service/controller/v5/resource/instance/create.go
+++ b/service/controller/v5/resource/instance/create.go
@@ -18,16 +18,6 @@ import (
 )
 
 const (
-	Stage = "Stage"
-)
-
-const (
-	DeploymentInitialized  = "DeploymentInitialized"
-	InstancesUpgrading     = "InstancesUpgrading"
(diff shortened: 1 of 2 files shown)


xh3b4sd:
Why does it make it harder for you that way?

kopiczko:
Let's say we have a PR with code calling `setResourceStatus`. I don't know what `setResourceStatus` so I click `View` and search the file for `setResourceStatus`. If this isn't in the same file I have to do other exercises in order to find it out.

I understand people have different workflows and approaches. And brains in general. So if the team likes it I commit.

xh3b4sd:
Most of the time, most of the code presented in a pull request, is never in the same file. We have vastly complex projects with millions of lines of code involved. We will almost never see the code we need to see when viewing the file being changed.

xh3b4sd:
Fishing for LGTM.

xh3b4sd:
Talked with Pawel. We are good.

PR #407, opened 2018-11-15
fixes duplicated resource status from older versions

xh3b4sd:
This PR enables the operator to do some proper housekeeping with the resource statuses. This is what we kind of see right now. 

```
  resources:
  - conditions:
    - lastHeartbeatTime: 0001-01-01T00:00:00Z
      lastTransitionTime: 0001-01-01T00:00:00Z
      status: DeploymentInitialized
      type: Stage
    status: instancev3
  - conditions:
    - lastHeartbeatTime: 0001-01-01T00:00:00Z
      lastTransitionTime: 0001-01-01T00:00:00Z
      status: DeploymentInitialized
      type: Stage
    status: instancev4
```

This is what the operator does now with this change when transitioning to other versions. Tests have been extended to cover certain cases. 

```
  resources:
  - conditions:
    - lastHeartbeatTime: 0001-01-01T00:00:00Z
      lastTransitionTime: 0001-01-01T00:00:00Z
      status: DeploymentInitialized
      type: Stage
    status: instancev4
```


service/controller/v5/resource/instance/status.go
```diff
@@ -160,3 +161,19 @@ func resourceStatusExists(customObject providerv1alpha1.AzureConfig, t string) b
...
+// unversionedName provides a comparable name without the exact version number
+// suffix. When the resource name should ever change, the mechanism of managing
+// the resource status has to be adapted accordingly. Otherwise unversionedName
+// will return an empty string and the mechanism will break in unexpected ways.
+func unversionedName(name string) string {
+	if !strings.HasPrefix(name, "instancev") {
+		return ""
+	}
+
+	return "instancev"
```
kopiczko: I think it would be good to drop the `v4` thing at all. I doesn't fit really well versioning of operator roadmap. When this is in place I think the resource will be just called `instance`. What I thought long term might make sense is adding `version` and `controller` to the log context in `operatorkit` instead.

xh3b4sd: Agreed. This code here is the perfect transitioning tool for it. We can then just change the name and once migrated remove the obsolete magic.

kopiczko: Can't we change the value of the `status` to `instance` right away?

PR #408, opened 2018-11-19
deletes k8s service when cluster is being deleted

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4604.


xh3b4sd:
The service is created inside of the namespace. When the namespace is gone the service would be as well. When we delete anyway we would catch the not found error.

PR #414, opened 2018-11-28
update readme service principal command

r7vme:
By default it's only one year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Otherwise create a service provider with the azure-operator role (replace
 SUBSCRIPTION_ID):
 
 ```bash
-export CODENAME=ghost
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +35,8 @@ Otherwise create a service provider with the azure-operator role (replace
 SUBSCRIPTION_ID):
 
 ```bash
-az ad sp create-for-rbac -n azure-operator-sp --role="azure-operator" --scopes="/subscriptions/${SUBSCRIPTION_ID}"
+export CODENAME=ghost
```
calvix: maybe let's use more generic name `cluster1` rather than `ghost`

r7vme: makes sense, done

PR #421, opened 2018-11-30
v5/key: fix storage account name for integration tests

kopiczko:
In integration tests were hitting errors like below. This was due to
invalid storage account name. This is fortunately used only in non
active release so we can change it w/o consequences.

```
{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.","details":[{"code":"BadRequest","message":"{\r\n \"error\": {\r\n \"code\": \"AccountNameInvalid\",\r\n \"message\": \"ci-wip-a9f28-c1c86gsstorageaccount is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only.\"\r\n }\r\n}"}]}
```


xh3b4sd:
What does this mean for existing clusters that we upgrade now?

kopiczko:
> What does this mean for existing clusters that we upgrade now?

As written in the description this is non-active release so I think it's fine to not worry too much.

xh3b4sd:
The release will be activated in the future and we will upgrade clusters to this version and then it still has to work out. So there is an implication. When the operator just sorts it out and there is no problem with renaming this key, let's go. But we should make sure this is the case.

kopiczko:
We don't have service accounts in v4 so there is nothing to handle AFAIK.

xh3b4sd:
We talked. All good.

kopiczko:
Tested on `ghost`. It works. It was aiming to fix e2e tests but it doesn't. Hopefully we are one step closer. Merging.

PR #429, opened 2018-12-04
Prevent emitting duplicate metrics

marians:
For https://github.com/giantswarm/giantswarm/issues/4601

### DIFF
--- a/service/collector/deployment.go
+++ b/service/collector/deployment.go
@@ -91,7 +91,7 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 	}
 	defer watcher.Stop()
 
-	alreadyEmitted := map[string]bool{}
+	alreadyEmitted := map[string]struct{}{}
 
 	for {
 		select {
@@ -126,7 +126,7 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 							if found {
 								continue
 							}
-							alreadyEmitted[k] = true
+							alreadyEmitted[k] = struct{}{}
 
 							ch <- prometheus.MustNewConstMetric(
 								deploymentDesc,
@@ -144,7 +144,7 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 							if found {
 								continue
 							}
-							alreadyEmitted[k] = true
+							alreadyEmitted[k] = struct{}{}
 
 							ch <- prometheus.MustNewConstMetric(
 								deploymentDesc,
@@ -162,7 +162,7 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 							if found {
 								continue
 							}
-							alreadyEmitted[k] = true
+							alreadyEmitted[k] = struct{}{}
 
 							ch <- prometheus.MustNewConstMetric(
 								deploymentDesc,
--- a/service/collector/deployment.go
+++ b/service/collector/deployment.go
@@ -2,6 +2,7 @@ package collector
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
@@ -90,6 +91,8 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 	}
 	defer watcher.Stop()
 
+	alreadyEmitted := map[string]bool{}
+
 	for {
 		select {
 		case event, ok := <-watcher.ResultChan():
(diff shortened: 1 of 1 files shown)


service/collector/deployment.go
```diff
@@ -90,6 +91,8 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 	}
 	defer watcher.Stop()
 
+	alreadyEmitted := map[string]bool{}
```
xh3b4sd: FYI the more memory efficient version and best practise for these kinds of usages of maps is to use `struct{}` as value.

JosephSalisbury:
I don't see why we're emitting the same metric multiple times - could you shed some light?

marians:
No, I can't

JosephSalisbury:
Please try - this feels like we're patching over the issue without understanding it.

xh3b4sd:
We have the same problem here as we had in `operatorkit`, where the fix also comes from as it looks to me. Here we use a watcher interface, which provides k8s api runtime object events. When you read an event of object A and object A is updated or deleted since the time you started watching the stream, you get a second event for object A. This here is the right approach to fix the issue as long as we use the watcher instead of a lister.

JosephSalisbury:
Oh, I see. We watch the CRs to get the cluster IDs, so we can then list the deployments.

Why can't we list the deployments directly?

xh3b4sd:
We probably could but there are all kinds of deployments eventually not related to the actual installation or tenant clusters.

rossf7:
Throwing this in here but we had this problem with cluster-operator and configmaps. Helm was creating hundreds of them and cluster-operator was fetching them as part of the current state.

@tuommaki fixed it by using the service-type and managed-by labels. I really like the approach and maybe we could use it more.

https://github.com/giantswarm/cluster-operator/blob/22868d3e46b063083e35d9edf96623419ba29c57/pkg/v9/configmap/current.go#L32-L34

Not into the details of whether its right in this case but I thought I'd mention it.

marians:
Handed PR over to @JosephSalisbury

marians:
@JosephSalisbury Are you aware that this still exists?

JosephSalisbury:
I was not. I'll take care.

service/collector/deployment.go
```diff
@@ -90,6 +91,8 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
 	}
 	defer watcher.Stop()
 
+	alreadyEmitted := map[string]struct{}{}
```
TheoBrigitte: Why using a `map[string]struct{}{}`, it would be simpler to use a `map[string]bool` IMO and assign true when metric is emitted. WDYT ?

service/collector/deployment.go
```diff
@@ -102,43 +105,74 @@ func (d *Deployment) Collect(ch chan<- prometheus.Metric) error {
...
+			clusterID := key.ClusterID(customObject)
+
 			{
 				deploymentsClient, err := d.getDeploymentsClient(customObject)
 				if err != nil {
 					return microerror.Mask(err)
 				}
 
-				r, err := deploymentsClient.ListByResourceGroup(context.Background(), key.ClusterID(customObject), "", to.Int32Ptr(100))
+				r, err := deploymentsClient.ListByResourceGroup(context.Background(), clusterID, "", to.Int32Ptr(100))
```
TheoBrigitte: Using [`ListByResourceGroupComplete`](https://godoc.org/github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources#Client.ListByResourceGroupComplete) allow to write one for loop less.

```go
r, _ := deploymentsClient.ListByResourceGroup(...)
for r.NotDone() {
    v := range r.Value()

    ...
}
```

xh3b4sd:
What's up with this PR? Can it be merged or closed?

JosephSalisbury:
I take care this week.

JosephSalisbury:
@kopiczko this is the older PR. we should configure the clientsets, instead of deduplicating emitted metrics. Closing here.

xh3b4sd:
So instead of just merging and having the problem fixed we close the PR and do not fix the problem? I don't think this is the best we can do. We could actually just merge and improve the implementation later regardless, if this is what we want.

kopiczko:
Why do we watch the CRs instead of listing them?

xh3b4sd:
I think this is for some legacy reasons. When we did not have listers for all types we where using watchers. This came to some interfaces and then we only ever used watchers. It may came from the `operatorkit` informer implementation. Using listers would make total sense for collectors. We should then document such best practise and tell everybody. I would also weigh in on some collective refactoring efforts if this is something we want.

kopiczko:
> @kopiczko this is the older PR. we should configure the clientsets, instead of deduplicating emitted metrics. Closing here.

I checked the code. Clientset configuration isn't the root cause of the issue because we list resources by resource group so this should be fine if we do listings per `azureconfig`.

I agree with @xh3b4sd here:

> Here we use a watcher interface, which provides k8s api runtime object events. When you read an event of object A and object A is updated or deleted since the time you started watching the stream, you get a second event for object A. This here is the right approach to fix the issue as long as we use the watcher instead of a lister.

I don't think using watcher makes sense here in general, but this line is concerning:

https://github.com/giantswarm/azure-operator/blob/62200e78d8fd8d38805247f70fc41776d2e11df5/service/collector/deployment.go#L184-L186

When we have context passed into `Collect` (see https://github.com/giantswarm/giantswarm/issues/4909) we can do timeouts using the context.

But this is concerning in general:
- Why only one second?
- Is this enough time to call Azure APIs?

xh3b4sd:
The one second timeout is the hack to make the watcher behave like a lister. The watcher streams runtime objects and once no object is streamed anymore all known runtime objects got streamed. In case we do not see any runtime object anymore after one second, we assume we "listed" all of them.

kopiczko:
This sounds like we deifiantelly should just list the objects.

PR #432, opened 2018-12-11
remove old versions

teemow:


service/collector/deployment.go
```diff
@@ -15,7 +15,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/giantswarm/azure-operator/client"
-	"github.com/giantswarm/azure-operator/service/controller/v4/key"
+	"github.com/giantswarm/azure-operator/service/controller/v4patch1/key"
```
teemow: This is rather ugly. @xh3b4sd this was added by you. what's the plan here? wait until versioning is done properly?

xh3b4sd: I know. We have this in a couple of places. I tried to come up with something better but then thought it is not worth the effort. The operator versioning will certainly make it less weird.

xh3b4sd: Note that we have such things in AWS as well.

xh3b4sd:
Note that the build fails because of some import weirdness.

Issue #311, opened 2018-08-06
VMs are not created until VPN gateway is created


r7vme:
While testing `1.1.0` wip release, i've noticed that azure-operator does not try to create VMs (master/worker vmss resources) until initial deployment (includes VPN gateway) is not finished.

Problem that VPN gateway takes up to 45 minutes (in reality 20-30 minutes) to be created. We don't really blocked by VPN gateway imo as it's used by ops access and monitoring only.

r7vme:
I think we either can move VPN gateway to separate ARM template or reasses the reason, why we postponing VMs creation.

xh3b4sd:
@TheoBrigitte is working in this direction and he actually implemented this. Would be cool to check back here and see if we can improve the situation. I noticed the longer startup times and it would be nice to get it lower again.

xh3b4sd:
It is not about the template but about the scheduling and execution of the operatorkit resource implementations and how they interfere. Since the VPN gateway resource got introduced the times went up a lot and I recall the deployment resource and/or the instance resource get locked or blocked due to that somehow. I am not in the details there, just thought you have a vision or something.

xh3b4sd:
@r7vme for the future please create all issues in giantswarm/giantswarm so we have all the usual labels available. I was looking for this one and couldn't find it. 

@TheoBrigitte IIRC you said you wanted to address this in the scope of some other work you were doing. Can you say something about your timeline? Cluster creation taking super long is a blocker for all devs and everybody would greatly benefit from your work here.

TheoBrigitte:
This is actually the next issue on my list. Will try to find some time for this next week, if not I will start after I am back from holiday on September 19th.

xh3b4sd:
Is this still the next issue on your list? :D

TheoBrigitte:
Nope, priority has shifted towards PMs.

I know this issue is a big paint point for azure-operator as it makes it very slow now to spin up tenant cluster.

xh3b4sd:
How difficult can this be? Do you want to pair on it again?

TheoBrigitte:
VPN gateway is now handled by the vpn resource which happens in parallel of other resources.
This brings down tenant cluster creation down to approx. 10mn.


```
# cluster created at 10:33:17
$ k --context giantswarm-godsmack get azureconfig -ndefault o9b05 -oyaml|yq -y -r '.status.cluster.conditions[]'
lastHeartbeatTime: '2018-12-21T10:33:17.026566'
[...]

PR #434, opened 2018-12-11
add vpn template

TheoBrigitte:
Towards: https://github.com/giantswarm/azure-operator/issues/311

This is an attempt to automates embedding of json files into Go source code.

A simpler approach would be to directly write a file like `virtual_network_gateway.json.go` but this makes manipulation of json tedious.

My motivation are:
- easier manipulation of json files (syntax highlighting, jq processing,  ...)
- keep the benefit of go constants

It requires to run `go generate` prior to `go build` in order to generate the go source file. This could be added as an new step in `architect` workflow.


service/controller/v5/resource/vpn/deployment.go
```diff
@@ -1,5 +1,7 @@
 package vpn
 
+//go:generate go run github.com/logrusorgru/textFileToGoConst -in virtual_network_gateway.json -c vpn_gateway_json
```
kopiczko: For the record. We talked with @TheoBrigitte. This would be fine (at least with me) but there are two problems.

* Easy to solve problem

We'd need to ensure we control the version of `textFileToGoConst` we run. We could solve that by just forking the repo to make sure.

* Hard to solve problem

In `main.json` we import templates from public URL. We could use go templating and create a big single template. The problem here is Azure is far from being perfect with reporting template errors. When we have them split into multiple files it's easier to find a problem in smaller scope while having a cryptic error message.

Bottom line:
We should find a solution which can be applied to the most complex case first. Which is `main.json` template.

TheoBrigitte:
No progress here.

PR #437, opened 2018-12-13
updates vendor deps

xh3b4sd:
This PR is to get the heartbeat updates via the `statusresource` in. See also https://github.com/giantswarm/statusresource/pull/22.


xh3b4sd:
Verified cluster creation manually on `godsmack`. It took 90 minutes to create a cluster. I bet Azure is slow which is why all the e2e tests time out.

PR #439, opened 2018-12-13
unify template URI functions across resources

TheoBrigitte:
Towards: https://github.com/giantswarm/azure-operator/issues/311

deployment and instance resources are both defining the same set of functions to format arm template URI.

I aligned them here under the key package.


service/controller/v5/resource/deployment/deployment.go
```diff
@@ -27,7 +31,7 @@ func (r Resource) newDeployment(ctx context.Context, customObject providerv1alph
 		"kubernetesAPISecurePort": key.APISecurePort(customObject),
 		"masterSubnetCidr":        sc.AzureNetwork.Master.String(),
 		"storageAccountName":      key.StorageAccountName(customObject),
-		"templatesBaseURI":        baseTemplateURI(r.templateVersion),
+		"templatesBaseURI":        key.TemplateBaseURI(r.templateVersion, "deployment"),
```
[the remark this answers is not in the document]
TheoBrigitte: good point

TheoBrigitte:
Fishing for review.

service/controller/v5/key/key.go
```diff
@@ -65,6 +66,24 @@ func APISecurePort(customObject providerv1alpha1.AzureConfig) int {
...
+// ARMTemplateURI returns URI to a resource's template file.
+// The returned URI point to file hosted on github in this repository.
+//
+// version refers to a branch or commit.
+// resource refers to directory name of the resource.
+// template refers to filename under resource's template folder.
+//
+// e.g. ARMTemplateURI("master", "deployment", "main.json")
+func ARMTemplateURI(version, resource, template string) string {
+	return fmt.Sprintf(templateURIFmt, version, resource, template)
```
kopiczko: Can we make `templateURIFmt` "inline"? I personally don't see a reason to jump around a look for the value.

service/controller/v5/resource/deployment/deployment.go
```diff
@@ -27,7 +27,7 @@ func (r Resource) newDeployment(ctx context.Context, customObject providerv1alph
 		"kubernetesAPISecurePort": key.APISecurePort(customObject),
 		"masterSubnetCidr":        sc.AzureNetwork.Master.String(),
 		"storageAccountName":      key.StorageAccountName(customObject),
-		"templatesBaseURI":        baseTemplateURI(r.templateVersion),
+		"templatesBaseURI":        key.ARMTemplateBaseURI(r.templateVersion, "deployment"),
```
kopiczko: When we are here we can use this in the template instead:

```
 "vmssTemplateUrl": "[uri(deployment().properties.templateLink.uri, 'vmss.json')]"
```

See: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-linked-templates#using-variables-to-link-templates

TheoBrigitte: Nice simplification

PR #445, opened 2019-01-08
adds 'Running' provisioning state to deployment collector

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4920. The PM is about stuck deployments. The idea is to track the `Updating` state and have an alert when it appears longer than two hours or something like this.


xh3b4sd:
OMFG I tried `Deploying` and `Updating`, but it is apparently `Running`. So here we go. 

screenshot 2019-01-08 at 12 11 00

PR #446, opened 2019-01-08
Remove nodename_file_optional from calico configmap

r7vme:
related: https://github.com/giantswarm/giantswarm/issues/4113

### DIFF
--- a/vendor/github.com/giantswarm/backoff/backoff.go
+++ b/vendor/github.com/giantswarm/backoff/backoff.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	LongMaxWait   = 40 * time.Minute
+	LongMaxWait   = 80 * time.Minute
 	MediumMaxWait = 10 * time.Minute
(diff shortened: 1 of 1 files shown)


vendor/github.com/giantswarm/backoff/backoff.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	LongMaxWait   = 40 * time.Minute
+	LongMaxWait   = 80 * time.Minute
```
tuommaki: There's no changes for `Gopkg.toml` or `Gopkg.lock` in this PR. Is this maybe some kind of leftover when the only other change is removal of `nodename_file_optional`?

r7vme: This was done for testing only, i'm not going to merge this.

We discussing here https://gigantic.slack.com/archives/C2MS2VB37/p1547199975002200

PR #454, opened 2019-02-06
Add containerurl resource

corest:
Part of refactoring for this https://github.com/giantswarm/azure-operator/pull/395

`containerurl` resource creates `containerurl` to manager container with ignition blobs. This containerurl is then shared via context with `blobobject`


service/controller/v5/resource/containerurl/resource.go
```diff
@@ -0,0 +1,101 @@
...
+	if IsStorageAccountNotFound(err) {
+		return false, nil
+	} else if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
+func (r *Resource) getAccountPrimaryKey(ctx context.Context, groupName, storageAccountName string) (string, error) {
```
xh3b4sd: Can you please flip the method with `storageAccountExists` to have it ordered alphabetically?

service/controller/v5/resource/containerurl/update.go
```diff
@@ -0,0 +1,26 @@
+package containerurl
+
+import (
+	"context"
+
+	"github.com/giantswarm/azure-operator/service/controller/v5/key"
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) EnsureUpdated(ctx context.Context, obj interface{}) error {
```
xh3b4sd: That method does not do anything. I would say we drop it. When you have the normal resource it is only `EnsureCreated` and `EnsureDeleted`. When the container URL is necessary for deletion, the latter should be sorted as well.

service/controller/v5/resource_set.go
```diff
@@ -166,6 +167,26 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		}
 	}
 
+	var containerURLResource controller.Resource
+	{
+		azureClients, err := client.NewAzureClientSet(config.HostAzureClientSetConfig)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		storageAccountsClient := azureClients.StorageAccountsClient
```
xh3b4sd: Are you sure this is the right way to configure the clients? I remember that is different with BYOC. Can you double check? I feel like I miss something here.

corest: I see how whole `HostAzureClientsetConfig` is passed into resources. But I need only StorageAccountsClient. And it is implemented in the same as all other clients, so I don't see how that can be different even with byoc case. Maybe @kopiczko or @TheoBrigitte can suggest something if this approach is wrong.

kopiczko: The question is if this should really be the client talking to control plane subscription. Shouldn't it be a client talking to  tenant subscription?

service/controller/v5/resource/containerurl/resource.go
```diff
@@ -0,0 +1,102 @@
...
+	p := azblob.NewPipeline(sc, azblob.PipelineOptions{})
+	u, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", storageAccountName))
+	serviceURL := azblob.NewServiceURL(*u, p)
+	containerURL := serviceURL.NewContainerURL(containerName)
+	_, err = containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "setting containerurl to context")
```
kopiczko: I think this log should be at the top of the method and `set containerurl to context` log just before the return.

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
```
kopiczko:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", "finding storage account")

```

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,45 @@
...
+	groupName := key.ClusterID(customObject)
+	storageAccountName := key.StorageAccountName(customObject)
+
+	storageAccountExists, err := r.storageAccountExists(ctx, groupName, storageAccountName)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if !storageAccountExists {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "storage account not found")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "did not find storage account")
```

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,45 @@
...
+		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil
+	}
+
+	err = r.addContainerURLToContext(ctx, containerName, groupName, storageAccountName)
+	if IsStorageAccountNotProvisioned(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "storage account not provisioned")
+		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,45 @@
...
+	if !storageAccountExists {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "storage account not found")
+		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
+		return nil
+	}
+
+	err = r.addContainerURLToContext(ctx, containerName, groupName, storageAccountName)
+	if IsStorageAccountNotProvisioned(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "storage account not provisioned")
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "found storage account is not provisioned")
```

corest:
Rewrote it with approach, used in other resources

service/controller/v5/resource/containerurl/resource.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return newResource, nil
+}
+
+func (r *Resource) Name() string {
+	return Name
+}
+
+func (r *Resource) addContainerURLToContext(ctx context.Context, containerName, groupName, storageAccountName, primaryKey string) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "set containerurl to context")
```
kopiczko:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", "setting containerurl to context")
```

service/controller/v5/resource/containerurl/resource.go
```diff
@@ -0,0 +1,78 @@
...
+	p := azblob.NewPipeline(sc, azblob.PipelineOptions{})
+	u, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", storageAccountName))
+	serviceURL := azblob.NewServiceURL(*u, p)
+	containerURL := serviceURL.NewContainerURL(containerName)
+	_, err = containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	cc.ContainerURL = &containerURL
```
kopiczko:
```suggestion
	cc.ContainerURL = &containerURL

	r.logger.LogCtx(ctx, "level", "debug", "message", "set containerurl to context")
```

service/controller/v5/resource/containerurl/create.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if len(*(keys.Keys)) == 0 {
+		return microerror.Maskf(executionFailedError, "storage account key's list is empty")
+	}
+	primaryKey := *(((*keys.Keys)[0]).Value)
+
```
kopiczko:
```suggestion

	r.logger.LogCtx(ctx, "level", "debug", "message", "found storage account")
```

service/controller/v5/resource_set.go
```diff
@@ -166,6 +167,19 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+	{
+		c := containerurl.Config{
+			Logger: config.Logger,
+		}
+
+		containerURLResource, err = containerurl.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
```
xh3b4sd:
```suggestion
```

PR #456, opened 2019-02-08
migration: fix setting default credential secret

kopiczko:
Fixes https://github.com/giantswarm/giantswarm/issues/5251

Default credential secret should be set when the string is `== ""` and not `!= ""`.


kopiczko:
Yes I'll check if we still need it. It fixes the PM. Tests on gollum with @JosephSalisbury <3

Merging.

PR #457, opened 2019-02-08
Switch to ignition

corest:

Also includes encryption, but probably it makes sense move encryption part into separate PR


service/controller/v5/cloudconfig/cloud_config.go
```diff
@@ -207,19 +246,48 @@ func (c CloudConfig) NewWorkerCloudConfig(customObject providerv1alpha1.AzureCon
...
 
 	return newCloudConfig(k8scloudconfig.WorkerTemplate, params)
 }
 
+// certFiles returns list of certificates paths for particular role.
+// It is used for rendering cert-decrypter systemd unit.
+func (c CloudConfig) certFiles(customObject providerv1alpha1.AzureConfig, role string) (certs.Files, error) {
+	var certFiles certs.Files
+
+	clusterCerts, err := c.certsSearcher.SearchCluster(key.ClusterID(customObject))
```
xh3b4sd: The `cloudconfig` service should not rely on external dependencies IMO. It should be configured with data and render templates. Making the templating communicate over network smells and usually implies that network calls are done multiple times where they should only be done once. I suggest to provide the certs via method call arguments like we do in AWS and KVM.

service/controller/v5/cloudconfig/cloud_config.go
```diff
@@ -63,6 +74,11 @@ func New(config Config) (*CloudConfig, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.AzureConfig.%s", config, err)
 	}
 
+	encrypter, err := NewEncrypter()
```
xh3b4sd: Same as with the certs searcher. This dependency should be configured at runtime and not magically be used internally. Think about unit tests where you want to test different aspects of the template rendering. You may want to verify rendering works as expected without having to deal with encrypted content. Further the encryption protocol is now hard coded so there is no chance to get around it during testing. You may want to have a test encrypter which does not encrypt at all so you can verify the rendered templates. Then one aspect of unit testing further could be to also verify that the encryption with different specifications and requirements works as expected.

service/controller/v5/resource/instance/deployment.go
```diff
@@ -2,17 +2,49 @@ package instance
...
+
+	for _, key := range cloudConfigURLs {
+		blobExists, err := blobclient.BlobExists(ctx, key, cc.ContainerURL)
+		if err != nil {
+			return azureresource.Deployment{}, microerror.Mask(err)
+		}
+		if !blobExists {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("cloudconfig blob %#q not found", key))
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```
xh3b4sd: IMO second level methods used in the resource interface implementation should not know about control flow primitives of the resource implementation specifics. Here we should rather return an error, which is asserted by the caller. The caller here is either `EnsureCreated` or `EnsureDeleted` or the like. There we can do the resource cancelation. Look at it again from the isolated test perspective. When you want to ensure `newDeployment` behaves as expected, it is way easier to verify the returned error, as compared to no result at all and having to verify the control flow primitives. Further the control flow primitives should stay as close as possible to the resource's interface specifications as mentioned above.

service/controller/v5/resource/instance/deployment.go
```diff
@@ -35,34 +67,70 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
...
+	c := SmallCloudconfigConfig{
+		BlobURL:       masterBlobURL,
+		EncryptionKey: encryptionKey,
+		InitialVector: initialVector,
+	}
+	masterCloudConfig, err := templates.Render(key.CloudConfigSmallTemplates(), c)
+	if err != nil {
+		return azureresource.Deployment{}, microerror.Mask(err)
+	}
+	encodedMasterCloudConfig := base64.StdEncoding.EncodeToString([]byte(masterCloudConfig))
```
xh3b4sd: Why do we have to encode the rendered template all the time? What about letting the rendering do this? Are there exceptions where we do not want the encoding to happen?

corest:
I'm gonna move encryption part into separate PR. That should be easier to debug and follow

service/controller/v5/resource/instance/deployment.go
corest: There are different encodings: files encoding for ignition and final config ignition for VMSS user-data

service/controller/v5/resource/instance/deployment.go
kopiczko: I 100% agree with that.

kopiczko: That would be a good input to sig-operator repo.

service/controller/v5/templates/cloudconfig/small.go
```diff
@@ -0,0 +1,9 @@
+package cloudconfig
```
kopiczko: I'm confused. Is it cloudconfig or ignition?

service/controller/v5/cloudconfig/cloud_config.go
corest: I'll address this in separate PR

service/controller/v5/blobclient/blob_client.go
```diff
@@ -76,6 +80,33 @@ func GetBlockBlob(ctx context.Context, blobName string, containerURL *azblob.Con
...
+		// give readonly access
+		Permissions: azblob.BlobSASPermissions{Add: false, Read: true, Write: false}.String(),
+		Protocol:    azblob.SASProtocolHTTPS,
+	}.NewSASQueryParameters(sharedKeyCredential)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	blobURL := fmt.Sprintf("%s/%s/%s?%s", fmt.Sprintf(blobFormatString, storageAccountName),
+		containerName, blobName, sasQueryParams.Encode())
```
kopiczko: Nit: I'd prefer this to be:

```go
	blobURL := fmt.Sprintf("https://%s.blob.core.windows.net/%s/%s?%s", storageAccountName, containerName, blobName, sasQueryParams.Encode())
```

service/controller/v5/cloudconfig/cloud_config.go
```diff
@@ -115,15 +123,14 @@ func (c CloudConfig) NewMasterCloudConfig(customObject providerv1alpha1.AzureCon
 	customObject.Spec.Cluster.Calico.Subnet = c.azureNetwork.Calico.IP.String()
 	customObject.Spec.Cluster.Calico.CIDR, _ = c.azureNetwork.Calico.Mask.Size()
 
-	params := k8scloudconfig.Params{
```
kopiczko: Why the initialization changed here?

We usually do something like below. Initializing fields in between struct fields settings make things hard to read IMO.

```go
	ignitionPath := k8scloudconfig.GetIgnitionPath(c.ignitionPath)
	files, err := k8scloudconfig.RenderFiles(ignitionPath, params)
	if err != nil {
	}

	params = k8scloudconfig.Params{
		...
	}

```

service/controller/v5/cloudconfig/cloud_config.go
```diff
@@ -242,43 +266,5 @@ func newCloudConfig(template string, params k8scloudconfig.Params) (string, erro
...
-	if err != nil {
-		return "", microerror.Mask(err)
-	}
-	err = w.Close()
-	if err != nil {
-		return "", microerror.Mask(err)
-	}
-
-	return base64.StdEncoding.EncodeToString(buf.Bytes()), nil
+	return cloudConfig.String(), nil
```
kopiczko: This name is still confusing.

service/controller/v5/key/key.go
```diff
@@ -91,6 +94,12 @@ func BlobName(customObject providerv1alpha1.AzureConfig, role string) string {
 	return fmt.Sprintf("%s-%s-%s", VersionBundleVersion(customObject), cloudConfigVersion, role)
 }
 
+func CloudConfigSmallTemplates() []string {
```
kopiczko: Should be `IgnitionSmallTemplate`? Does it make sense to return a slice here?

service/controller/v5/resource/instance/deployment.go
```diff
@@ -2,17 +2,43 @@ package instance
...
+	}
+
+	cloudConfigURLs := []string{
+		key.BlobName(obj, key.PrefixMaster()),
+		key.BlobName(obj, key.PrefixWorker()),
+	}
+
+	for _, key := range cloudConfigURLs {
+		blobURL := cc.ContainerURL.NewBlockBlobURL(key)
+		_, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{})
```
kopiczko: Why is this line required?

service/controller/v5/resource/instance/deployment.go
```diff
@@ -35,34 +61,64 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
...
+	keys, err := storageAccountsClient.ListKeys(ctx, groupName, storageAccountName)
 	if err != nil {
 		return azureresource.Deployment{}, microerror.Mask(err)
 	}
 
-	masterCloudConfig, err := cc.CloudConfig.NewMasterCloudConfig(obj)
+	if len(*(keys.Keys)) == 0 {
+		return azureresource.Deployment{}, microerror.Maskf(executionFailedError, "storage account key's list is empty")
+	}
+	primaryKey := *(((*keys.Keys)[0]).Value)
```
kopiczko: This is to fill my knowledge gaps. What are those keys? Why the first one is primary? What are other keys? Does it make sense to create `storageAccountsClient.GetPrimaryKey` method?

service/controller/v5/cloudconfig/cloud_config.go
corest: I think it changes because of this https://github.com/giantswarm/k8scloudconfig/blob/master/v_4_1_0/cloudconfig.go#L32-L41 There are some default parameters already, which we need on the files rendering stage and not on the cloudconfig init  https://github.com/giantswarm/k8scloudconfig/blob/master/v_4_1_0/cloudconfig.go#L49

service/controller/v5/cloudconfig/cloud_config.go
corest: It is the same across all the operators as the package name is still `k8scloudconfig`. Maybe we can change it after as kvm/aws are still using `cloudconfig`. Don't want to create this inconsistency right now.
 I can track that within issue about moving ignition into separate repo https://github.com/giantswarm/giantswarm/issues/5083

service/controller/v5/resource/instance/deployment.go
corest: If blob is not ready - we return error that instance resource creation should be canceled?

service/controller/v5/key/key.go
corest: Same above - we can track that after move ignition to separate repo

service/controller/v5/resource/instance/deployment.go
corest: Keys are api keys for accessing storage account.
I'd love to skip extending basic azure client. I don't see we're doing that anywhere. I can move it to separate function, but that's not used that often

service/controller/v5/cloudconfig/cloud_config.go
kopiczko: Why do we have those `DefaultParams`? Why aren't they set in `NewCloudConfig`? That should be the place for that.

service/controller/v5/cloudconfig/cloud_config.go
kopiczko: 👍

service/controller/v5/resource/instance/deployment.go
kopiczko: Ack.

kopiczko: Would be good to add a comment.

service/controller/v5/resource/instance/deployment.go
kopiczko: How about replacing `ListKeys` with `GetPrimaryKey` instead of extending the object?
Will we use `ListKeys` anywhere else?

kopiczko: Scratch that. I just realized this is from the azure library.

kopiczko: Can we encapsulate getting the primary key in `GetBlobURL` functionality? I think we could just replace `groupName` with `groupName` and do everything there.

service/controller/v5/cloudconfig/cloud_config.go
corest: They are not there because NewCloudConfig is executed after rendering files. So if there are templates, which needs some of those parameters - those templates are generated with empty values (e.g. empty etcd port)

service/controller/v5/resource/instance/deployment.go
corest: We can, but then we will call keys listing two times

corest: Also I don't passing storage account client into blob_client or retrieving storage account again in blob_client

service/controller/v5/cloudconfig/cloud_config.go
kopiczko: Ack. Then we should do defaulting after setting params. Also if we do rendering before calling NewCloudConfig why does it even need to know about those params? But let's not change this in this PR.

corest: How can we do defaulting after settings params if setting params overrides defaulting?

kopiczko: You don't set the default if the value is already set. E.g. if EtcdPort is set to 1234 you don't change it. If it is 0 then you set it to 443.

service/controller/v5/blobclient/blob_client.go
```diff
@@ -76,6 +80,32 @@ func GetBlockBlob(ctx context.Context, blobName string, containerURL *azblob.Con
...
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	// Set the desired SAS signature values and sign them
+	//with the shared key credentials to get the SAS query parameters.
+	sasQueryParams, err := azblob.BlobSASSignatureValues{
+		BlobName:      blobName,
+		ContainerName: containerName,
+		ExpiryTime:    time.Now().UTC().Add(blobSASValidityTime * time.Hour),
```
kopiczko: Does it mean that if the instance is restarted after an hour it will fail to boot?

service/controller/v5/cloudconfig/cloud_config.go
corest: Ok, I'll add the issue for this

service/controller/v5/blobclient/blob_client.go
corest: Nope. It is 180days as `blobSASValidityTime` is 4320 https://github.com/giantswarm/azure-operator/blob/dd4ef8de6d268041803dcc81aa3250b46f0881a3/service/controller/v5/blobclient/blob_client.go#L16

service/controller/v5/resource/instance/deployment.go
```diff
@@ -35,34 +62,64 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
...
+
+	masterBlobURL, err := blobclient.GetBlobURL(ctx, masterBlobName, containerName, storageAccountName, primaryKey, cc.ContainerURL)
```
kopiczko: Is it secure to expose this URL in deployment template? Won't it make it available in the whole subscription? Or only this instance is authorized to access the data with this URL?

corest: We are not exposing it. We render small cloudconfig and put it into secure data field of the instance. Small clousconfig has this secure url for main blob (https://github.com/giantswarm/azure-operator/pull/457/files#diff-a07238412aec949b2b2c30e032093adfR92) , which appended to small ignition during boot

calvix:
any idea when this can be merged?

corest:
Currently testing this. There is one more to be done - encryption. So hope for end of a next week

calvix:
Ok, as this is kinda blocking releasing k8 1.13 on azure,  but no big deal.

corest:
Tested manually, works fine. Next step is encryption

corest:
Scaling doesn't work. No ideas for now

corest:
Tested scaling manually again. It just took ~30min. Merging

PR #458, opened 2019-02-11
Update v5 with cloudinit 3.8.0

ferrandinand:
Current WIP version of the cloudinit is `3.8.0`


corest:
This should be abandoned. Resource reservation is in place. We just missed it.

PR #461, opened 2019-02-12
Switch to coreos 1967.5.0 in v4patch2 and v5

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5334


corest:
Tested manually create/upgrade from previous version. Upgrade broken in general as it replaces all the machine at the same time

PR #465, opened 2019-02-21
Vendor audit policy

mazzy89:
Update the vendor for the recent change in the Audit policy. @corest should we consider also to add other changes in the `version_bundle.go`?


service/controller/v5/version_bundle.go
```diff
@@ -52,6 +52,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Improved node draining during updates and scaling.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "cloudconfig",
+				Description: "Change Audit policy to reduce the amount of Audit logs",
```
teemow:
```suggestion
				Description: "Improved Audit policy to reduce the amount of Audit logs (high-volume and low-risk).",
```

service/controller/v5/version_bundle.go
```diff
@@ -52,6 +52,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Improved node draining during updates and scaling.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "cloudconfig",
```
teemow:
```suggestion
				Component:   "kubernetes",
```

PR #466, opened 2019-02-21
Vendor update statusresource

tuommaki:
Statusresource got reverted back to older version at some point so restoring it
back to latest. This contains a fix for status deadlock problem.


tuommaki:
These fixes were already incorporated there earlier and quick smoke test on `talos` was successful. Going with this.

PR #467, opened 2019-02-21
update calico and k8s

calvix:
update calico to 3.5.1
update k8s to 1.13 
update etcd to 3.3.12


calvix:
> Just to understand here. I miss the general context. Why the update of those components requires an update of RBAC?

New calico does new things > it needs new RBAC

service/controller/v5/cloudconfig/template.go
```diff
@@ -576,6 +597,7 @@ subjects:
 - kind: ServiceAccount
   name: calico-node
   namespace: kube-system
+---
```
MarcelMue: Don't think this is needed, but not a big deal.

service/controller/v5/cloudconfig/template.go
```diff
@@ -556,15 +555,37 @@ rules:
...
+  - apiGroups: [""]
+    resources:
+      - nodes
+    verbs:
+      - get
+      - list
       - watch
-
+  # These permissions are only requried for upgrade from v2.6, and can
+  # be removed after upgrade or on fresh installations.
```
MarcelMue: Is this tracked in an issue? So we remember to remove them again?

calvix:
Tested on godmsack,  both:
* new k8s 1.13 cluster
* upgrade from 1.11 -> 1.13  - fails,  after upgrade workers cannot access API -> timeout, tried to investigate but found no issues

All good

calvix:
merging as i crated around 10+ clusters and all were fine ( both fro new version and updates)

PR #471, opened 2019-02-26
update vendor deps

xh3b4sd:


integration/key/release_name.go
```diff
@@ -0,0 +1,29 @@
+package key
+
+import (
+	"fmt"
+)
+
+func AWSConfigReleaseName(clusterID string) string {
```
rossf7: Shouldn't this be for Azure?

integration/key/release_name.go
```diff
@@ -0,0 +1,29 @@
...
+	return fmt.Sprintf("e2esetup-awsconfig-%s", clusterID)
```
MarcelMue: Seems like a mistake

integration/key/release_name.go
xh3b4sd: I copied the whole file from the `aws-operator`. Dropped the dead code.

integration/key/release_name.go
xh3b4sd: I copied the whole file from the `aws-operator`. Dropped the dead code.

PR #474, opened 2019-03-01
fix e2e tests

xh3b4sd:


integration/setup/setup.go
```diff
@@ -1,5 +1,3 @@
-// +build k8srequired
```
TheoBrigitte: Can you explain why is this build flag removed ?

xh3b4sd: The build tags are needed for the e2e code. The actual tests define them. We do not have them everywhere. In my special case it was that some editor integrations are not working with these tags. It was also that within the `setup` package not all files had that. Now it is consistent and no files have that within this package.

PR #479, opened 2019-03-05
Update CoreOS

mazzy89:


service/controller/v6/version_bundle.go
```diff
@@ -7,12 +7,16 @@ import (
...
 		Changelogs: []versionbundle.Changelog{
-
 			{
 				Component:   "azure-operator",
 				Description: "Fix ingress load-balancer creation.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "containerlinux",
+				Description: "Update CoreOS to 2023.4.0.",
```
teemow:
```suggestion
				Description: "Update CoreOS to 2023.4.0. Fixes CVE-2019-8912",
```

mazzy89:
e2e tests fail all and after a quick investigation I've found this issue:

```
{"caller":"github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/controller/controller.go:370","controller":"azure-operator","event":"update","level":"error","loop":"6","message":"stop reconciliation due to error","object":"/apis/provider.giantswarm.io/v1alpha1/namespaces/default/azureconfigs/ci-cur-b9de7-c1c86","resource":"containerurlv5","stack":"[{/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/controller/controller.go:530: } {/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/metricsresource/resource_wrapper.go:39: } {/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/resource_wrapper.go:79: } {/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/resource_wrapper.go:67: } {/go/src/github.com/giantswarm/azure-operator/service/controller/v5/resource/containerurl/create.go:34: } {storage.AccountsClient#GetProperties: Invalid input: autorest/validation: validation failed: parameter=accountName constraint=MaxLength value=\"gsstorageaccountcicurb9de7c1c86\" details: value length must be less than or equal to 24}]","time":"2019-03-05T15:08:49.011504+00:00","version":"602"}
```

It might be related to the storage.

mazzy89:
@giantswarm/sre the e2e fails because the storage account name is longer than the max length which is supposed to be 24 chars. Should we change the way to generate storage account name in e2e tests? why does it fail now?

mazzy89:
No changes in the schema from MS side. The max is 24 as reported in e2e errors https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#storage

corest:
Did we change e2e test cluster name? I’m aware of that 24 symbols requirement and that wasn’t a blocker earlier

calvix:
Yeah we hit this in terraform before and have to shorten the name, so question is what changed

mazzy89:
There are a series of fixes in the e2e tests https://github.com/giantswarm/azure-operator/commit/60cfc12b862af8d1655db5fd8dc8c2ba2c4be70d

@xh3b4sd can you recall any changes in this regard?

xh3b4sd:
I checked the code history and the ID generation changed last back in October or something like that.

mazzy89:
Ok I'll fix here so we can move with this PR. @xh3b4sd could you please point out where is the ID generation is? I was not able to find it

xh3b4sd:
See https://github.com/giantswarm/azure-operator/blob/master/integration/env/common.go#L106-L125.

mazzy89:
@giantswarm/sig-operator @giantswarm/sig-releng I was thinking to shorten the Cluster Id in this way:

`ci-wip-3cc75-5e958` → `ci-wip-3cc75`

Basically removing the hash of the integration test dir. any objections?

xh3b4sd:
Yeah, because that does not work. There is this differentiator so we can run multiple tests in parallel.

corest:
It is easier to drop giantswarm to gs in storage account name template

mazzy89:
@corest that won't work. the name is `gsstorageaccountcicurb9de7c1c86`. dropping `gs` we won't get into the 24 lengths necessary to pass the validation.

mazzy89:
Let's drop `storageaccount` so it will become `gscicurb9de7c1c86`

corest:
Name should be meaningful and consistent across code and tests. Maybe gssa at least

mazzy89:
The summary of our work:

> There are only two hard things in Computer Science: cache invalidation and naming things.
Phil Karlton

😂

Agree with you @corest

kopiczko:
> Name should be meaningful and consistent across code and tests. Maybe gssa at least

Agreed but I don't see the point of embedding storageaccount in storage account URL. Same applies to e.g. the resource groups. We don't call them `eggs2-resource-group` but `eggs2`.

PR #492, opened 2019-03-26
Vendor update certs dependency

tuommaki:
In order to import fix for corner case error handling in certs searching, but
avoid breaking change from upstream opencensus dependency, only vendor certs.


tuommaki:
Stale

PR #493, opened 2019-03-26
Use k8scloudconfig v4.1.2 for WIP

tuommaki:
In latest patch version of k8scloudconfig calico-kube-controllers POD has node
selector for master so that it doesn't get rescheduled so often during upgrade.

Towards https://github.com/giantswarm/giantswarm/issues/5584


kopiczko:
How about adding changelog entry?

tuommaki:
> How about adding changelog entry?

Thanks! Added.

PR #494, opened 2019-03-26
fix-missing-ingress-loadbalancer

calvix:
during https://github.com/giantswarm/giantswarm/issues/3585 testing i found that we actually dont apply loadbalancer manifest to the cluster so no ingress access is possible that seems to be wrong

fixing for WIP and also  backporting to current active release


calvix:
Well it was there probably on older versions so old cluster that were upgraded are fine, but new ones are broken

PR #502, opened 2019-04-12
fix-mount-units

calvix:
fix mount units with new ignition


service/controller/v7/templates/ignition/docker_mount_unit.go
```diff
@@ -2,8 +2,6 @@ package ignition
 
 const DockerMountUnit = `[Unit]
 Description=Mounts disk to /var/lib/docker
-Requires=format-docker-disk.service
-After=format-docker-disk.service
```
xh3b4sd: Do we remove these because they are already gone?

calvix: Yes, this units are no longer available as we do format by ignition

PR #517, opened 2019-05-31
Add azure file storage classes

pipo02mix:
Add Azure File Storage Classes by default in our clusters.

### DIFF
--- a/service/controller/v8/templates/ignition/default_storage_class.go
+++ b/service/controller/v8/templates/ignition/default_storage_class.go
@@ -36,4 +36,58 @@ provisioner: kubernetes.io/azure-disk
 parameters:
   kind: Managed
   storageaccounttype: Standard_LRS
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:azure-files-cloud-provider
+rules:
+- apiGroups: ['']
+  resources: ['secrets']
+  verbs:     ['get','create']
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:azure-files-cloud-provider
+roleRef:
+  kind: ClusterRole
+  apiGroup: rbac.authorization.k8s.io
+  name: system:azure-files-cloud-provider
+subjects:
+- kind: ServiceAccount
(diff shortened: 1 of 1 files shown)


pipo02mix:
Test on godsmack. The cluster created successfully and storage classes there and they worked.

```
~/go/src/github.com/giantswarm/azure-operator(storage_class) » k get sc                                                                                                          
NAME                PROVISIONER                AGE
af-premium          kubernetes.io/azure-file   11m
af-standard         kubernetes.io/azure-file   11m
default (default)   kubernetes.io/azure-disk   11m
managed-premium     kubernetes.io/azure-disk   11m
```

```
~/Code/debug/pvc » k get pvc                                                                                                                                                     (giantswarm-3krj8/default)
NAME        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
azurefile   Bound    pvc-dd955f7d-8392-11e9-b5ea-000d3a39dbab   5Gi        RWX            af-standard    2m38s
```

xh3b4sd:
What happens when you delete the cluster?

pipo02mix:
> What happens when you delete the cluster?

The cluster was deleted successfully. Why adding a storage class has an impact in the cluster deletion?

xh3b4sd:
I don't know, which is why I ask. Depending on who manages the classes effectively somebody has to clean them up. When ARM is doing it then cool. When this is something we do somehow, then we have to do it. My thinking was just to make sure the lifecycle is properly taken care of.

pipo02mix:
Ah so you mean, no k8s resource but the volumes provisioned by this storage classes?

xh3b4sd:
I mean anything related to that cluster. When there is a new definition of whatever, we need to make sure that all the implications of that definition is properly managed during creation, update and deletion.

pipo02mix:
then we are covered there is no problem to add these new storage classes (tested thoroughly). About the provisioned volumes are something can take responsibility the customer

rossf7:
Adding my 2 cents. We definitely had this problem on AWS and we added the `ebsvolume` resource that goes "garbage collection". As leftover volumes were blocking the cluster deletion.

But Azure has the resource group concept So since we delete the resource group for the cluster we should be fine here.

PR #518, opened 2019-06-11
Vendor update latest k8scc

MarcelMue:


corest:
Vendoring latest core-components updated with labels, required for network-policies

PR #519, opened 2019-06-11
Change azure cloud config ownership

corest:
Change azure config group ownership so that external-dns, which is running as `nobody` user/group can access that file from container.


service/controller/v8/cloudconfig/master_template.go
```diff
@@ -142,35 +142,51 @@ func (me *masterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
...
 			Path:         "/srv/calico-azure.yaml",
 			Owner: k8scloudconfig.Owner{
-				User:  FileOwnerUser,
-				Group: FileOwnerGroup,
+				User: k8scloudconfig.User{
+					Name: FileOwnerUserName,
+				},
+				Group: k8scloudconfig.Group{
+					Name: FileOwnerGroupName,
+				},
```
rossf7: I'd prefer this to be ordered.

PR #521, opened 2019-06-21
Ensure pod runs as non-root user

glitchcrab:
Towards giantswarm/giantswarm/issues/6013


glitchcrab:
> I don’t see Dockerfile updated with image 3.9-giantswarm, which has giantswarm user

I'm holding off on the images at the moment until this is finalised: giantswarm/giantswarm/issues/6290

corest:
But then this PR will not work. There is no user giantswarm with is 1000 in current image.

glitchcrab:
I've experimented with that - it just runs as UID 1000 - the user doesn't actually need to exist in the image:

```
PID   USER     TIME  COMMAND
    1 1000      0:00 sleep 10h
```

corest:
What happens when you exec into container? Which use is used?

glitchcrab:
You still get uid 1000:

```
/ $ whoami
whoami: unknown uid 1000
```

There's a non-root pod in the default namespace on gauss at the mo if you want to take a look.

glitchcrab:
Just to follow this up, aws-operator is running as non-root without adding a user to the image: https://github.com/giantswarm/aws-operator/commit/a86f4c5ea56a3b633c890a8787a02f4e0a3300a4

glitchcrab:
Operator tested and working fine on godsmack, no errors in the logs.

PR #522, opened 2019-06-26
k8s 1.14.3

mazzy89:


mazzy89:
tested and works

service/controller/v8/version_bundle.go
```diff
@@ -8,31 +8,46 @@ func VersionBundle() versionbundle.Bundle {
...
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "calico",
+				Description: "Update calico to 3.7.2. More info here: https://docs.projectcalico.org/v3.7/release-notes/",
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "containerlinux",
+				Description: "Updated to 2079.5.1. More info here: https://github.com/coreos/manifest/releases/tag/v2079.5.1",
```
xh3b4sd: The speech is different. Above is `Update` and here is `Updated`. @rossf7 since we had this yesterday already, how should it be now? Apologies for asking again. 🙈

mazzy89: Copied pasted from also the other operators. if it must be changed then I'll go to change everwhere.

MarcelMue: Autopiloted through this to fast - well spotted.
It should be past tense AFAIK, so `Updated`.

mazzy89: Going to change for that.

rossf7: It should be imperative so `Update`. `fmt` is outdated doing a PR now.

mazzy89: ouch 👍

rossf7: See https://github.com/giantswarm/fmt/pull/55.

mazzy89: Thank you @rossf7. I've changed it ;-)

mazzy89:
Merging here after updating vendor

PR #526, opened 2019-06-28
update vendor deps

xh3b4sd:
Follow up of https://github.com/giantswarm/operatorkit/pull/300.


integration/setup/config.go
```diff
@@ -87,8 +87,7 @@ func NewConfig() (Config, error) {
 		c := framework.HostConfig{
 			Logger: logger,
 
-			ClusterID:  env.ClusterID(),
-			VaultToken: env.VaultToken(),
+			ClusterID: env.ClusterID(),
```
xh3b4sd: The Vault token got dropped because the legacy framework does not do anything with it.

integration/setup/teardown.go
```diff
@@ -2,58 +2,22 @@ package setup
...
-		c.Host.DeleteGuestCluster(ctx, "azure")
-
-		// only do full teardown when not on CI
-		if env.CircleCI() == "true" {
-			return nil
-		}
-	}
-
-	{
-		err = framework.HelmCmd(fmt.Sprintf("delete %s-azure-operator --purge", c.Host.TargetNamespace()))
```
xh3b4sd: This is the same story as in the `aws-operator`, just more sad.

PR #531, opened 2019-07-03
add usage client to clientset

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/5826. I want to try to add a new collector for the quotas and see if the usage client is what we need.


client/azure_client_set.go
```diff
@@ -14,44 +12,6 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
-type AzureClientSetConfig struct {
```
xh3b4sd: I moved this to its own file.

client/azure_client_set.go
```diff
@@ -66,6 +26,8 @@ type AzureClientSet struct {
 	InterfacesClient *network.InterfacesClient
 	//StorageAccountsClient manages blobs in storage containers.
 	StorageAccountsClient *storage.AccountsClient
+	// UsageClient is used to work with limits and quotas.
+	UsageClient *compute.UsageClient
```
xh3b4sd: Added the usage client here.

client/azure_client_set.go
```diff
@@ -82,20 +44,21 @@ type AzureClientSet struct {
 
 // NewAzureClientSet returns the Azure API clients.
 func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
-	if err := config.Validate(); err != nil {
```
xh3b4sd: Bad kitty.

client/azure_client_set.go
```diff
@@ -82,20 +44,21 @@ type AzureClientSet struct {
...
-		return nil, microerror.Maskf(invalidConfigError, "config.%s", err)
+	err := config.Validate()
```
xh3b4sd: Good kitty.

client/azure_client_set.go
```diff
@@ -82,20 +44,21 @@ type AzureClientSet struct {
...
 	// cloud. For empty config.EnvironmentName returns Azure public cloud.
 	env, err := parseAzureEnvironment(config.EnvironmentName)
 	if err != nil {
-		return nil, microerror.Maskf(err, "parsing Azure environment")
+		return nil, microerror.Mask(err)
 	}
 
 	servicePrincipalToken, err := newServicePrincipalToken(config, env)
 	if err != nil {
-		return nil, microerror.Maskf(err, "creating service principal token")
```
xh3b4sd: I find these super useless to be honest. It does not add value during debugging.

client/fakeclient/azure.go
```diff
@@ -1,15 +0,0 @@
-package fakeclient
-
-import (
-	"github.com/giantswarm/azure-operator/client"
-)
-
-// TODO(PK) remove it as soon as we sort AzureClient and calico CC extention.
-func NewAzureConfig() client.AzureClientSetConfig {
```
xh3b4sd: I dropped this as it doesn't appear to be used anywhere anymore.

client/azure_client_set.go
tuommaki: As long as the original error has enough context for debugging, I agree.

xh3b4sd: The stack is the context. At least this is for me the most relevant thing. The error message almost never tells me anything useful. I look at the stack. I look at the code. I fix the problem.

PR #532, opened 2019-07-03
fix e2e test

xh3b4sd:


integration/test/clusterstate/main_test.go
```diff
@@ -44,9 +44,9 @@ func init() {
 
 	{
 		c := clusterstate.Config{
-			GuestFramework: config.Guest,
-			Logger:         config.Logger,
-			Provider:       p,
+			LegacyFramework: config.Guest,
```
xh3b4sd: That changed in the vendor deps.

PR #535, opened 2019-07-04
add usage collector

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/5826. Idea here is to have metrics for currently used resources and their limits. Then we can have alerts for certain thresholds.

### DIFF
--- a/service/collector/usage.go
+++ b/service/collector/usage.go
@@ -110,7 +110,7 @@ func (u *Usage) Collect(ch chan<- prometheus.Metric) error {
 	}
 
 	// We generate clients and group them by subscription.
-	var clients map[string]*compute.UsageClient
+	clients := map[string]*compute.UsageClient{}
 	{
 		for _, cr := range crs {
 			c, err := u.getUsageClient(cr)
--- a/service/collector/usage.go
+++ b/service/collector/usage.go
@@ -123,11 +123,10 @@ func (u *Usage) Collect(ch chan<- prometheus.Metric) error {
 		for r.NotDone() {
 			for _, v := range r.Values() {
 				fmt.Printf("\n")
-				fmt.Printf("%#v\n", u.location)
 				fmt.Printf("    %#v\n", *v.CurrentValue)
 				fmt.Printf("    %#v\n", *v.Limit)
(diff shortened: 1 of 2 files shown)


service/collector/usage.go
```diff
@@ -0,0 +1,178 @@
...
+			return microerror.Mask(err)
+		}
+
+		for r.NotDone() {
+			for _, v := range r.Values() {
+				ch <- prometheus.MustNewConstMetric(
+					usageCurrentDesc,
+					prometheus.GaugeValue,
+					float64(*v.CurrentValue),
+					*v.Name.LocalizedValue,
```
MarcelMue: What does LocalizedValue mean here? Does this mean that if the subscription is german we get the german name here?

xh3b4sd: There are two "values". One is camel cased and has sometimes insufficient information as I found out. The localized thingy is what you see in Prometheus. It is more human readable. I thought this is the better option. 

Screenshot 2019-07-04 at 14 08 45
Screenshot 2019-07-04 at 14 08 26

MarcelMue: I see, makes sense.

I am a little concerned about adding 53 new series to prometheus and what impact on memory usage it will have. But we can try and see.

xh3b4sd: It is 106 per subscription, but why should that matter? That is really not that much data IMO.

MarcelMue: I remember that we limitied the amount of series we get from tenants multiple times to improve performance of prometheus. I do not have a reference point as to what is `a lot` of data unfortunately - hence my concern.

You are right that it should be fine but I am lacking a sense of scale here.

kopiczko: I also would prefer to have only metrics that we care for. But there are places where we can probably shave way more.

As for the localized name won't it be awkward to use that for creating alerting rules? I wouldn't be surprised if those names change over time.

xh3b4sd: What I found weird with the other names is that the CPU information or VM family wasn't always super clear. So far I did not write an alert but when we do it is rather easy to improve now after most of the hustle got sorted out.

kopiczko: Let's iterate.

PR #536, opened 2019-07-05
Switch wip to 4 5 0

corest:
WIP versions need namespace labeling to work with network policies


service/controller/v8/version_bundle.go
```diff
@@ -27,6 +27,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Update to 3.3.13. More info here: https://github.com/etcd-io/etcd/blob/master/CHANGELOG-3.3.md#v3313-2019-05-02",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "ignition",
+				Description: "Add name label for default and kube-system namespaces.",
+				Kind:        versionbundle.KindAdded,
+			},
```
kopiczko: Isn't v8 already release?

corest: `v8` is `2.4.0` and it is WIP

kopiczko: So why do we have v9?

corest: That's a question to @mazzy89 https://github.com/giantswarm/azure-operator/pull/525

PR #541, opened 2019-07-15
vendor: updated k8scloudconfig

rossf7:
Towards giantswarm/giantswarm#6367

Updated k8scloudconfig with kubelet image pull progress deadline. We continue to use the default 1 minute.


rossf7:
Tested on godsmack.

PR #542, opened 2019-07-17
Increase eth0 tx queue size

corest:
Fixing network performance issues


corest:
Yep, tested. 
I have no strong reasoning over queue size. But just aligning with AWS. Max tx queue size is twice bigger

PR #543, opened 2019-07-19
use standart LB for azure resources

calvix:
migrate LB type from 'Basic' to 'Standart' as Basic is more entry level LB and does not have any SLA

This is nontrivial migration. These changes are aplied:
* change public IP SKU to standard
* change LB SKU to standard
* change API version for public IP and LB
* add new systemd unit to enable NAT for Pod traffic (required with new LB) - https://github.com/Azure/azure-container-networking/blob/master/docs/cni.md#outbound-connectivity-from-pods
* as new LB only allow traffic outgoing traffic type that is used on LB we needed to add UDP loadblancing rule  via kubernetes option `service.beta.kubernetes.io/azure-load-balancer-mixed-protocols: "true" ` and add iptables rules to block this traffic, otherwise UDP traffic outside wont work


service/controller/v9/resource/vpn/template/main.json
```diff
@@ -61,7 +61,7 @@
           {
             "id":"string",
             "properties":{
-              "privateIPAllocationMethod":"Dynamic",
+              "privateIPAllocationMethod":"Static",
```
calvix: are you sure with this? this looks kinda weird

T-Kukawka: i am just checking out if that works, this is what Azure is saying why it failed 😢  but i think i have but it in wrong place actually so just changing it

calvix:
@T-Kukawka SO should we merge this PR?

service/controller/v9/templates/ignition/azure_cni_nat_rules.go
```diff
@@ -0,0 +1,15 @@
...
+Description=Setup iptables Nat rules for Azure CNI
+Wants=systemd-networkd.service
+After=systemd-networkd.service
+Before=docker.service
+[Service]
+Type=oneshot
+ExecStartPre=/bin/sh -c "iptables -I INPUT 1 -m udp -p udp --source-port 80 -j DROP"
+ExecStartPre=/bin/sh -c "iptables -I INPUT 1 -m udp -p udp --source-port 443 -j DROP"
+ExecStart=/bin/sh -c "iptables -t nat -A POSTROUTING -m addrtype ! --dst-type local ! -d {{.VnetCIDR}} -j MASQUERADE"
+ [Install]
```
kopiczko:
```suggestion
[Install]
```

service/controller/v9/version_bundle.go
```diff
@@ -32,6 +32,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Add VNET service endpoint for accessing Azure Storage.",
 				Kind:        versionbundle.KindAdded,
 			},
+			{
+				Component:   "kubernetes",
+				Description: "Update kubernetes to 1.14.5 (CVE-2019-11247) https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#v1145",
+				Kind:        versionbundle.KindChanged,
```
MarcelMue: This PR only added a changelog about the k8s version but not about the LB. Can someone take care or should I provide a PR?

PR #548, opened 2019-07-26
remove support for .DockerTag and releases

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6557

Remove `.DockerTag` as we only want to use either the old `.SHA` or the new `.Version`.
Going with old way here, as we are now focusing on aws-operator.


TheoBrigitte:
Tested on ghost.

TheoBrigitte:
One more approval would be nice

PR #549, opened 2019-08-02
Improve scrape of usage

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6426

Same as done here https://github.com/giantswarm/cluster-service/commit/135da30bc9f30b6dc2831f9e669620a140d09ef4 we improve the scraping so a failure of scraping external resources that might be nonresponsive does not put the service down from a Prom perspective and trigger us with annoying `TargetDown` pages

### DIFF
--- a/service/collector/usage.go
+++ b/service/collector/usage.go
@@ -140,7 +140,7 @@ func (u *Usage) Collect(ch chan<- prometheus.Metric) error {
 	for _, c := range clients {
 		r, err := c.List(context.Background(), u.location)
 		if err != nil {
-			u.logger.Log("level", "warning", "message", "An error occured during the scraping of usage", "stack", fmt.Sprintf("%v", err))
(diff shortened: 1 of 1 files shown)


service/collector/usage.go
```diff
@@ -127,28 +140,30 @@ func (u *Usage) Collect(ch chan<- prometheus.Metric) error {
...
-					prometheus.GaugeValue,
-					float64(*v.Limit),
-					*v.Name.LocalizedValue,
-				)
-			}
-
-			err := r.Next()
-			if err != nil {
-				return microerror.Mask(err)
+			u.logger.Log("level", "warning", "message", "An error occured during the scraping of usage", "stack", fmt.Sprintf("%v", err))
```
kopiczko:
```suggestion
			u.logger.Log("level", "warning", "message", "an error occurred during the scraping of usage", "stack", fmt.Sprintf("%v", err))
```

mazzy89:
Yeah @kopiczko you are right. let me think how i can name that thing better

kopiczko:
It's fine with me. :shipit:

PR #552, opened 2019-08-08
v8patch1: update imports and packages

schjan:
Merge after https://github.com/giantswarm/azure-operator/pull/551


service/controller/v8patch1/resource/vpnconnection/resource.go
```diff
@@ -4,11 +4,12 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-06-01/network"
-	"github.com/giantswarm/azure-operator/client"
-	"github.com/giantswarm/azure-operator/service/controller/setting"
-	"github.com/giantswarm/azure-operator/service/controller/v8/controllercontext"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/setting"
+	"github.com/giantswarm/azure-operator/service/controller/v8patch1/controllercontext"
```
mazzy89: Any reasons why you moved this down?

service/controller/v8patch1/resource_set.go
```diff
@@ -37,6 +17,27 @@ import (
 	"github.com/giantswarm/statusresource"
 	"github.com/giantswarm/tenantcluster"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/azure-operator/client"
```
mazzy89: Any reasons why you moved this down?

schjan:
Just automatic Go Imports going on.

First stdlib imports, then other Imports then from same module. Standard
setting of my IDE but I also think we agreed on that in giantswarm/fmt.

Salvatore Mazzarino <notifications@github.com> schrieb am Fr. 9. Aug. 2019
um 08:04:

xh3b4sd:
You could just have merged the copy PR and made the imports PR afterwards. The PR here is even bigger than the first one because it had all changes of the copy branch plus the import changes which is then in total compared to master.

mazzy89:
Yeah indeed @xh3b4sd is right. Not sure why you opened sequentially all these PRs before merging it. 🤔

schjan:
Oh yeah.. nothing a force push can't handle but certainly not the best approach :neckbeard: :rocket:

PR #555, opened 2019-08-13
Move v9 to v10

mazzy89:
Signed-off-by: Salvatore Mazzarino <apocalipse89@gmail.com>

This PR

* rename the bundle **v9** → **v10**
* update the imports
* wire v10

This because the bundle **v9** will be used to deliver the release 8.3.0 for the Standard LB


MarcelMue:
Why do we need to move this? I don't really follow. We could just use it for the LB change

mazzy89:
@MarcelMue yeah you are out for holiday. let me explain this. 

we want to deliver the LB change only and move all the other azure changes for the next release which will contain above all k8s 1.15.

so the 8.3.0 will container 1.14.5 + LB changes 

does it make sense for you?

MarcelMue:
I understand that we want to release this before 1.15 and so on, but v9 has barely any changes in it right now already and moving it will be confusing IMO.

It does not have any changes for 1.15 from what I can see.

mazzy89:
ok so what are you proposing here?

MarcelMue:
Only changes here:
https://github.com/giantswarm/azure-operator/blob/master/service/controller/v9/version_bundle.go#L10-L35
So only bugfixes and ignition.

Proposal would be to add the changes for LB change to v9 right now. No need to copy or whatever. 

Then have v10 for k8s 1.15.

MarcelMue:
Vaclavs PR is also already based and tested on v9 : https://github.com/giantswarm/azure-operator/pull/543/files

So we could just merge it.

mazzy89:
Ok i understand your point. just are we 100% sure of all the changes that are in the v9? personally i'm not sure and i don't have time to check that

MarcelMue:
I don't know if vaclavs PR even works without the other changes in v9.

mazzy89:
@MarcelMue Vaclav's change should be tested and they work. am I right? @T-Kukawka 

i went down this path because i don't feel confident to deliver with the other changes as well especially with this super narrow time window where the LB changes must go out today, tomorrow monrning maxiumum and the customer is pretty pushy.

mazzy89:
Let's close this as discussed

PR #558, opened 2019-08-13
Wire v10

mazzy89:
Signed-off-by: Salvatore Mazzarino <apocalipse89@gmail.com>

Wire v10

I put as azure-operator version 3.0.0 cause this will have k8s 1.15.

Feel free to correct it if I'm wrong

### DIFF
--- a/service/controller/v10/version_bundle.go
+++ b/service/controller/v10/version_bundle.go
@@ -36,6 +36,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "azure-operator",
-		Version: "3.0.0",
+		Version: "2.6.0",
 	}
 }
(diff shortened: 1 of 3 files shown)


MarcelMue:
Something is not right here - I can take over:
```
# github.com/giantswarm/azure-operator/service/controller
service/controller/cluster.go:216:6: v10ResourceSet declared and not used
```

mazzy89:
I'll take a look asap

service/controller/v10/version_bundle.go
```diff
@@ -61,6 +36,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "azure-operator",
-		Version: "2.5.0",
+		Version: "3.0.0",
```
MarcelMue:
```suggestion
		Version: "2.6.0",
```

Let's not bump too much here - we usually bump this high if there are many changes in the operator. This here is mainly in k8scc IMO.

mazzy89: ok it make sense

PR #559, opened 2019-08-15
integration: install credential-default Secret during setup

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/6736.

PR #560, opened 2019-08-19
Vendor update

tuommaki:
Update vendor to incorporate latest updates including operatorkit context
memory leak fix.


MarcelMue:
More build errors :/

integration/setup/provider.go
```diff
@@ -22,12 +22,6 @@ func provider(ctx context.Context, config Config) error {
 			},
 			Secret: chartvalues.AzureOperatorConfigSecret{
 				AzureOperator: chartvalues.AzureOperatorConfigSecretAzureOperator{
-					CredentialDefault: chartvalues.AzureOperatorConfigSecretAzureOperatorCredentialDefault{
```
tuommaki: Is this fine to be just deleted? That field doesn't exist in upstream library any more and tests pass with this.

MarcelMue: It got deleted but I am not sure if something needs to be added to the e2e test then. Pawel worked on that for AWS.

MarcelMue: See his changes here:
https://github.com/giantswarm/aws-operator/pull/1880/files

tuommaki: Ok. This got fixed here: https://github.com/giantswarm/azure-operator/pull/559/files

tuommaki:
CI Passes build. E2e expected to fail so far. Going with this.

PR #561, opened 2019-08-20
Update k8scc to include k8s 1.14.6

MarcelMue:
Upgrades calico and k8s 1.14.6 - I am certain that 1.14.6 is fine.
I am a bit unsure about calico - it seems fine when testing but the jump is quite big on azure which was an accident.


mazzy89:
Tested creation and upgrade in azure from 8.3.0 to 8.4.0 and it worked like a charm

mazzy89:
Going to merge this

PR #562, opened 2019-08-26
Add subscription label to usage metrics

tfussell:
For https://github.com/giantswarm/giantswarm/issues/6558, we are missing certain metrics exported by `azure-operator` on `talos` leading to spurious alerts. `talos` is a BYOC installation. Prometheus shows that the `azure-operator` metrics endpoint is returning a 500 error. 

azure-metrics-down

Investigating with @MarcelMue, we discovered the following errors returned by the operator:

```
WARNING: CONTROL-PLANE MASTER | giantswarm@master0 ~ $ curl  http://10.0.132.17:8000/metrics
An error has occurred while serving metrics:
318 error(s) occurred:
* collected metric "azure_operator_usage_current" { label:<name:"name" value:"Availability Sets" > gauge:<value:0 > } was collected before with the same name and label values
...
```

This pull request adds the `SubscriptionID` as a label called `subscription` for usage metrics to ensure that metric names/labels are always unique.

As indicated by the comments, this issue was foreseen, but I think this installation is unique because it has tenant clusters on multiple accounts so it hadn't been fully tested.


MarcelMue:
Update the branch first though :D

PR #499, opened 2019-04-02
build and release tags

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5206

This PR enable tagged version to be built and released.

What are the results in case a `v1.0.2` tag  is created:
- docker image `quay.io/giantswarm/azure-operator:1.0.2`
- helm chart `quay.io/giantswarm/azure-operator-chart@1.0.2`
- github v1.0.2 release with `azure-operator-chart-1.0.2.tgz` asset

Breaking change:
- ~chart version for untagged commit will change from `1.0.0-<sha>` to `v0.0.0-<sha>`~

Manual update:
- circleci token has to be updated with full access to private and public github repositories in order for architect to be able to publish github releases. see `repo` in "Available scopes" https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes

PR #569, opened 2019-09-24
Revert "build and release tags (#499)"

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/7044

This reverts commit 05f0061f06e069268eb640af38f2e2e018aa73ca.

This PR revert the chart version to fixed `1.0.0-<sha>`. So we do not rely on the version detected by architect which is going to change https://github.com/giantswarm/architect/pull/335


kopiczko:
Why not tagging the first commit with `v1.0.0` instead?

PR #581, opened 2019-10-15
Add rate limiting to cloud provider config

MarcelMue:
This should stop us (hopefully) from being rate limited because the cloud provider is sending too many requests.


service/controller/v11/templates/ignition/cloud_provider_conf.go
```diff
@@ -9,6 +9,7 @@ location: {{ .Location }}
 aadClientId: {{ .AADClientID }}
 aadClientSecret: {{ .AADClientSecret }}
 {{- end}}
+cloudProviderRateLimit: true
```
xh3b4sd: How should this help? Can we have a comment that explains why this is necessary and what it is doing?

fiunchinho: Don't we need to specify `cloudProviderRateLimitQPSWrite` and `cloudProviderRateLimitQPS`?

MarcelMue: @xh3b4sd This is a config file that is passed to cloud provider magic on azure nodes (the one that adds labels to nodes and so on). This simply enables the cloud provider implementation here to honor rate limits (as it is set to false by default). Not sure what I should comment here.

@fiunchinho My understanding of the docs was that they are only _valid_ if this flag is set to true, not _required_. I would expect azure to set sensible defaults here, but it wouldn't be the first time I was wrong with that assumption.

xh3b4sd: Put yourself into the shoes of some new PE or SRE who was not involved in all the hustle here before. Then he deals with all kinds of weird flags and config files. What would you like to tell them in the comment? Let's just explain the mechanism a bit and provide reasoning why we need it and what would happen when we would not have this.

kopiczko: IMO comment isn't necessary here. This is provider config. You need to google that as all other settings in the file. Putting comments on all of would obscure the config more than help imo.

JosephSalisbury:
Looks like this alone doesn't fix the issue. Updated all kubelets on `gollum` to use this configuration, and we're still facing the issue (verified by checking any tenant cluster master vmss, and seeing rate limit error)

service/controller/v11/templates/ignition/cloud_provider_conf.go
xh3b4sd: I totally disagree with this. People brag about missing docs all the time and now you refuse to add them. I leave you alone with this now. Disagree and commit.

TheoBrigitte:
We might have to try setting the rate limits

Seems like most people use those as default:
```
    cloudProviderRatelimit: true
    cloudProviderRateLimitQPS: 3
    cloudProviderRateLimitBucket: 10
```

service/controller/v11/templates/ignition/cloud_provider_conf.go
TheoBrigitte: Comment would be good, and a link to an issue with more details.

service/controller/v11/templates/ignition/cloud_provider_conf.go
```diff
@@ -9,6 +9,9 @@ location: {{ .Location }}
...
+cloudProviderRateLimitQPS: 3
```
tuommaki: Can this be fractional number? According to documentation here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-request-limits that config is reasonable if given k8s cluster is the only one using this specific service principal token, but in our case I believe we share it among all tenant clusters. Therefore the correct value would be rather something like `0.15` (=3/20).

MarcelMue: Yes it can be - I don't quite follow the logic for the `0.15` though. Can you expand?

MarcelMue: Here are some docs about the types and which options exist:
https://github.com/kubernetes/cloud-provider-azure/blob/master/docs/cloud-provider-config.md#cluster-config

tuommaki: > I don't quite follow the logic for the 0.15 though. Can you expand?

My logic there is that number `3` (QPS) probably comes from this:
> Resource Manager allows up to 12,000 read requests per hour

Which calculates 12000/3600 == 3.33333. Now that's the absolute max for one service principle if I understand the azure documentation correctly. When we have say 20 tenant clusters on an installations and they all are sharing the same principles, it leaves single k8s with quota of 3.33333/20 which nicely rounded (with some room for error) lands on `0.15`.

MarcelMue: But this cloud provider thing is running on each node so it would have to be `3.3333 / Number of nodes` in that case.
And then I would have no idea what to put here :D

tuommaki: Oh! Right. Then it must be rather like wtf.. `0.00375` (coming from 3 QPS / 20 clusters / 40 nodes per cluster) ?

tuommaki: ...but that allows then single kubelet to issue a query once every 266s (bit over 4 minutes..) - not sure if that works very well either...

calvix:
lgtm if it helps and the cluster is still working :)

fiunchinho:
I will merge this later today unless people raise their hand!

MarcelMue:
Raising my hand :hand: 
This should be moved to v12 - v11 is an active release already. Would be cool if you can take care of moving it.

fiunchinho:
Done

PR #582, opened 2019-10-15
Add VMSS collector

MarcelMue:


service/collector/vmss.go
```diff
@@ -0,0 +1,183 @@
...
+					reads = 0
+				}
+			}
+			var writes int64
+			{
+				writes, err = strconv.ParseInt(r.Response().Header.Get("x-ms-ratelimit-remaining-subscription-writes"), 10, 64)
+				if err != nil {
+					writes = 0
+				}
+			}
```
fiunchinho: IIUC, it seems that we need to do a _write request_ to get the write rate limit. Listing things will only return the read rate limit.

service/collector/vmss.go
```diff
@@ -0,0 +1,183 @@
...
+			var reads int64
+			{
+				reads, err = strconv.ParseInt(r.Response().Header.Get("x-ms-ratelimit-remaining-subscription-reads"), 10, 64)
+				if err != nil {
+					reads = 0
+				}
+			}
+			var writes int64
+			{
+				writes, err = strconv.ParseInt(r.Response().Header.Get("x-ms-ratelimit-remaining-subscription-writes"), 10, 64)
```
whites11: There are 8 HTTP headers exposed in total (mentioned in the documentation), should we consider getting metrics for all of them?
https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-request-limits#remaining-requests

service/collector/vmss.go
MarcelMue: Might be, but I am not sure how we could send a potentially non disruptive write request.

whites11: What about creating/updating a tag on the resource group?

fiunchinho: That sounds good to me

service/collector/vmss.go
```diff
@@ -138,7 +145,22 @@ func (u *VMSS) Collect(ch chan<- prometheus.Metric) error {
 			}
 			var writes int64
 			{
-				writes, err = strconv.ParseInt(r.Response().Header.Get("x-ms-ratelimit-remaining-subscription-writes"), 10, 64)
+				future, err := vmssClient.CreateOrUpdate(ctx, key.ResourceGroupName(cr), *vmss.Name, vmss)
```
MarcelMue: I am not sure if this is a good idea. IIRC metrics collection can run in parallel in the operator to its normal execution.
There might be conflicts where you get the information and the operator does something with it before you createOrUpdate here. 

I think a mutex lock would be necessary for that.

service/collector/vmss.go
```diff
@@ -0,0 +1,234 @@
...
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890"
+
+func randStringBytes(n int) string {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+
+	return string(b)
+}
```
fiunchinho: should I use UUID's instead of generating my own random alphanumeric string? What do you think?

whites11: You could use UUIDs, yes. That will make it 100% unique

tuommaki: I figure random string is fine. UUID could be easily confusing when seen because Microsoft uses GUIDs for everything.

tuommaki:
From my first review round I figure this is probably _ready for review_ from sig-operator (i.e. no more _draft_ anymore) :slightly_smiling_face:

service/collector/vmss.go
```diff
@@ -0,0 +1,234 @@
...
+				return microerror.Mask(err)
+			}
+
+			deleteResponse, err := azureClients.GroupsClient.Delete(ctx, resourceGroupName)
+
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			writes, err = strconv.ParseFloat(deleteResponse.Response().Header.Get(remainingWritesHeaderName), 64)
```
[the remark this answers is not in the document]
fiunchinho: Yes, according to https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-request-limits

service/collector/vmss.go
```diff
@@ -0,0 +1,233 @@
+package collector
+
+import (
+	"context"
+	"fmt"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/go-autorest/autorest/to"
```
xh3b4sd: Imports please. See https://github.com/giantswarm/fmt/blob/master/go/imports.md.

service/collector/vmss.go
```diff
@@ -0,0 +1,233 @@
...
+				resources.Group{
+					Location: to.StringPtr(u.location),
+				})
+
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			deleteResponse, err := azureClients.GroupsClient.Delete(ctx, resourceGroupName)
+
```
xh3b4sd:
```suggestion
```

service/collector/vmss.go
```diff
@@ -0,0 +1,233 @@
...
+		var writes float64
+		{
+			resourceGroupName := randStringBytes(16)
+			_, err := azureClients.GroupsClient.CreateOrUpdate(
+				ctx,
+				resourceGroupName,
+				resources.Group{
+					Location: to.StringPtr(u.location),
+				})
+
```
xh3b4sd:
```suggestion
```

service/collector/vmss.go
```diff
@@ -0,0 +1,234 @@
...
+				)
+			}
+		}
+
+		// Remaining write requests can be retrieved sending a write request.
+		// We create a resource group with a random name, so we can delete it.
+		var writes float64
+		{
+			resourceGroupName := randStringBytes(16)
+			_, err := azureClients.GroupsClient.CreateOrUpdate(
```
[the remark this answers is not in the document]
fiunchinho: I added a prefix to the resource group name. Also I tagged it, so that later can be easily found using the SDK.

service/collector/vmss.go
```diff
@@ -0,0 +1,234 @@
...
+
+		var reads float64
+		{
+			vmssResponse, err := azureClients.VirtualMachineScaleSetsClient.ListAll(ctx)
+			if err != nil {
+				u.logger.Log("level", "warning", "message", "an error occurred during the scraping of current compute resource VMSS information", "stack", fmt.Sprintf("%v", err))
+			} else {
+				reads, err = strconv.ParseFloat(vmssResponse.Response().Header.Get(remainingReadsHeaderName), 64)
+				if err != nil {
+					reads = 0
```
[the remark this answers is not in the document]
fiunchinho: Done

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+const (
+	remainingReadsHeaderName  = "x-ms-ratelimit-remaining-subscription-reads"
+	remainingWritesHeaderName = "x-ms-ratelimit-remaining-subscription-writes"
+)
+
+var (
+	ReadsDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName("azure_operator", "RateLimit", "reads"),
```
kopiczko: Shouldn't this be rather following?

```suggestion
		prometheus.BuildFQName("azure_operator", "rate_limit", "reads"),
```

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+		"Remaining number of reads allowed.",
+		[]string{
+			"subscription",
+			"clientid",
+		},
+		nil,
+	)
+	WritesDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName("azure_operator", "RateLimit", "writes"),
```
kopiczko:
```suggestion
		prometheus.BuildFQName("azure_operator", "rate_limit", "writes"),
```

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+	G8sClient versioned.Interface
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+
+	// EnvironmentName is the name of the Azure environment used to compute the
+	// azure.Environment type. See also
+	// https://godoc.org/github.com/Azure/go-autorest/autorest/azure#Environment.
+	EnvironmentName          string
+	Location                 string
+	HostAzureClientSetConfig client.AzureClientSetConfig
```
kopiczko:
```suggestion
	CPAzureClientSetConfig client.AzureClientSetConfig
```

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+type RateLimit struct {
+	g8sClient versioned.Interface
+	k8sClient kubernetes.Interface
+	logger    micrologger.Logger
+
+	environmentName          string
+	location                 string
+	HostAzureClientSetConfig client.AzureClientSetConfig
```
kopiczko:
```suggestion
	cpAzureClientSetConfig client.AzureClientSetConfig
```

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+	// We track RateLimit metrics for each client labeled by SubscriptionID and
+	// ClientID.
+	// That way we prevent duplicated metrics.
+	for _, clientConfig := range clientConfigBySubscription {
+		azureClients, err := client.NewAzureClientSet(clientConfig)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		resourceGroup, err := u.createEmptyResourceGroup(ctx, azureClients)
```
kopiczko: When creating randomly named RG on each Collect it's likely to have an orphaned RG. E.g. when we hit rate limits.

I would prefer having some RG with a fixed name and never delete it. I assume you can take write limit from  `azureClients.GroupsClient.CreateOrUpdate` call.

service/collector/rate_limit.go
```diff
@@ -0,0 +1,262 @@
...
+		{
+			groupResponse, err := azureClients.GroupsClient.Get(ctx, *resourceGroup.Name)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			reads, err = strconv.ParseFloat(groupResponse.Response.Header.Get(remainingReadsHeaderName), 64)
+			if err != nil {
+				u.logger.Log("level", "warning", "message", "an error occurred parsing to float the value inside the rate limiting header for read requests", "stack", microerror.Stack(microerror.Mask(err)))
+				reads = 0
```
kopiczko: Would it make sense to have error counter here and alert on it as a fallback?

service/collector/rate_limit.go
tuommaki: If this works, then +1.

service/collector/rate_limit.go
fiunchinho: Done

service/collector/rate_limit.go
fiunchinho: Done

service/collector/rate_limit.go
```diff
@@ -0,0 +1,269 @@
...
+
+	"github.com/giantswarm/azure-operator/client"
+	"github.com/giantswarm/azure-operator/service/controller/v9/key"
+	"github.com/giantswarm/azure-operator/service/credential"
+)
+
+const (
+	remainingReadsHeaderName  = "x-ms-ratelimit-remaining-subscription-reads"
+	remainingWritesHeaderName = "x-ms-ratelimit-remaining-subscription-writes"
+	resourceGroupName         = "azure-operator-empty-rg-for-collector"
```
kopiczko:
```suggestion
	resourceGroupName         = "azure-operator-empty-rg-for-metrics"
```

service/collector/rate_limit.go
```diff
@@ -0,0 +1,253 @@
...
+	WritesDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName(metricsNamespace, metricsSubsystem, "writes"),
+		"Remaining number of writes allowed.",
+		[]string{
+			"subscription",
+			"clientid",
+		},
+		nil,
+	)
+	ReadsErrorCounter prometheus.Counter = prometheus.NewCounter(prometheus.CounterOpts{
```
kopiczko: Correct me if I'm wrong but I think they need to be registered somewhere. Check `prometheus.MustRegister`.

fiunchinho: I think you are right. But that means this counter is not working? https://github.com/giantswarm/azure-operator/blob/af47ec54be13a7b1b04b776262b0c67770992ab8/service/collector/usage.go#L91-L96

It's the one I based my counter on.

fiunchinho:
Seems to be working fine

image

All the available metrics

image

service/collector/rate_limit.go
kopiczko: Screenshot 2019-10-31 at 13 31 01

I don't see metric for that one. Maybe there are no errors but I guess it isn't registered.

PR #590, opened 2019-10-25
wire v12

tfussell:
Setting up WIP version after 9.0 release


service/controller/v12/cloudconfig/base_extension.go
```diff
@@ -6,8 +6,8 @@ import (
 
 	"github.com/giantswarm/azure-operator/client"
 	"github.com/giantswarm/azure-operator/service/controller/setting"
-	"github.com/giantswarm/azure-operator/service/controller/v11/encrypter"
-	"github.com/giantswarm/azure-operator/service/controller/v11/key"
+	"github.com/giantswarm/azure-operator/service/controller/v12/encrypter"
```
MarcelMue: We usually do imports separately from wiring
Fine for me here

tfussell: Oh, this sort of felt like wiring to me. I'll do it in the copy step next time (if VOO doesn't land first).

PR #591, opened 2019-10-28
Document credentials used when talking to the Azure API

fiunchinho:


docs/azure_authentication.md
```diff
@@ -0,0 +1,43 @@
...
+			
+## Azure Operator
+This process runs on the worker nodes on the control planes. [When deployed](https://github.com/giantswarm/installations/blob/ad518ba95584685bf37cd9c3cca403115153dc14/ghost/draughtsman-secret-values.yaml#L20-L27), it receives a set of [parameters](https://github.com/giantswarm/azure-operator/blob/d0185b1de2cf6975f8e795a38bb55dfc875ecdeb/main.go#L125-L126) 
+such as the `Azure.ClientID`, the `Azure.ClientSecret`, or `Azure.MSI.Enabled` to select whether to enabled [Managed Service Identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview).
+
+## Credentiald server
+It's using the same credentials than Azure Operator is using.
+
+## Control plane clusters
+## Cloud Provider k8s integration on masters
```
kopiczko:
```suggestion
### Cloud Provider k8s integration on masters
```

docs/azure_authentication.md
```diff
@@ -0,0 +1,43 @@
+# Azure Operator authentication with Azure
```
kopiczko: I think this is related: https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/azure-service-principals.md

Maybe you want to link it somewhere in this doc.

fiunchinho: Done

fiunchinho:
Do you think I should put these docs on the giantswarm repository? where exactly? Otherwise we can just merge it here.

T-Kukawka:
You could merge it, but is it the ops-recipe then?

docs/azure_authentication.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Azure operator running on the control plane worker nodes
+- `credentiald` running on the control plane worker nodes
+- Kubernetes cloud provider integration running on the control plane master nodes
+- Kubernetes cloud provider integration running on the control plane worker nodes
+- Kubernetes cloud provider integration running on tenant clusters master nodes
+- Kubernetes cloud provider integration running on tenant clusters worker nodes
+			
+## Azure Operator
+This process runs on the worker nodes on the control planes. [When deployed](https://github.com/giantswarm/installations/blob/ad518ba95584685bf37cd9c3cca403115153dc14/ghost/draughtsman-secret-values.yaml#L20-L27), it receives a set of [parameters](https://github.com/giantswarm/azure-operator/blob/d0185b1de2cf6975f8e795a38bb55dfc875ecdeb/main.go#L125-L126)
```
fiunchinho: Correct me if I'm wrong, but I think what I wrote is not 100% true. 

The Azure Operator only uses the credentials parameters to set up some DNS and VPN resources. It uses a different set of credentials to create the clusters: the credentials stored in a Kubernetes secret referenced from the `AzureConfig` object. 
These are typically the same set of credentials, but they don't have to be.

Can you confirm my understanding is correct? In that case, I'd add that to the document.

kopiczko: Yes it uses configured credentials to set DNS and VPN gateways in CP subscription and then it uses so called BYOC (bring your own cloud) credentials referenced in AzureConfig CR. They are configured per cluster basis. This is because we allow to create clusters in different subscriptions. This is similar for AWS.

fiunchinho: Updated

PR #592, opened 2019-11-05
Added new virtual disk on /var/lib/kubelet for workers and master

whites11:
The goal of this PR is to add to the master and workers template a new volume mounted on /var/lib/kubelet.
The volume size will be adjustable by modifying the CR AzureConfig.
This PR also includes a fix on the previously existing Docker volume which wasn't mounted correctly on worker nodes.
The related issue is https://github.com/giantswarm/giantswarm/issues/7526

### DIFF
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -177,8 +177,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 				ws, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue)
 				desiredDiskSizes := map[string]int32{
-					dockerDiskName:  int32(customObject.Spec.Azure.Masters[0].DockerVolumeSizeGB),
-					kubeletDiskName: int32(customObject.Spec.Azure.Masters[0].KubeletVolumeSizeGB),
+					key.DockerDiskName:  int32(customObject.Spec.Azure.Masters[0].DockerVolumeSizeGB),
+					key.KubeletDiskName: int32(customObject.Spec.Azure.Masters[0].KubeletVolumeSizeGB),
 				}
 
 				masterInstanceToUpdate, masterInstanceToDrain, masterInstanceToReimage, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue, desiredDiskSizes)
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -182,6 +182,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 				r.logger.LogCtx(ctx, "level", "debug", "message", "processing master VMSSs")
 
 				ws, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue)
+				desiredDiskSizes := map[string]int32{
+					dockerDiskName:  int32(customObject.Spec.Azure.Masters[0].DockerVolumeSizeGB),
+					kubeletDiskName: int32(customObject.Spec.Azure.Masters[0].KubeletVolumeSizeGB),
+				}
+
+				masterInstanceToUpdate, masterInstanceToDrain, masterInstanceToReimage, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue, desiredDiskSizes)
 				if err != nil {
 					return microerror.Mask(err)
 				}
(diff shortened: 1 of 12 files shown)


service/controller/v12/resource/instance/create.go
```diff
@@ -601,7 +612,13 @@ func firstInstanceInProgress(customObject providerv1alpha1.AzureConfig, list []c
 // bundle version of the custom object and the current version bundle version of
 // the instance's tags applied. In case all instances are reimaged
 // firstInstanceToReimage return nil.
-func firstInstanceToReimage(customObject providerv1alpha1.AzureConfig, list []compute.VirtualMachineScaleSetVM, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (*compute.VirtualMachineScaleSetVM, error) {
+func firstInstanceToReimage(
+	customObject providerv1alpha1.AzureConfig,
+	list []compute.VirtualMachineScaleSetVM,
+	instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error),
+	versionValue map[string]string,
+	desiredDiskSizes map[string]int32,
```
tuommaki: We don't usually pass this kind of information as separate parameter, but extract it with key function from `customObject`.

service/controller/v12/resource/instance/create.go
```diff
@@ -613,14 +630,26 @@ func firstInstanceToReimage(customObject providerv1alpha1.AzureConfig, list []co
...
+			desiredDiskSize, ok := desiredDiskSizes[*disk.Name]
+			if !ok {
+				// this disk has no target size specified, ignore it
+				continue
+			}
+			// desired disk size is changed
+			if *disk.DiskSizeGB != desiredDiskSize {
+				return &v, nil
+			}
+		}
```
tuommaki: I'd maybe change this to be something like following:
```go
		for _, disk := range *v.StorageProfile.DataDisks {
			switch(*disk.Name) {
			case key.DockerVolumeDiskName:
				if *disk.DiskSizeGB != key.DockerVolumeSizeGB(customObject) {
					return &v, nil
				}
			case key.KubeletVolumeDiskName:
				if *disk.DiskSizeGB != key.KubeletVolumeSizeGB(customObject) {
					return &v, nil
				}
			}
		}
```

For that you need to add corresponding key functions (similar to this: https://github.com/giantswarm/aws-operator/blob/master/service/controller/key/machine_deployment.go#L16-L18) and constants to match for disk name.

`azure-operator` key functions are here: https://github.com/giantswarm/azure-operator/blob/master/service/controller/v12/key/key.go

service/controller/v12/resource/instance/create.go
```diff
@@ -21,6 +21,8 @@ import (
 
 const (
 	vmssDeploymentName = "cluster-vmss-template"
+	dockerDiskName     = "DockerDisk"
```
tuommaki: Move these to `key.DockerVolumeDiskName` and corresponding `key.KubeletVolumeDiskName` as noted below.

whites11:
> In general this looks good, but extra parameter in function can be avoided when extraction of volume size from CR is done with key functions abstractions. That is also more aligned with our operator code bases in general.

Not sure if this is possible, because the missing information inside firstInstanceToReimage function is if we're handling master or worker nodes in the current function call.
Thanks for the tips anyway, I'll have a look tomorrow morning.

tuommaki:
Let's Meet and pair on this tomorrow :+1:

service/controller/v12/resource/instance/create.go
whites11: Done

whites11:
@tuommaki following what we discussed during today's standup, I'd merge this PR with the current features it brings, which are:

1) dedicated disk for /var/lib/kubelet
2) fix to correctly mount the Docker disk on /var/lib/docker

This will work on new TCs only and will NOT react to disk size changes to the CR.

The idea is to provide an update to the reconciliation process in a different PR.

WDYT?

service/controller/v12/resource/instance/create.go
whites11: removed

tuommaki:
> The idea is to provide an update to the reconciliation process in a different PR.

Works for me. Having configurable volumes in cluster creation is already a big plus. What about clusters that don't mount these yet - would they get missing disks and mount them during upgrade?

whites11:
> > The idea is to provide an update to the reconciliation process in a different PR.
> 
> Works for me. Having configurable volumes in cluster creation is already a big plus. What about clusters that don't mount these yet - would they get missing disks and mount them during upgrade?

First of all, this patch only applies to v12, so tenant clusters that have previous versions are not affected at all.
That being said, AFAIK an existing v12 cluster's instances would not get the new disks attached unless the process of recycling the instances in the VMSS is done manually.
That potentially means that the systemd unit that should mount the new disks will fail (because the disks are not going to be there) but that should be harmless I think.

tuommaki:
> First of all, this patch only applies to v12, so tenant clusters that have previous versions are not affected at all.

But didn't earlier releases have the disk - it just wasn't mounted?

whites11:
> > First of all, this patch only applies to v12, so tenant clusters that have previous versions are not affected at all.
> 
> But didn't earlier releases have the disk - it just wasn't mounted?

Yes, but still the mount unit fix is only applied to v12 clusters. So only those will get the disk mounted (as long as we don't backport the fix to v11 as well)

tuommaki:
> Yes, but still the mount unit fix is only applied to v12 clusters. 

Sure, but clusters shall be eventually upgraded to it :slightly_smiling_face:

PR #593, opened 2019-11-05
Add the diskSizeGB option to the workers VMSS to allow adjusting root…

whites11:
… disk size

This will allow customizing the root volume size by editing the CR for the tenant cluster.
This is not reactive yet, but it can be edited before the VMSS are spawn.
Plan is to make it automatically reconciled on change (but will ask for help).
Tested on godsmack, looks good.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -263,7 +263,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:fcb75264e7f1a4276044b71f706baf6b1eba4e619f3a3bc95f33c677cbe25f6c"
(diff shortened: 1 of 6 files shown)


tuommaki:
Instead of configuring root volume size, I'd prefer adjustable `/var/lib/kubelet` which is the location for `EmptyDir` volumes.

whites11:
> Instead of configuring root volume size, I'd prefer adjustable `/var/lib/kubelet` which is the location for `EmptyDir` volumes.

I was thinking about doing both.
I agree it is more legit to have a separate and size-configurable /var/lib/kubelet (PR already on his way) but having a configurable root disk size won't hurt I reckon.
Opinions more than welcome.

whites11:
We decided to not proceed with this approach anymore

PR #594, opened 2019-11-06
Using CR field to decide docker volume size

whites11:
There is a `DockerVolumeSizeGB` field in the `AzureConfig` CRD that allows customizing the Azure disk mounted on `/var/lib/docker` for both masters and workers.
We are not using that field yet in the VMSS template, and we hardcoded 50Gb size for the docker disk.
This PR aims at using the CRD field for sizing the docker disk.
I tested it by setting the Disk size manually on the CR and it worked just fine.

### DIFF
--- a/service/controller/v12/resource/instance/template/main.json
+++ b/service/controller/v12/resource/instance/template/main.json
@@ -106,7 +106,7 @@
               },
               {
                 "name":"DockerDisk",
-                "sizeGB":"[parameters('masterNodes')[0].dockerVolumeSizeGB]"
+                "sizeGB":"[if(greater(parameters('masterNodes')[0].dockerVolumeSizeGB, 0), parameters('masterNodes')[0].dockerVolumeSizeGB, 50)]"
               }
             ]
           },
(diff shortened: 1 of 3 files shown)


service/controller/v12/resource/instance/template/main.json
```diff
@@ -106,7 +106,7 @@
               },
               {
                 "name":"DockerDisk",
-                "sizeGB":50
+                "sizeGB":"[if(greater(parameters('masterNodes')[0].dockerVolumeSizeGB, 0), parameters('masterNodes')[0].dockerVolumeSizeGB, 50)]"
```
fiunchinho: Would it make sense to have this conditional in the `newDeployment` method? Setting the defaults in Go code seems easier to read. What do you think?

whites11: Default values and hardcoded values are all in the ARM templates. I feel putting this single default in the Go code would be choosing a different solution for an already solved problem, but I'm open to other opinions on this.

tuommaki: @fiunchinho has a good point, but I think it's ok to have this in ARM template as a last catch-all parameter (so in all cases the template itself is functional with minimum configuration).

For legacy clusters (that all Azure clusters are) the defaulting is slightly distributed in several places (`api`, `cluster-service`, `kubernetesd` and `azure-operator`), but for newer style clusters with Cluster API types, the defaulting takes place in `api` now and later we most probably will introduce validation and mutating webhooks that can complement minimal CR YAMLs.

PR #596, opened 2019-11-14
Fixed an error in the process of checking if the master nodes were in…

whites11:
… an updating phase before proceeding with the worker nodes update.

Related issue: https://github.com/giantswarm/giantswarm/issues/7659

@tuommaki I think I found the cause of the infinite looping or at least a mitigation.
If the merge request and my comments are not clear please let's talk about it.
Thx

### DIFF
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -421,8 +421,6 @@ func (r *Resource) nextInstance(ctx context.Context, customObject providerv1alph
 	{
 		r.logger.LogCtx(ctx, "level", "debug", "message", "looking for the next instance to be updated, drained or reimaged")
 
-		emptyWorkingSet := workingSet{}
-
 		ws, err = getWorkingSet(customObject, instances, drainerConfigs, instanceNameFunc, versionValue)
 		if IsVersionBlobEmpty(err) {
 			// When no version bundle version is found it means the cluster just got
@@ -431,36 +429,36 @@ func (r *Resource) nextInstance(ctx context.Context, customObject providerv1alph
 			// must not select an instance to be reimaged because we would roll a node
 			// that just got created and is already up to date.
 			r.logger.LogCtx(ctx, "level", "debug", "message", "no instance found to be updated, drained or reimaged")
-			return &emptyWorkingSet, nil
+			return nil, nil
 		} else if err != nil {
-			return &emptyWorkingSet, microerror.Mask(err)
+			return nil, microerror.Mask(err)
 		}
 
 		if !ws.IsWIP() {
 			// Neither did we find an instance to be updated nor to be reimaged.
 			// Nothing has to be done or we already processes all instances.
 			r.logger.LogCtx(ctx, "level", "debug", "message", "no instance found to be updated, drained or reimaged")
-			return &emptyWorkingSet, nil
+			return nil, nil
 		}
 
 		if ws.instanceToUpdate != nil {
 			instanceName, err := instanceNameFunc(customObject, *ws.instanceToUpdate.InstanceID)
 			if err != nil {
-				return &emptyWorkingSet, microerror.Mask(err)
+				return nil, microerror.Mask(err)
 			}
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be updated", instanceName))
 		}
 		if ws.instanceToDrain != nil {
 			instanceName, err := instanceNameFunc(customObject, *ws.instanceToDrain.InstanceID)
 			if err != nil {
-				return &emptyWorkingSet, microerror.Mask(err)
+				return nil, microerror.Mask(err)
 			}
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be drained", instanceName))
 		}
 		if ws.instanceToReimage != nil {
 			instanceName, err := instanceNameFunc(customObject, *ws.instanceToReimage.InstanceID)
 			if err != nil {
-				return &emptyWorkingSet, microerror.Mask(err)
+				return nil, microerror.Mask(err)
 			}
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found instance '%s' has to be reimaged", instanceName))
 		}
@@ -560,7 +558,7 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
 func getWorkingSet(customObject providerv1alpha1.AzureConfig, instances []compute.VirtualMachineScaleSetVM, drainerConfigs []corev1alpha1.DrainerConfig, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (*workingSet, error) {
 	var err error
 
-	var ws workingSet
+	var ws *workingSet
 
 	instanceInProgress := firstInstanceInProgress(customObject, instances)
 	if instanceInProgress != nil {
@@ -576,12 +574,12 @@ func getWorkingSet(customObject providerv1alpha1.AzureConfig, instances []comput
 	var instanceToReimage *compute.VirtualMachineScaleSetVM
 	instanceToReimage, err = firstInstanceToReimage(customObject, instances, instanceNameFunc, versionValue)
 	if err != nil {
-		return &ws, microerror.Mask(err)
+		return ws, microerror.Mask(err)
 	}
 	if instanceToReimage != nil {
 		instanceName, err := instanceNameFunc(customObject, *instanceToReimage.InstanceID)
 		if err != nil {
-			return &ws, microerror.Mask(err)
+			return ws, microerror.Mask(err)
 		}
 		if isNodeDrained(drainerConfigs, instanceName) {
 			return ws.WithInstanceToReimage(instanceToReimage), nil
@@ -590,7 +588,7 @@ func getWorkingSet(customObject providerv1alpha1.AzureConfig, instances []comput
 		}
 	}
 
-	return &ws, nil
+	return ws, nil
 }
 
 // firstInstanceInProgress returns the first instance in the list not having a
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -162,7 +162,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			drainerConfigs = list.Items
 		}
 
-		var masterUpgraded bool
+		var masterUpgradeInProgress bool
 		{
 			allMasterInstances, err := r.allInstances(ctx, customObject, key.MasterVMSSName)
 			if IsScaleSetNotFound(err) {
@@ -172,7 +172,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			} else {
 				r.logger.LogCtx(ctx, "level", "debug", "message", "processing master VMSSs")
 
-				masterInstanceToUpdate, masterInstanceToDrain, masterInstanceToReimage, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue)
+				masterInstanceToUpdate, masterInstanceToDrain, masterInstanceToReimage, instanceAlreadyBeingUpdated, err := r.nextInstance(ctx, customObject, allMasterInstances, drainerConfigs, key.MasterInstanceName, versionValue)
 				if err != nil {
 					return microerror.Mask(err)
 				}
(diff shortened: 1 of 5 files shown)


service/controller/v12/resource/instance/create.go
```diff
@@ -112,12 +112,19 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		} else {
 			r.debugger.LogFailedDeployment(ctx, d, err)
 
-			err := r.deleteResourceStatus(customObject, Stage, DeploymentInitialized)
-			if err != nil {
-				return microerror.Mask(err)
+			if key.IsFinalProvisioningState(s) {
```
whites11: The state could be non final in two cases:
1) deployment is running
2) deployment is failed.
If the deployment is failed, we can retry applying it in the next loop, so we cancel the resource status to make it happen.
If the deployment is running, we shouldn't do anything more until it finishes otherwise we keep updating the deployment at the beginning of the loop and never get to complete the reconciliation.

That's the goal of this if.

service/controller/v12/resource/instance/create.go
```diff
@@ -554,38 +555,38 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
 
 // findActionableInstance either returns an instance to update or an instance to
 // reimage, but never both at the same time.
-func findActionableInstance(customObject providerv1alpha1.AzureConfig, instances []compute.VirtualMachineScaleSetVM, drainerConfigs []corev1alpha1.DrainerConfig, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (*compute.VirtualMachineScaleSetVM, *compute.VirtualMachineScaleSetVM, *compute.VirtualMachineScaleSetVM, error) {
+func findActionableInstance(customObject providerv1alpha1.AzureConfig, instances []compute.VirtualMachineScaleSetVM, drainerConfigs []corev1alpha1.DrainerConfig, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (workingSet, error) {
```
xh3b4sd: What about renaming this function now? It creates a new working set now as it seems.

service/controller/v12/resource/instance/create.go
xh3b4sd: Is there potential for some infinite loop or deadlock situation?

service/controller/v12/resource/instance/workingset.go
```diff
@@ -0,0 +1,45 @@
+package instance
+
+import "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+
+type workingSet struct {
+	instanceToUpdate            *compute.VirtualMachineScaleSetVM
+	instanceToDrain             *compute.VirtualMachineScaleSetVM
+	instanceToReimage           *compute.VirtualMachineScaleSetVM
+	instanceAlreadyBeingUpdated *compute.VirtualMachineScaleSetVM
+}
```
tuommaki: Proposal for slightly different API:

```go
func (ws *workingSet) WithInstanceToUpdate(instance *compute.VirtualMachineScaleSetVM) *workingSet {
	if ws == nil {
		ws = &workingSet{}
	}
	ws.instanceToUpdate = instance
	return ws
}

func (ws *workingSet) WithInstanceToDrain(instance *compute.VirtualMachineScaleSetVM) *workingSet {
	if ws == nil {
		ws = &workingSet{}
	}
	ws.instanceToDrain = instance
	return ws
}

func (ws *workingSet) WithInstanceToReimage(instance *compute.VirtualMachineScaleSetVM) *workingSet {
	if ws == nil {
		ws = &workingSet{}
	}
	ws.instanceToReimage = instance
	return ws
}

func (ws *workingSet) WithInstanceAlreadyBeingUpdated(instance *compute.VirtualMachineScaleSetVM) *workingSet {
	if ws == nil {
		ws = &workingSet{}
	}
	ws.instanceAlreadyBeingUpdated = instance
	return ws
}

func (ws *workingSet) IsWIP() bool {
	if ws == nil {
		return false
	}

	return ws.instanceToUpdate != nil ||
		ws.instanceToDrain != nil ||
		ws.instanceToReimage != nil ||
		ws.instanceAlreadyBeingUpdated != nil
}
```

This would then work as follows:
```go
func Test_WorkingSet(t *testing.T) {
	var ws *workingSet

	if ws.IsWIP() != false {
		t.Fatal("<nil>.IsWIP() == true")
	}

	ws = ws.WithInstanceToUpdate(&compute.VirtualMachineScaleSetVM{})

	if ws.IsWIP() != true {
		t.Fatal("workingSet{...}.IsWIP() == false")
	}

	ws = nil

	ws = ws.WithInstanceToDrain(&compute.VirtualMachineScaleSetVM{})

	if ws.IsWIP() != true {
		t.Fatal("workingSet{...}.IsWIP() == false")
	}

	ws = nil

	ws = ws.WithInstanceToReimage(&compute.VirtualMachineScaleSetVM{})

	if ws.IsWIP() != true {
		t.Fatal("workingSet{...}.IsWIP() == false")
	}

	ws = nil

	ws = ws.WithInstanceAlreadyBeingUpdated(&compute.VirtualMachineScaleSetVM{})

	if ws.IsWIP() != true {
		t.Fatal("workingSet{...}.IsWIP() == false")
	}

	ws = nil
}
```

Few pros with this:
* Readability as text / story is slightly nicer (opinionated of course, but still).
* You can combine different elements while still having one-line function to "set them" - given you declare variable once in the beginning.
* `IsWIP()` works with `nil` and value.
* Return in error cases can be usual `return nil, microerror.Mask(err)`

No strong opinions on this. Just a proposal. :slightly_smiling_face:

service/controller/v12/resource/instance/create.go
```diff
@@ -209,8 +214,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		// worker instance selected to be reimaged in order to not update its
 		// version information. The next reconciliation loop will catch up here
 		// and instruct the worker instance to be reimaged again.
-		var workerUpgraded bool
-		if !masterUpgraded {
+		var workerUpgradeInProgess bool
+		if !masterUpgradeInProgress {
```
xh3b4sd: `masterUpgraded` and `masterUpgradeInProgress` are semantically different. Yet the condition negates with `!` in both cases. That looks wrong to me. How can this work that way now? Was the initial wording just wrong?

whites11: Yes, as far as I understood the original wording was misleading, thus the rename.

service/controller/v12/resource/instance/create.go
whites11: function renamed

service/controller/v12/resource/instance/create.go
whites11: In theory there is, when azure deployment never gets out of the "Running" state.
In that case we will never update that deployment again.
Not sure if this is possible or not (I'd expect at least a timeout on the azure side) and not sure how we could possibly handle this case it anyway.

service/controller/v12/resource/instance/workingset.go
whites11: I like it, applied

service/controller/v12/resource/instance/create.go
tuommaki: Infinite loops & sort of "deadlocks" here are difficult to handle in code, but we do have already some alerts for them and we can improve them even further when we see what's the right direction.

service/controller/v12/resource/instance/create.go
```diff
@@ -411,59 +414,59 @@ func (r *Resource) deleteDrainerConfig(ctx context.Context, customObject provide
...
 
-	var instanceToUpdate *compute.VirtualMachineScaleSetVM
-	var instanceToDrain *compute.VirtualMachineScaleSetVM
-	var instanceToReimage *compute.VirtualMachineScaleSetVM
+	var ws *workingSet
 	{
 		r.logger.LogCtx(ctx, "level", "debug", "message", "looking for the next instance to be updated, drained or reimaged")
 
-		instanceToUpdate, instanceToDrain, instanceToReimage, err = findActionableInstance(customObject, instances, drainerConfigs, instanceNameFunc, versionValue)
+		emptyWorkingSet := workingSet{}
```
tuommaki: No need for this.

service/controller/v12/resource/instance/create.go
```diff
@@ -411,59 +414,59 @@ func (r *Resource) deleteDrainerConfig(ctx context.Context, customObject provide
...
+		ws, err = getWorkingSet(customObject, instances, drainerConfigs, instanceNameFunc, versionValue)
 		if IsVersionBlobEmpty(err) {
 			// When no version bundle version is found it means the cluster just got
 			// created and the version bundle versions are not yet tracked within the
 			// parameters of the tenant cluster's VMSS deployment. In this case we
 			// must not select an instance to be reimaged because we would roll a node
 			// that just got created and is already up to date.
 			r.logger.LogCtx(ctx, "level", "debug", "message", "no instance found to be updated, drained or reimaged")
-			return nil, nil, nil, nil
+			return &emptyWorkingSet, nil
```
tuommaki:
```suggestion
			return nil, nil
```
With the latest `workingSet` implementation, it's safe to call its methods on nil value. Also see similar patterns later.

service/controller/v12/resource/instance/create.go
```diff
@@ -552,40 +555,42 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
 }
 
-// findActionableInstance either returns an instance to update or an instance to
+// getWorkingSet either returns an instance to update or an instance to
 // reimage, but never both at the same time.
-func findActionableInstance(customObject providerv1alpha1.AzureConfig, instances []compute.VirtualMachineScaleSetVM, drainerConfigs []corev1alpha1.DrainerConfig, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (*compute.VirtualMachineScaleSetVM, *compute.VirtualMachineScaleSetVM, *compute.VirtualMachineScaleSetVM, error) {
+func getWorkingSet(customObject providerv1alpha1.AzureConfig, instances []compute.VirtualMachineScaleSetVM, drainerConfigs []corev1alpha1.DrainerConfig, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error), versionValue map[string]string) (*workingSet, error) {
 	var err error
 
+	var ws workingSet
```
tuommaki:
```suggestion
	var ws *workingSet
```

service/controller/v12/resource/instance/create.go
```diff
@@ -552,40 +555,42 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
 	if instanceToUpdate != nil {
-		return instanceToUpdate, nil, nil, nil
+		return ws.WithInstanceToUpdate(instanceToUpdate), nil
 	}
 
 	var instanceToReimage *compute.VirtualMachineScaleSetVM
 	instanceToReimage, err = firstInstanceToReimage(customObject, instances, instanceNameFunc, versionValue)
 	if err != nil {
-		return nil, nil, nil, microerror.Mask(err)
+		return &ws, microerror.Mask(err)
```
tuommaki:
```suggestion
		return ws, microerror.Mask(err)
```

service/controller/v12/resource/instance/create.go
```diff
@@ -552,40 +555,42 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
 	}
 	if instanceToReimage != nil {
 		instanceName, err := instanceNameFunc(customObject, *instanceToReimage.InstanceID)
 		if err != nil {
-			return nil, nil, nil, microerror.Mask(err)
+			return &ws, microerror.Mask(err)
```
tuommaki:
```suggestion
			return ws, microerror.Mask(err)
```

service/controller/v12/resource/instance/create.go
```diff
@@ -552,40 +555,42 @@ func containsInstanceID(list []compute.VirtualMachineScaleSetVM, id string) bool
...
-			return nil, nil, instanceToReimage, nil
+			return ws.WithInstanceToReimage(instanceToReimage), nil
 		} else {
-			return nil, instanceToReimage, nil, nil
+			return ws.WithInstanceToDrain(instanceToReimage), nil
 		}
 	}
 
-	return nil, nil, nil, nil
+	return &ws, nil
```
tuommaki:
```suggestion
	return ws, nil
```

service/controller/v12/resource/instance/create.go
whites11: I kinda have preferred to have an explicit difference between the nil case and the emptyWorkingSet, despite the fact that they behave the same way.
But I applied your suggestions anyway.

service/controller/v12/resource/instance/create.go
whites11: done

service/controller/v12/resource/instance/create.go
whites11: done

service/controller/v12/resource/instance/create.go
whites11: done

service/controller/v12/resource/instance/create.go
whites11: done

service/controller/v12/resource/instance/create.go
whites11: done

service/controller/v12/resource/instance/create.go
tuommaki: Let's keep that in mind and change it if there are any problems with the current approach. I do also agree that it could confuse sometimes when working with nil (while even valid), but in this case it's not exposed/exported type, it's use is quite narrow and I personally liked the convenience / conciseness of it.

FYI: As an author in general you have right to disagree with reviewer[s] on things like this and proceed with your approach. It's totally fine and happens sometimes. Disagree and commit it is.

PR #597, opened 2019-11-18
Vendor update errors & statusresource

tuommaki:


tuommaki:
Allows `azure-operator` now to continue reconciliation even if tenant cluster API has expired certs. That is needed so that `azure-operator` can reconcile the cluster state into such that certs get updated.

PR #598, opened 2019-11-18
Tag for autoscaling

fiunchinho:
The `cluster-autoscaler` auto-discovery feature needs that we tag our VMSS. These labels will be later passed as parameters to `cluster-autoscaler`, so that the autoscaler knows which VMSS to scale.

Any tag can be used, but for consistency sake, I've chosen the same tags that we are using on AWS for `cluster-autoscaler`.

### DIFF
--- a/service/controller/v12/resource/instance/template/vmss.json
+++ b/service/controller/v12/resource/instance/template/vmss.json
@@ -181,7 +181,10 @@
       "name":"[parameters('vmssName')]",
       "location":"[parameters('location')]",
       "tags":{
-        "provider":"[toUpper(parameters('GiantSwarmTags').provider)]"
+        "provider":"[toUpper(parameters('GiantSwarmTags').provider)]",
+        "k8s.io/cluster-autoscaler/enabled":"true",
+        "[concat('k8s.io/cluster-autoscaler/', parameters('vmssName'))]": "true"
+
       },
       "sku":{
         "name":"[parameters('vmssVmSize')]",
--- a/service/controller/v12/resource/instance/template/vmss.json
+++ b/service/controller/v12/resource/instance/template/vmss.json
@@ -1,9 +1,9 @@
 {
-  "$schema":"http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "$schema":"https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
   "contentVersion":"1.0.0.0",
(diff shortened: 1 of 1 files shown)


service/controller/v12/resource/instance/template/vmss.json
```diff
@@ -181,7 +181,10 @@
       "name":"[parameters('vmssName')]",
       "location":"[parameters('location')]",
       "tags":{
-        "provider":"[toUpper(parameters('GiantSwarmTags').provider)]"
+        "provider":"[toUpper(parameters('GiantSwarmTags').provider)]",
+        "k8s.io/cluster-autoscaler/enabled":"true",
+        "[concat('k8s.io/cluster-autoscaler/', parameters('vmssName'))]": "true"
```
whites11: Did you test it? Never saw an expression used as a key in the template (only as value).
I think it should work though.

fiunchinho: I wasn't sure either. I tested it using this endpoint https://docs.microsoft.com/en-us/rest/api/resources/deployments/validate.

whites11: What about testing it in the field?
We can do it together if you want.

whites11: You can also check that the outcome is what you actually need it to be (i.e. that the tags are really there in the deployment)

fiunchinho: When deploying it for real, I got the following errors
```
The tag names 'k8s.io/cluster-autoscaler/enabled, k8s.io/cluster-autoscaler/i9ssh-master' have reserved characters '<,>,%,&,\\\\,?,/' or control characters. These characters are only allowed for tags that start with the prefix 'hidden, link'.
```

Which tags should we use? Since any tag can be used, we could just add a `name` tag with the current name, and another one to indicate that the cluster is being auto scaled. Something like
```name: i9ssh
cluster-autoscaler-enabled: true
```

WDYT?

whites11: I'd use the `cluster-autoscaled-` prefix for the name as well in order to avoid future changes to give a different meaning to the  `name` tag given it's generic name and break autoscaling.

fiunchinho: done

fiunchinho: It works now

image

PR #600, opened 2019-11-19
Using checksum calculation for instances templates

whites11:
To avoid applying the Deployments again and again on azure

Issue: https://github.com/giantswarm/giantswarm/issues/7687

WARNING: there is quite a big refactoring in this issue that I felt like doing to make the single files a little smaller and hopefully more readable.


service/controller/v12/resource/instance/create.go
```diff
@@ -3,9 +3,7 @@ package instance
 import (
```
whites11: I refactored this go file to make the EnsureCreated function simpler.
I simply left the basic state-machine logic in that function and moved the code handling each state in a separate function.
I also moved some of the pre-existing functions to different go files in the same module.
Sorry about the ugly diff, but I reckon this refactoring will help writing more unit tests and the general readability of the code

whites11:
The e2e tests are failing for an unknown reason.
I made some troubleshooting on `e2eTestWIPPRClusterState` and the problem there is that calico does not run at all (there is no daemonset in the tenant cluster).
Not sure why, but this doesn't happen when making a real deployment so it might be an e2e issue. 
Any advice on this very appreciated.

tfussell:
I restarted the CircleCI job with SSH so I could see the `azure-operator` logs. Here's what I found ```{"body":"","caller":"github.com/giantswarm/azure-operator/service/controller/v12/debugger/debugger.go:41","controller":"azure-operator","correlationID":"<censored>","event":"update","id":"/subscriptions/<censored>/resourceGroups/ci-wip-20e25-c1c86/providers/Microsoft.Resources/deployments/cluster-main-template","level":"error","loop":"4","message":"deployment failed","name":"cluster-main-template","object":"/apis/provider.giantswarm.io/v1alpha1/namespaces/default/azureconfigs/ci-wip-20e25-c1c86","resource":"deploymentv12","stack":"\u003cnil\u003e","status":"200 OK","time":"2019-11-26T17:29:03.035391+00:00","version":"690"}```
`"stack":"\u003cnil\u003e"`=>`"stack":"<nil>"` so it looks like a nil pointer somewhere. Maybe that can help.

whites11:
> I restarted the CircleCI job with SSH so I could see the `azure-operator` logs. Here's what I found `{"body":"","caller":"github.com/giantswarm/azure-operator/service/controller/v12/debugger/debugger.go:41","controller":"azure-operator","correlationID":"<censored>","event":"update","id":"/subscriptions/<censored>/resourceGroups/ci-wip-20e25-c1c86/providers/Microsoft.Resources/deployments/cluster-main-template","level":"error","loop":"4","message":"deployment failed","name":"cluster-main-template","object":"/apis/provider.giantswarm.io/v1alpha1/namespaces/default/azureconfigs/ci-wip-20e25-c1c86","resource":"deploymentv12","stack":"\u003cnil\u003e","status":"200 OK","time":"2019-11-26T17:29:03.035391+00:00","version":"690"}`
> `"stack":"\u003cnil\u003e"`=>`"stack":"<nil>"` so it looks like a nil pointer somewhere. Maybe that can help.

Where exactly do you see that error? I left the azure operator run until the failure of the pipeline and got no errors at all. See attached log

[azure-operator-ci.log](https://github.com/giantswarm/azure-operator/files/3895836/azure-operator-ci.log)

whites11:
Ok I found the error finally.
The k8s-addons unit is failing in the master with this error:

> Nov 27 12:09:24 ci-wip-20e25-c1c86-master-000000 k8s-addons[1627]: Error from server (NotFound): error when creating "/srv/network_policies.yaml": namespaces "giantswarm" not found
> Nov 27 12:09:24 ci-wip-20e25-c1c86-master-000000 k8s-addons[1627]: failed to apply /srv/network_policies.yaml, retrying in 5 sec

After creating the namespace manually the cluster comes up almost immediately.
Have to check how to fix this

tfussell:
This sounds like a result of https://github.com/giantswarm/k8scloudconfig/pull/609 I guess we could add a check for the giantswarm namespace before installing add-ons or create it if it doesn't exist.

As far as debugging, you can restart a job using the "Restart" drop-down in CircleCI to give SSH access. Then you can SSH into the runner to see logs for all of your components in real time using `docker logs`.

service/controller/v12/resource/instance/azure.go
```diff
@@ -0,0 +1,180 @@
...
+	"context"
+	"fmt"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/azure-operator/service/controller/v12/key"
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) reimageInstance(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance *compute.VirtualMachineScaleSetVM, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string, instanceNameFunc func(customObject providerv1alpha1.AzureConfig, instanceID string) (string, error)) error {
```
MarcelMue: Did this function just move to a different file? What changed? It is hard to tell for me with such a big PR / confusing diff.

whites11:
> It would be much more readable for me if this was split into multiple PRs:
> 
> 1. PR to refactor functions without logic changes (maybe one PR for instance and PR for drainer)
> 2. PR to add logic for checksum calculation.
> 
> Right now with this diff it is really hard for me to see what was actually changed and what was moved.

That makes sense, I'll try to get some sense out of the PR. Thanks for feedback

service/controller/v12/key/key.go
```diff
@@ -15,6 +15,7 @@ const (
 	clusterTagName      = "GiantSwarmCluster"
 	installationTagName = "GiantSwarmInstallation"
 	organizationTagName = "GiantSwarmOrganization"
+	VmssDeploymentName  = "cluster-vmss-template"
```
kopiczko: Not used?

service/controller/v12/resource/instance/error.go
```diff
@@ -99,3 +99,19 @@ var versionBlobEmptyError = &microerror.Error{
...
+
+func IsNilTemplateLinkError(err error) bool {
+	return microerror.Cause(err) == nilTemplateLinkError
+}
+
+var unableToGetTemplateError = &microerror.Error{
+	Kind: "unableToGetTemplate",
+}
+
+func IsUnableToGetTemplateError(err error) bool {
```
kopiczko: Since we don't use `IsNilTemplateLinkError` & `IsNilTemplateLinkError` we should replace those errors with `executionFailedError`. https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors

service/controller/v12/key/key.go
whites11: Used 3 times. 
create.go lines 59 and 119 and 704

service/controller/v12/resource/instance/error.go
whites11: I am actually using those in unit tests to check that the right error type is returned.
Is it acceptable or should I replace them with the generic error as you suggested?

service/controller/v12/resource/instance/azure.go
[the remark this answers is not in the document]
whites11: Both suggestions make a lot of sense to me.
Given the advice given by Marcel, and the fact this function was already like this before, I'd leave that refactoring for a later stage if that's fine to you.

whites11:
> I have a feeling that there are many code fragments copied from other resources instead of being reused. Am I wrong?

I don't think that's the case. Not by my hands at least.
I reverted the refactoring in order to make this PR easier to follow.

service/controller/v12/resource/instance/azure.go
whites11: The refactoring is reverted now, closing this conversation

service/controller/v12/resource/instance/create.go
```diff
@@ -77,14 +74,48 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			if deploymentTemplateChk != "" {
+				err = r.setResourceStatus(customObject, DeploymentTemplateChecksum, deploymentTemplateChk)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("set %s to '%s'", DeploymentTemplateChecksum, deploymentTemplateChk))
+			} else {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Unable to get a valid Checksum for %s", DeploymentTemplateChecksum))
+				// todo remove any DeploymentTemplateChecksum leftovers from the CR
```
tuommaki: What's this? Is this tracked somewhere?

FYI: https://github.com/giantswarm/fmt/blob/master/go/todo_comments.md

service/controller/v12/resource/instance/create.go
```diff
@@ -77,14 +74,48 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			if deploymentTemplateChk != "" {
+				err = r.setResourceStatus(customObject, DeploymentParametersChecksum, deploymentParametersChk)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("set %s to '%s'", DeploymentParametersChecksum, deploymentParametersChk))
+			} else {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Unable to get a valid Checksum for %s", DeploymentParametersChecksum))
+				// todo remove any DeploymentParametersChecksum leftovers from the CR
```
tuommaki: Same here as above :slightly_smiling_face:

service/controller/v12/resource/instance/create.go
```diff
@@ -667,3 +695,90 @@ func isNodeDrained(drainerConfigs []corev1alpha1.DrainerConfig, instanceName str
...
 }
+
+func (r *Resource) handleDeploymentCompletedStatus(ctx context.Context, customObject providerv1alpha1.AzureConfig) error {
+	deploymentsClient, err := r.getDeploymentsClient(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	d, err := deploymentsClient.Get(ctx, key.ClusterID(customObject), key.VmssDeploymentName)
+	if IsDeploymentNotFound(err) {
```
tuommaki: We usually try to keep the reconciliation control flow at resource's top level. Here I'd propose to just mask the error and return it and then specifically handle the error in `EnsureCreated()`.

service/controller/v12/resource/instance/create.go
```diff
@@ -131,6 +162,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
+	if hasResourceStatus(customObject, Stage, DeploymentCompleted) {
+		return r.handleDeploymentCompletedStatus(ctx, customObject)
```
tuommaki: This here is missing a stack trace entry due to missing error masking. Later down I wrote a comment that suggests to actually lift the error handling to this level so that the resource control flow is kept at top level.

service/controller/v12/resource/instance/create.go
```diff
@@ -667,3 +695,90 @@ func isNodeDrained(drainerConfigs []corev1alpha1.DrainerConfig, instanceName str
...
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	s := *d.Properties.ProvisioningState
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", s))
+
+	if key.IsSucceededProvisioningState(s) {
+		computedDeployment, err := r.newDeployment(ctx, customObject, nil)
+		if controllercontext.IsInvalidContext(err) {
```
tuommaki: Similar to above, lift the error handling to top level so that control flow is also logged & managed there. Masking should be enough at this level.

service/controller/v12/resource/instance/create.go
whites11: this is completely harmless, I'll remove the todo

service/controller/v12/resource/instance/create.go
whites11: this is harmless, I'll remove the todo

service/controller/v12/resource/instance/checksum.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func getDeploymentTemplateChecksum(deployment resources.Deployment) (string, error) {
+
+	templateLink := deployment.Properties.TemplateLink
+	if templateLink == nil {
+		return "", nilTemplateLinkError
+	}
+
+	resp, err := http.Get(*templateLink.URI)
```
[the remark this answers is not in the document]
whites11: Not sure I got your point here. When would this case happen?
The checksum is calculated twice:
1) whenever sending the updated deployment
2) when checking if the template has changed

I don't see in which case we could avoid making the calculation.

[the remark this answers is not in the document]
tuommaki: There are few assumptions here so bear with me:

When the template already exists as deployment, I'd expect something like following if I have understood everything right:

*CR changes / upgrade changes parameters*:
1. Compute checksum for template and its parameters.
2. Compare checksum against the one in CR status.
3. If it doesn't match, update deployment. No need to check the deployment in Azure.

*CR doesn't change / reconcile Azure state*
1. Compute checksum for template and its parameters.
2. Compare checksum against the one in CR status.
3. It matches.
4. Invoke conditional GET with checksum against Azure to see if template or parameters in there are up to date.
5. If either of them is not up to date, `CreateOrUpdate` latest deployment.

Does this make sense? Or am I somehow misunderstanding something?

service/controller/v12/resource/instance/create.go
```diff
@@ -131,6 +160,28 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		} else if blobclient.IsBlobNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "ignition blob not found")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		}
+
+		return err
```
tuommaki:
```suggestion
		return nil
```

service/controller/v12/resource/instance/create.go
```diff
@@ -131,6 +160,28 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", "missing dispatched output values in controller context")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not ensure deployment")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		} else if blobclient.IsBlobNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "ignition blob not found")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		}
```
tuommaki:
```suggestion
		} else if err != nil {
			return microerror.Mask(err)
		}
```

service/controller/v12/resource/instance/create.go
```diff
@@ -131,6 +160,28 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
```
tuommaki: Is there possibly some useful log message that could be written here? I think with general aligning, we would in this case at least log `canceling resource` since resource execution prematurely ends here (it doesn't follow all the way to end). Does this make sense?

service/controller/v12/resource/instance/create.go
```diff
@@ -256,17 +307,20 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 			}
 
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("removed resource status '%s/%s'", Stage, InstancesUpgrading))
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("set resource status to '%s/%s'", Stage, DeploymentCompleted))
+
 			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
-			return nil
 		}
+
+		return nil
```
tuommaki:
```suggestion
```
I think this is not needed. The next line seems to close this scope and after that is `return nil`.

whites11: right

service/controller/v12/resource/instance/create.go
whites11: I think that log happens only when `reconciliationcanceledcontext.SetCanceled(ctx)` is called which is not the case in this `if` branch.

whites11: I put a generic error log just to be more verbose. That should `if` branch should be an edge case anyway.

service/controller/v12/resource/instance/checksum.go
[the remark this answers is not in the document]
whites11: It would make sense in theory, but you are misunderstanding one thing I guess.
The checksum is never made towards the Azure deployment. that's simply not possible.
When the azure deployment is updated, the checksum is calculated and stored in the CR. From that moment on, we assume that what's in ARM has the checksum that we saved in the CR.
If for any reason the azure deployment changes because of an external factor, the checksum stored in the CR would be simply wrong.

The HTTP get call we make is towards github to download the actual ARM template file content:

> https://raw.githubusercontent.com/giantswarm/azure-operator/%s/service/controller/v12/resource/%s/template/%s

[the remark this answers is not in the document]
tuommaki: Ah, right. Makes sense. All good here.

service/controller/v12/resource/instance/create.go
```diff
@@ -77,14 +74,46 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+				err = r.setResourceStatus(customObject, DeploymentParametersChecksum, deploymentParametersChk)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("set %s to '%s'", DeploymentParametersChecksum, deploymentParametersChk))
+			} else {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Unable to get a valid Checksum for %s", DeploymentParametersChecksum))
+			}
+
```
tuommaki:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
```

PR #601, opened 2019-11-20
Send Partner Program GUID in the User-Agent when interacting with Azure API

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/7475

According to [Azure docs](https://docs.microsoft.com/en-us/azure/marketplace/azure-partner-customer-usage-attribution#tag-a-deployment-with-the-resource-manager-apis)
> To enable customer usage attribution, when you design your API calls, include a GUID in the user agent header in the request. Add the GUID for each offer or SKU. Format the string with the pid- prefix and include the partner-generated GUID. 

This PR reads the partner GUID from the same secret that we are using to fetch the Subscription ID, and passes it in the User-Agent to the calls done by the operator.

Customers wouldn't be able to upgrade to the release containing these changes without adding the Partner Program ID to the `Secret` in Kubernetes.


client/azure_client_set.go
```diff
@@ -89,89 +92,103 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
 func newDeploymentsClient(config *clientConfig) *resources.DeploymentsClient {
 	c := resources.NewDeploymentsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+	_ = c.AddToUserAgent(config.partnerIdUserAgent)
```
kopiczko: What does this thing return?

fiunchinho: error when the passed parameter is empty, leaving the user-agent unchanged

kopiczko: Then I would prefer to panic here.

fiunchinho: What do you think about checking if config.PartnerID contains something before creating `clientConfig` and panicking on that case? Otherwise I need to get the error returned by `AddToUserAgent` on every function that creates a client. It's more explicity, but also more repetitive.

fiunchinho: I pushed the explicity/repetitive version. Still, let me know what you think about failing earlier if `config.PartnerID` is not set.

client/azure_client_set.go
```diff
@@ -63,117 +65,185 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
 
 	c := &clientConfig{
 		subscriptionID:          config.SubscriptionID,
+		partnerIdUserAgent:      fmt.Sprintf("pid-%s", config.PartnerID),
```
MarcelMue: Can we have this in a `key.go`file? Or does that make sense?

fiunchinho: I'm not sure what you mean, the partner id is not part of the `AzureConfig` CR. To me, it made sense to treat this value the same way we treat the subscription id. WDYT?

kopiczko: Not used outside the file so I'd say we shouldn't move it to key.go.

client/azure_client_set.go
```diff
@@ -63,117 +65,185 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
...
 
-	return &c
+	return &c, err
 }
 
-func newVirtualMachineScaleSetsClient(config *clientConfig) *compute.VirtualMachineScaleSetsClient {
+func newVirtualMachineScaleSetsClient(config *clientConfig) (*compute.VirtualMachineScaleSetsClient, error) {
 	c := compute.NewVirtualMachineScaleSetsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+	err := c.AddToUserAgent(config.partnerIdUserAgent)
```
kopiczko: error masking is missing.

fiunchinho: IIUC it's masked on line 119

client/azure_client_set.go
```diff
@@ -63,117 +65,185 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
...
 }
 
-func newDeploymentsClient(config *clientConfig) *resources.DeploymentsClient {
+func newDeploymentsClient(config *clientConfig) (*resources.DeploymentsClient, error) {
 	c := resources.NewDeploymentsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+	err := c.AddToUserAgent(config.partnerIdUserAgent)
 
-	return &c
+	return &c, err
```
tuommaki:
```suggestion
	return &c, microerror.Mask(err)
```
As Pawel wrote below, errors should be masked / wrapped. See https://github.com/giantswarm/fmt/blob/master/go/errors.md#wrapped-errors

This is required in every function to produce entry in stack trace.

client/azure_client_set.go
fiunchinho: Ok, I see what you mean. Fixed now.

client/azure_client_set.go
```diff
@@ -63,117 +65,185 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
...
 	c := &clientConfig{
 		subscriptionID:          config.SubscriptionID,
+		partnerIdUserAgent:      fmt.Sprintf("pid-%s", config.PartnerID),
 		resourceManagerEndpoint: env.ResourceManagerEndpoint,
 		servicePrincipalToken:   servicePrincipalToken,
 	}
 
+	deploymentsClient, err := newDeploymentsClient(c)
+	if err != nil {
+		return nil, err
```
kopiczko: It should be like that everywhere:

```suggestion
		return nil, microerror.Mask(err)
```

fiunchinho: OK, I'm slow today 😭 Fixed now.

client/azure_client_set.go
```diff
@@ -63,117 +65,185 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
...
 }
 
-func newDeploymentsClient(config *clientConfig) *resources.DeploymentsClient {
+func newDeploymentsClient(config *clientConfig) (*resources.DeploymentsClient, error) {
 	c := resources.NewDeploymentsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
+	err := c.AddToUserAgent(config.partnerIdUserAgent)
 
-	return &c
+	return &c, microerror.Mask(err)
```
kopiczko: There is a convention to return masked error only when it's not `nil`. One reason for that is better diffs if more erroring calls are added.

```suggestion
	if err != nil {
		return nil, microerror.Mask(err)
	}
	
	return &c, nil
```

fiunchinho: PTALA 😉

kopiczko: Already approved LGTM 👍

service/credential/credential.go
```diff
@@ -42,11 +44,21 @@ func GetAzureConfig(k8sClient kubernetes.Interface, name string, namespace strin
...
 	}
 
+	partnerID, err := valueFromSecret(credential, PartnerIDKey)
+	if err != nil {
+		// No having partnerID in the secret means that customer has not
+		// upgraded yet to use the Azure Partner Program. In that case we set a
+		// constant random generated GUID that we haven't registered with Azure.
+		// When all customers have migrated, we should error out instead.
+		partnerID = defaultAzureGUID
+	}
```
fiunchinho: Instead of erroring out when the value is not present, I'm using a default GUID that's not registered with Azure. That way this is not a breaking change anymore, and customers can start using these changes on the Azure operator even before getting a valid GUID registered on the Azure Partners Program.

tuommaki: Just to be sure: When GUIDs actually have the group part of the ID, does this match with valid ones? At least this needs to be tested so that Azure doesn't flip with it :slightly_smiling_face:

fiunchinho:
I'm testing the changes with a valid GUID that @MartaD registered on the Azure Partner Portal, and I won't call it success yet, but the GUID has showed up here

image

I'm not merging this PR until Azure support tells @MartaD what's the best way to make sure this is working as expected.

PR #602, opened 2019-11-22
Fixed nil pointer error

whites11:
I discovered a missing nil-check in the code that caused a panic for new clusters.

Fixed just that.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -359,11 +359,11 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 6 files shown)


fiunchinho:
Would it be possible to add a unit test that reproduces the error somehow? When do both `ws` and `err` are nil?

whites11:
> Would it be possible to add a unit test that reproduces the error somehow? When do both `ws` and `err` are nil?

You are right, there are better ways of fixing this. Let me work on it a little bit.

PR #604, opened 2019-11-25
Improve Ingress LB PIP name

tuommaki:
Original name for ingress LB public IP was just a "dummy-pip". Now that k8s
provider integration has actually support for re-using that on services with
type `LoadBalancer`, it makes sense to use more descriptive name for that.

In order to manage the migration of this name, store it in CR status. Older
clusters, which are migrated into this version, should still use the old name
as that cannot be changed, but newer ones must use the more descriptive one.

Towards https://github.com/giantswarm/giantswarm/issues/6636


service/controller/v12/key/key.go
```diff
@@ -217,6 +218,14 @@ func DNSZones(customObject providerv1alpha1.AzureConfig) providerv1alpha1.AzureC
 	return customObject.Spec.Azure.DNSZones
 }
 
+func DefaultIngressPIPName(customObject providerv1alpha1.AzureConfig) string {
+	return fmt.Sprintf("%s-Ingress-PublicLoadBalancer-PublicIP", ClusterID(customObject))
```
tuommaki: This is the desired name for PIP now.

service/controller/v12/resource/deployment/template/kubernetes_load_balancer_setup.json
```diff
@@ -11,13 +11,14 @@
...
-    "loadBalancerSkuName": "Standard"
+    "loadBalancerSkuName": "Standard",
+    "loadBalancerPIPName": "[parameters('ingressPIPName')]"
   },
   "resources":[
     {
       "apiVersion":"2017-10-01",
       "type":"Microsoft.Network/publicIPAddresses",
-      "name":"dummy-pip",
+      "name":"[variables('loadBalancerPIPName')]",
```
tuommaki: Use the name from parameter.

service/controller/v12/resource/deployment/deployment.go
```diff
@@ -24,6 +24,7 @@ func (r Resource) newDeployment(ctx context.Context, customObject providerv1alph
 		"clusterID":               key.ClusterID(customObject),
 		"dnsZones":                key.DNSZones(customObject),
 		"hostClusterCidr":         r.azure.HostCluster.CIDR,
+		"ingressPIPName":          key.IngressPIPName(customObject),
```
tuommaki: Pick up PIP name from CR status.

service/controller/v12/resource_set.go
```diff
@@ -353,6 +367,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		containerURLResource,
 		encryptionkeyResource,
 		blobObjectResource,
+		ingressPIPNameResource,
```
tuommaki: `deploymentResource` creates the given public IP so this resource must be run right before it.

service/controller/v12/resource/ingresspipname/resource.go
```diff
@@ -0,0 +1,141 @@
...
+
+func New(config Config) (*Resource, error) {
+	if config.G8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.G8sClient must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	newResource := &Resource{
```
xh3b4sd:
```suggestion
	r := &Resource{
```

service/controller/v12/resource/ingresspipname/resource.go
```diff
@@ -0,0 +1,141 @@
...
+		g8sClient: config.G8sClient,
+		logger:    config.Logger,
+	}
+
+	return newResource, nil
```
xh3b4sd:
```suggestion
	return r, nil
```

service/controller/v12/resource/ingresspipname/resource.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "finding if ingress PIP name must be set to status")
```
xh3b4sd: I would drop this.

service/controller/v12/resource/ingresspipname/resource.go
```diff
@@ -0,0 +1,141 @@
...
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "updated CR status with ingress PIP name")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+		reconciliationcanceledcontext.SetCanceled(ctx)
+	}
+
+	return nil
+}
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
xh3b4sd: I would like to have `create.go`, `delete.go` and `resource.go`.

service/controller/v12/key/key.go
kopiczko: Does it have to prefixed with cluster ID? Isn't it enough to put in the cluster RG?

nprokopic: Technically, load balancer name should be unique on a [resource group scope](https://docs.microsoft.com/en-us/azure/architecture/best-practices/resource-naming#networking), but I would also consider that having a bit more explicit name can make our life easier when looking at the list of resources (Azure portal, CLI and similar). We are also prefixing other (resource group scoped) Azure resources with cluster ID, although it's not technically necessary.

kopiczko: IMO shorter IDs also make life easier. I personally don't look at the full list of resources. My 2c. Not blocker.

nprokopic: > IMO shorter IDs also make life easier.

Good point, couldn't disagree.

In general, prefixed resource names or not, my (non-blocking) vote goes for consistency.

tuommaki: I followed the pattern of other two LBs name here: https://github.com/giantswarm/azure-operator/blob/master/service/controller/v12/resource/deployment/template/public_load_balancer_setup.json#L36

kopiczko: Don't sped too much time on this 👍

fiunchinho:
What are we missing to be able to merge this?

service/controller/v13/resource/deployment/template/kubernetes_load_balancer_setup.json
```diff
@@ -11,13 +11,14 @@
     }
   },
   "variables":{
-    "loadBalancerSkuName": "Standard"
+    "loadBalancerSkuName": "Standard",
+    "loadBalancerPIPName": "[parameters('ingressPIPName')]"
```
fiunchinho: Can't we directly use the parameter instead of using a variable?

tuommaki: Do you mean dropping this and changing the line below to
`        "[concat('Microsoft.Network/publicIPAddresses/', parameters('ingressPIPName'))]"` :question:

fiunchinho: Yes

PR #603, opened 2019-11-22
Vendor dependencies, update k8s to 1.16.3

sslavic:
Towards giantswarm/giantswarm#7675

PR #605, opened 2019-11-26
Use apps/v1 for deployment

fiunchinho:
Updating the `Deployment` object to a more recent `apiVersion`, since it could lead to problems, like those fixed here https://github.com/giantswarm/aws-operator/pull/2077.

### DIFF
--- a/helm/azure-operator-chart/templates/deployment.yaml
+++ b/helm/azure-operator-chart/templates/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: azure-operator
@@ -10,6 +10,9 @@ metadata:
 spec:
   replicas: 1
   revisionHistoryLimit: 3
+  selector:
+    matchLabels:
+      app: azure-operator
   strategy:
     type: RollingUpdate
   template:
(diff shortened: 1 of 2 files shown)


helm/azure-operator-chart/templates/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
whites11: What version of k8s is the apps/v1 started being supported? Are we sure all azure installations are at least that version?

sslavic:
I covered this in https://github.com/giantswarm/azure-operator/pull/603 but will rebase.

helm/azure-operator-chart/templates/deployment.yaml
rossf7: It was added in 1.9 so we shouldn't have problems. We need to switch to `apps/v1` as `extensions/v1beta1` is removed in 1.16. 

https://github.com/kubernetes/kubernetes/issues/61430

fiunchinho: It started on `1.9`, and it's required for `1.16`.

sslavic: apps was GA since 1.9, it's safe to switch, latest we support is 1.13
https://kubernetes.io/blog/2017/12/kubernetes-19-workloads-expanded-ecosystem/

fiunchinho: `1.14` is the oldest version I could find in our installations.

PR #608, opened 2019-11-28
Allow user to select which AZ's to use for tenant clusters

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/7741

[Following our work](https://github.com/giantswarm/giantswarm/issues/7745#issuecomment-558590053), we need some changes on our Azure operator to support multi-AZ.

These changes depend on [a new version](https://github.com/giantswarm/apiextensions/pull/300) of the `AzureConfig` CRD which contains a property with the desired Availability Zones where the tenant cluster will be deployed. The operator will read the property value and apply it to the ARM template for both the master template and workers template.

[We also need to change the VPN Gateway](https://docs.microsoft.com/en-us/azure/vpn-gateway/about-zone-redundant-vnet-gateways) and the Public IP Address to support multi-AZ. 
> Migrating your existing virtual network gateways to zone-redundant or zonal gateways is currently not supported. You can, however, delete your existing gateway and re-create a zone-redundant or zonal gateway.

**[This is a breaking change](https://docs.microsoft.com/en-us/azure/vpn-gateway/about-zone-redundant-vnet-gateways#can-i-changemigrateupgrade-my-existing-virtual-network-gateways-to-zone-redundant-or-zonal-gateways) for our customers**.


service/controller/v12/key/key.go
```diff
@@ -300,6 +300,11 @@ func RouteTableName(customObject providerv1alpha1.AzureConfig) string {
 	return fmt.Sprintf("%s-%s", ClusterID(customObject), routeTableSuffix)
 }
 
+// AvailabilityZones returns the availability zones where the cluster will be created.
+func AvailabilityZones(customObject providerv1alpha1.AzureConfig) []int {
+	return customObject.Spec.Azure.AvailabilityZones
```
whites11: Might the key be missing from the CR? What happens in that case?

service/controller/v12/resource/instance/template/main.json
```diff
@@ -62,6 +62,13 @@
       "metadata":{
         "description":"Output value of the worker subnet ID as referenced from the virtual network setup."
       }
+    },
+    "zones":{
+      "type":"array",
+      "defaultValue": [1],
```
nprokopic: Does this mean that by default we would use multi-AZ and then set it to use a single zone (zone 1) by default?
If so, wouldn't we want to completely skip multi-AZ parameters if the customer didn't specify them (so we keep existing/old behavior)?

service/controller/v12/resource/vpn/template/main.json
```diff
@@ -38,7 +38,7 @@
         "provider":"[toUpper(parameters('GiantSwarmTags').provider)]"
       },
       "sku":{
-        "name": "Basic"
+        "name": "Standard"
```
nprokopic: See comment above for the default behavior (when multi-AZ parameters are not specified at all).
It would be good to keep using `Basic` load balancer by default, when multi-AZ params are not specified - one reason would be because `Basic` LB is free, while `Standard` is paid, so we could try to keep the costs down (unless we have other reasons to use `Standard` LB).

service/controller/v12/resource/instance/template/vmss.json
```diff
@@ -62,6 +62,13 @@
     "vmssUpgradePolicy":{
       "type":"string",
       "defaultValue":"Manual"
+    },
+    "zones":{
+      "type":"array",
+      "defaultValue": [1],
```
nprokopic: Same comment regarding the default values as for `main.json` above.

nprokopic:
Additionally, if we manage to keep the existing behavior (and therefore keep using the same Azure resources) when customers do not want multi-AZ, this change wouldn't be breaking anymore. Admittedly this would add certain degree of complexity/overhead (not much IMHO).

service/controller/v12/key/key.go
fiunchinho: It will break on that case I think.

I'm not sure I understand how versioning works. Since we didn't flatten our operator, our `azure-operator` reconciliates different versions of the `AzureConfig` CRD, depending on the `VersionBundle` field. So if my understanding is correct, old CR's without the `AvailabilityZones` field will be handled by old versions of the `azure-operator` , while new CR's will be handled by this new version. That way nothing breaks. Does that make sense?

tuommaki: There's actually more to it:
* When something with latest `apiextensions` updates the CR, the field gets added with empty value as it's not `omitempty`: https://github.com/giantswarm/apiextensions/pull/300/files#diff-f8ea2fc1fb00586a596f9bc54c5ebbb5R68
* No matter whether or not the field exists in CR, the data structure here is not using pointers so it's just empty list.
* In general Go handles nil slices gracefully so empty value should work.
* With the release that introduces multi-AZ, it must have migration code that updates CR with present AZ configuration.

service/controller/v12/resource/instance/template/main.json
tuommaki: For me it's ok to have default value in template, but we should ensure that even with single zone request, we randomize the tenant cluster specific AZ on creation. This brings little bit more resilience for customer so that not every single tenant cluster goes down when single AZ goes down on specific region.

service/controller/v12/resource/vpn/template/main.json
tuommaki: I'm curious how is it `Basic` still here? I was under impression we moved to `Standard` before due to various networking issues with `Basic` one. @T-Kukawka do you have more insight in here?

service/controller/v12/resource/instance/template/main.json
fiunchinho: I don't think we can randomize it. If we allow customers to select specific AZ's and they choose one single AZ, it would be a surprise to them to find the cluster using a different AZ because it was randomized it.

fiunchinho:
Just confirmed that the AZ's can [be changed only at creation time](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#create-time-properties).

service/controller/v12/resource/instance/template/main.json
tuommaki: Sorry about mixing up different things. This here is fine. On principal operator must not change a value set in CR. In some cases if user did not specify AZ, then it's usually randomly selected in `api` (i.e. before writing that into CR), but in case the end user always picks up an AZ, then there's no randomization at all. Nevertheless, for me this is fine.

fiunchinho:
I've added a new template that passes the AZ's parameter to the VMSS. This is duplicating the same template we already had. A better solution would be to use go templating and generate the templates in code. But since we are going to rewrite most of this when adopting the ClusterAPI, I'm not sure it is worth it. 

Right now it will use the AZ template or the regular template depending on whether or not the CR has the `zones` property set to nil. What do you think @tuommaki ?

tuommaki:
> Right now it will use the AZ template or the regular template depending on whether or not the CR has the zones property set to nil. What do you think @tuommaki ?

That seems good approach now that it's there. Can you create a [TODO](https://github.com/giantswarm/giantswarm/issues/new?template=todo.md) issue to track this so that we can actually remove the non-AZ template once all clusters have AZ specified (in case we need to live longer with `legacy`)?

fiunchinho:
> That seems good approach now that it's there. Can you create a [TODO](https://github.com/giantswarm/giantswarm/issues/new?template=todo.md) issue to track this so that we can actually remove the non-AZ template once all clusters have AZ specified (in case we need to live longer with `legacy`)?

Done

fiunchinho:
Tested with both CR's with and without AZ's set.

service/controller/v12/resource/instance/deployment.go
```diff
@@ -72,6 +72,24 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
...
+	// templates to create the ARM template and conditionally add the AZ
+	// parameters only when needed.
+	//
+	//     https://github.com/giantswarm/giantswarm/issues/7909
+	//
+	var vmssTemplateFile string
+	{
+		// Setting the AZ parameter on a template that doesn't already have
+		// it would fail, so we need to use a different ARM template.
+		if len(zones) > 0 {
```
fiunchinho: I'm using `len` so it works with both an empty slice and nil.

fiunchinho:
Merging and...

PR #610, opened 2019-12-05
Vendor operatorkit

tomahawk28:
See giantswarm/giantswarm#7963.


Gopkg.toml
```diff
@@ -88,9 +88,9 @@ required = [
   name = "github.com/giantswarm/micrologger"
   branch = "master"
 
-[[constraint]]
+[[override]]
+  branch = "cluster-controller"
```
rossf7: We don't want to merge this PR yet. So if we leave this on master we should be able to merge these changes now. 

Otherwise we can leave this open until https://github.com/giantswarm/operatorkit/pull/338 is merged.

tomahawk28: @rossf7 I think most PRs for the checklist have this change. For now, I will just test on testing installation and waiting for prior PR merged on operatorkit. WDYT?

tomahawk28: Nevermind, I will revert this one and would make this PR as ready.

Gopkg.toml
```diff
@@ -140,6 +140,11 @@ required = [
   name = "github.com/russross/blackfriday"
   version = "v1.5.2"
 
+[[override]]
+  source = "gopkg.in/fsnotify/fsnotify.v1"
+  name = "gopkg.in/fsnotify.v1"
```
rossf7: Should be able to remove this too.

Gopkg.toml
rossf7: Thanks I think that is easier. Better to be able to merge sooner.

tomahawk28:
Tested on `ghost`

tomahawk28:
8.4.1, 9.0.0 release tested

PR #613, opened 2019-12-10
Bump coreos version for 10 patch1 and 10 patch2 releases

whites11:
to make it compatible with westgermany azure zone.

See https://github.com/giantswarm/giantswarm/issues/8037


fiunchinho:
Can we close this?

PR #614, opened 2019-12-10
Read AZs from env and use it when creating tenant cluster

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/7744

It depends on https://github.com/giantswarm/apiextensions/pull/304 in order to build successfully.


integration/env/azure.go
```diff
@@ -94,6 +102,16 @@ func init() {
 	}
 }
 
+func AzureAvailabilityZones() []int {
+	azs := strings.Split(strings.TrimSpace(azureAvailabilityZones), " ")
+	zones := make([]int, len(azs))
+
+	for i, s := range azs {
+		zones[i], _ = strconv.Atoi(s)
```
xh3b4sd: We should never ever hide errors. When you think the error case is unlikely to ever happen for whatever reasons and you do not want to add the error to the interface then at least panic here. Next year something changes that nobody is aware of anymore and we waste a lot of time and man power because we were too lazy to invest 10 seconds to handle an obvious error case.

fiunchinho: Fixed, please take a look.

fiunchinho:
Superseeded by https://github.com/giantswarm/azure-operator/pull/603

PR #618, opened 2020-01-08
Compute deployment template digest by directly reading response body

tuommaki:
By use of `io.Copy(Writer, Reader)` this change skips making of full copy of ARM
template in temporary variable before computing hash.


tuommaki:
Tested on `godsmack`: This still computes the same checksum.

tuommaki:
> That thing looks like a unit test would be grand.

Indeed and there is: https://github.com/giantswarm/azure-operator/blob/master/service/controller/v12/resource/instance/checksum_test.go#L25-L31

PR #619, opened 2020-01-08
Double the size of VMSS before upgrading worker instances

tuommaki:
This is first part of changes towards more graceful upgrades in Azure.

General direction here is following:
  - Double the size of Worker VMSS.
  - Cordon old nodes.
  - Drain old nodes one-by-one.
  - Terminate old nodes.
  - Scale Worker VMSS down.

Towards https://github.com/giantswarm/giantswarm/issues/7735

### DIFF
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -338,10 +338,22 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		if !masterUpgradeInProgress && !workerUpgradeInProgess {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "neither masters nor workers upgraded")
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("scaling %s VMSS down to desired number of nodes %d", key.WorkerVMSSName(customObject), key.WorkerCount(customObject)))
 
+			// Scale down to the desired number of nodes in worker VMSS.
+			nodeCountFunc := func(_ int64) int64 {
+				return int64(key.WorkerCount(customObject))
+			}
+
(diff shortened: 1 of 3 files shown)


service/controller/v12/resource/instance/create.go
```diff
@@ -187,9 +188,36 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+	if hasResourceStatus(customObject, Stage, ScaleVMSSToDouble) {
+		// Double the desired number of nodes in worker VMSS in order to
+		// provide 1:1 mapping between new up-to-date nodes when draining and
+		// terminating old nodes.
+		nodeCountFunc := func(_ int64) int64 {
+			return int64(key.WorkerCount(customObject) * 2)
+		}
+
+		err := r.scaleVMSS(ctx, customObject, key.WorkerVMSSName, nodeCountFunc)
```
whites11: I don't get why you are passing a function instead of the computed int64 here.

tuommaki: My original implementation was following:
```go
nodeCountFunc := func(n int64) int64 {
    return n * 2
}
```

...but then I realized that in case the operator crashes during `InstancesUpgrading`, it would in the worst case double the size of VMSS on every iteration. Then I settled with the const value there, but I felt that the function signature is more flexible with the function. That way implementation can avoid reading of `VMSS` state several times during operation.

whites11: ok

whites11:
> > It would be nice to keep the flowchart in docs/instances-states-v12.svg up to date.
> 
> Yes, this is my intention. Did you originally create it in Google Docs? Can you share the link to that so I can modify it?

I used an app on gdrive to get the first chart done. But the app was freemium and I hit the limit so I ended up exporting the SVG and editing it manually with Inkscape.

tuommaki:
This got stale. Dropping.

PR #620, opened 2020-01-13
Allow passing of kubectl config file path instead of restconfig params

tuommaki:
In order to enable easier local development against real installation, allow
passing kubectl's config file path to operator for k8s connection
configuration.

Note: CI is b0rked while waiting for merge for relevant flag field: https://github.com/giantswarm/operatorkit/pull/354


tuommaki:
With this one may run `azure-operator` locally:
`./azure-operator daemon --service.kubernetes.incluster=false --config.files <configfile.yaml> --service.azure.template.uri.version <foobarcommitsha>`

service/service.go
```diff
@@ -100,36 +100,30 @@ func New(config Config) (*Service, error) {
...
-			RestConfig: restConfig,
+			Logger: config.Logger,
+		}
+
+		// Prefer KubeConfigPath if given, otherwise build restconfig.
+		{
+			inCluster := config.Viper.GetBool(config.Flag.Service.Kubernetes.InCluster)
+			kubeConfigPath := config.Viper.GetString(config.Flag.Service.Kubernetes.KubeConfigPath)
+			if !inCluster && kubeConfigPath != "" {
+				c.KubeConfigPath = kubeConfigPath
```
kopiczko: I'd rather fail when `inCluster && kubeConfigPath != ""`

main.go
```diff
@@ -152,3 +155,12 @@ func mainError() error {
...
 }
+
+func homeDir() string {
+	home := os.Getenv("HOME")
+	if home == "" {
+		// Windows homedir.
+		home = os.Getenv("USERPROFILE")
+	}
+	return home
+}
```
fiunchinho: I believe [go 1.12 introduced a new function](https://golang.org/doc/go1.12#os) called `UserHomeDir`. Would that work here?

service/service.go
tuommaki: Both are needed. One for driving logic and other one for validation.

main.go
tuommaki: Good shout, but after all, I had to drop that default for correct behaviour `inCluster`. :(

tuommaki:
Fishing for feedback

service/service.go
kopiczko: Why would you need `kubeConfigPath` if `inCluster` is true?

tuommaki: No, you don't. By "both" I meant that both cases of logic are needed plus the `else` case for rest config. The validation is in the merged code.

kopiczko: This is going wrong way :) Let me elaborate on my initial thought.

> I'd rather fail when `inCluster && kubeConfigPath != ""`

I mean if the user provides both `--incluster=true` and `--kubeconfigpath="non-empty"` this may be a misconfiguration / done by mistake. Someone may spin circles figuring why `--kubeconfigpath` flag isn't working not noticing that `--incluster=true` was specified. To prevent such situation I'd like to force the user to specify one or another by returning error on `inCluster && kubeConfigPath != ""` condition.

tuommaki: Isn't that what my final code does?
`master`: https://github.com/giantswarm/azure-operator/blob/master/service/service.go#L116-L126

```go
			if !inCluster && kubeConfigPath != "" {
				c.KubeConfigPath = kubeConfigPath
			} else if inCluster && kubeConfigPath != "" {
				return nil, microerror.Maskf(invalidConfigError, "inCluster and kubeConfigPath must not be defined at the same time")
			} else {
				restConfig, err := buildK8sRestConfig(config)
				if err != nil {
				}
				c.RestConfig = restConfig
			}
```

kopiczko: Ah cool sorry. I didn't notice that. This is a bit misleading. I would assume this should be the first check because then you don't have to check `!inCluster && kubeConfigPath != ""`. `kubeConfigPath != ""` is sufficient. But behaviour-wise this is good. Thanks!

PR #621, opened 2020-01-14
Update calico to 3.10.1

MarcelMue:


tfussell:
Comparing with https://github.com/giantswarm/k8scloudconfig/pull/615/files, might also need to change priority class of `calico-node` and maybe `typha` to `cluster-node-critical`.

MarcelMue:
> Comparing with https://github.com/giantswarm/k8scloudconfig/pull/615/files, might also need to change priority class of `calico-node` and maybe `typha` to `cluster-node-critical`.

You are right, done.

MarcelMue:
Tested and works on `ghost`

PR #623, opened 2020-01-15
Added test to assert we are using the specified AZ

fiunchinho:


integration/test/multiaz/multiaz_test.go
```diff
@@ -0,0 +1,62 @@
...
+func (s *MultiAZ) Test(ctx context.Context) error {
+	s.logger.LogCtx(ctx, "level", "debug", "message", "getting current availability zones")
+	zones, err := s.provider.GetClusterAZs(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	s.logger.LogCtx(ctx, "level", "debug", "message", "found availability zones", "azs", zones)
+
+	if len(zones) != 1 {
+		return microerror.Mask(wrongAzsUsed)
```
tuommaki: Is this error correct? Should there be something else when AZ's aren't present?

fiunchinho: You mean adding two errors: one when wrong AZ is used, another one when more than one AZs are used?

tuommaki: Yes, or change the error to more generic and use `Maskf(genericError, "problem description")`

kopiczko: Yeah, let's please avoid creating specific errors for everything. The default should be `executionFailedError` + description. Unless you need a matcher. Details here https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors

fiunchinho: ok, I just pushed some changes to fix that.

PR #624, opened 2020-01-16
Set used AZs on Status field

fiunchinho:
It would be cool to add the AZ's being used to create the VMSS in the `Status` field of the `AzureConfig` object.

It feels weird to just copy whatever is in the `Spec` like it's not _real_. So I ended up issuing a request to Azure's API to fetch that info. I don't know if that's the most efficient way. Are there other ways of getting that info?

### DIFF
--- a/service/controller/v12/resource/instance/create.go
+++ b/service/controller/v12/resource/instance/create.go
@@ -3,6 +3,8 @@ package instance
 import (
 	"context"
 	"fmt"
+	"strconv"
+
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
 	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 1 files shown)


fiunchinho:
The question is if we need to put the used AZ's in the Status field, or if we can live without it.

fiunchinho:
You are suggesting creating a new resource that takes the `Status` field from the `AzureConfig` object and sends an HTTP request to the k8s API to update the Status on the API. What would happen when another resource cancels the reconciliation before this resource gets executed? Wouldn't the changes to the status field on the object be lost?

tuommaki:
> You are suggesting creating a new resource that takes the Status field from the AzureConfig object and sends an HTTP request to the k8s API to update the Status on the API. What would happen when another resource cancels the reconciliation before this resource gets executed? Wouldn't the changes to the status field on the object be lost?

This status resource can be prioritized to be executed as one of the first in the resource set.

- If a resource before this one cancels the reconciliation, it probably had a reason and that should be fine.
- If status already contains the AZ, we don't need to fetch it again since it's immutable.
- If status is read from Azure and written to CR status, I think it makes sense to cancel reconciliation at that point. The `UpdateStatus()` change to CR would trigger new reconciliation round immediately.

Does that make sense?

fiunchinho:
I think so. Should we move the checksum logic to this new status resource?

tuommaki:
> Should we move the checksum logic to this new status resource?

Yes, that would probably make a lot of sense. :+1:

nprokopic:
This got a bit outdated now, the change is in unflattened version. Porting it to flattened version or closing? :)

fiunchinho:
Closing for now

PR #630, opened 2020-01-21
Adding docs on how to debug e2etests

fiunchinho:


docs/debug_e2etests.md
```diff
@@ -0,0 +1,32 @@
...
+
+The job will be re executed, but this time there is a new step that let's you connect to the worker using SSH.
+
+![](ssh.png)
+
+## Inspecting control plane cluster
+Let's _ssh_ into the CircleCI worker to install `kubectl` so we can talk to our test cluster.
+Copy and run the command from this step:
+
+![](kubectl.png)
```
MarcelMue: It would be better to have this as a code block here - I don't want to write off that command from the picture or navigate there to copy it :P

docs/debug_e2etests.md
```diff
@@ -0,0 +1,32 @@
...
+
+The `kubeconfig` file needed to connect to the test cluster was created during the environment preparation.
+We can inspect the control plane cluster using `kubectl` and that `kubeconfig`:
+
+```bash
+kubectl --kubeconfig="/home/circleci/.kube/config" get nodes
```
kopiczko: Maybe it's easier to do `export KUBECONFIG="/home/circleci/.kube/config"` and then use kubectl w/o `--kubeconfig` flag.

docs/debug_e2etests.md
fiunchinho: Yes, but I didn't want to hard-write the kubectl _version_, since it would be different depending on the minikube version.

fiunchinho: I kept the image _and_ added the code block

MarcelMue: Nice.

PR #631, opened 2020-01-21
Docs on how to use helm and see logs for the operator while runing e2e tests

fiunchinho:


docs/debug_e2etests.md
```diff
@@ -36,3 +39,63 @@ For example:
...
+
+```bash
+helm get values <release-name>
+```
+
+
+## Inspecting resources on the cloud provider
+
+The operator will create resources on the cloud provider, like the tenant clusters. Sometimes it's useful to go to the provider console and see what's going on.
+On Azure we can find the resource groups being used to hold the tenant clusters.
```
MarcelMue: Missing here in which account they are - I know this should be obvious but there are multiple GS subscriptions.

docs/debug_e2etests.md
```diff
@@ -36,3 +39,63 @@ For example:
...
+
+Let's go to the Azure portal. On the resource group search box we should look for resource groups whose name starts with `ci-`.
+
+We should see a list of the resource groups for our e2e tests. We need to look for the resource group name on the CircleCI job logs to know which resource group is being used on each job.
```
MarcelMue: Would be nice to have an example log here to show where the name of the resource group appears. Not super important though.

docs/debug_e2etests.md
fiunchinho: What would be best here, to specify the subscription id?

MarcelMue: Maybe something like `accountsgiantswarm` subscription? Not sure how cool it is to put the subscription id into a public repo, but not sure :)

fiunchinho: Added name of the subscription

docs/debug_e2etests.md
fiunchinho: Added example

PR #632, opened 2020-01-22
Introduce state machine for instance resource

tuommaki:
Instance resource has been working with persisted state for a long time
already. The state is kept in CR's status condition and updated accordingly as
instances upgrade proceeds. This has been necessary because in Azure the VMSS
instances upgrade process requires multiple steps for each VM and it's not
automated in similar way as the ASG in AWS.

Now that there needs to be more steps added into this process in order to make
upgrade more graceful for workload, maintenance and reasoning of this resource
became difficult during development process.

Hence this change adds a concept of resource internal state machine that drives
the resource behavior when reconciling a cluster upgrade.

Motivation behind this design is to make multi-step nature of instances upgrade
a first class citizen while maintaining cohesion for azure-operator overall.
This also makes certain level of Azure API call optimization possible when it
otherwise could not be so straightforward.


service/controller/v13/resource/instance/create.go
```diff
@@ -33,754 +39,42 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	if err != nil {
 		return microerror.Mask(err)
 	}
-	deploymentsClient, err := r.getDeploymentsClient(ctx)
```
tuommaki: The diff for this file is really bad. All the removed code was moved into other files and only the driving code for state machine is now in `EnsureCreated()` function. Here it is after changes:
```go
func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
        customObject, err := key.ToCustomObject(obj)
        if err != nil {
                return microerror.Mask(err)
        }

        var newState state.State
        var currentState state.State
        {
                s, err := r.getResourceStatus(customObject, Stage)
                if err != nil {
                        return microerror.Mask(err)
                }
                currentState = state.State(s)

                if currentState == "" {
                        // DeploymentUninitialized is the initial state for instance resource.
                        newState = DeploymentUninitialized
                        r.logger.LogCtx(ctx, "level", "debug", "message", "no current state present")
                } else {
                        r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("current state: %s", currentState))
                        newState, err = r.stateMachine.Execute(ctx, obj, currentState)
                        if err != nil {
                                return microerror.Mask(err)
                        }
                }
        }

        if newState != currentState {
                r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("new state: %s", newState))
                r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting resource status to '%s/%s'", Stage, newState))
                err = r.setResourceStatus(customObject, Stage, string(newState))
                if err != nil {
                        return microerror.Mask(err)
                }
                r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("set resource status to '%s/%s'", Stage, newState))
                reconciliationcanceledcontext.SetCanceled(ctx)
        } else {
                r.logger.LogCtx(ctx, "level", "debug", "message", "no state change")
        }

        return nil
}
```

service/controller/v13/resource/instance/create.go
```diff
@@ -3,21 +3,27 @@ package instance
...
+	"github.com/giantswarm/azure-operator/service/controller/v13/resource/instance/internal/state"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/controller/context/reconciliationcanceledcontext"
-	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// newStateMachine configures and returns state machine that is driven by
```
xh3b4sd:
```suggestion
// configureStateMachine configures and returns state machine that is driven by
```

service/controller/v13/resource/instance/create.go
```diff
@@ -33,754 +39,42 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	s := *d.Properties.ProvisioningState
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", s))
-
-	if key.IsSucceededProvisioningState(s) {
-		computedDeployment, err := r.newDeployment(ctx, customObject, nil)
+	if newState != currentState {
```
xh3b4sd: Why not going with the terminology of current and desired state? What does `newState` mean?

service/controller/v13/resource/instance/create_deployment_completed.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+
+	"github.com/giantswarm/azure-operator/service/controller/v13/blobclient"
+	"github.com/giantswarm/azure-operator/service/controller/v13/key"
+	"github.com/giantswarm/azure-operator/service/controller/v13/resource/instance/internal/state"
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) deploymentCompletedTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	customObject, err := key.ToCustomObject(obj)
```
xh3b4sd: SOTA says it is `cr`.

service/controller/v13/resource/instance/create_deployment_completed.go
```diff
@@ -0,0 +1,83 @@
...
+				// As current and desired state differs, start process from the beginning.
+				return DeploymentUninitialized, nil
+			}
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", "template and parameters unchanged")
+
+			return currentState, nil
+		}
+	} else if key.IsFinalProvisioningState(s) {
+		// Deployment has failed. Restart from beginning.
```
xh3b4sd: The key function and the comment are a bit contradictory. Why final vs failed?

service/controller/v13/resource/instance/create_deployment_completed.go
```diff
@@ -0,0 +1,83 @@
...
+		return DeploymentUninitialized, nil
+	}
+
+	r.logger.LogCtx(ctx, "level", "warning", "message", "instances reconciliation process reached unexpected state")
+
+	// Normally the process should never get here. In case this happens, start
+	// from the beginning.
+	return DeploymentUninitialized, nil
```
xh3b4sd: Looks like we do not even need the else if above because we return `DeploymentUninitialized` here anyway.

service/controller/v13/resource/instance/create_deployment_completed.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func (r *Resource) deploymentCompletedTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return currentState, microerror.Mask(err)
+	}
+	deploymentsClient, err := r.getDeploymentsClient(ctx)
+	if err != nil {
+		return currentState, microerror.Mask(err)
```
xh3b4sd: Why is the current state dispatched here? Is it supposed to be mutated inside the step implementations of the state machine? I think it is an anti pattern to return anything else than the zero value in case of an returned error. Nobody should ever expect anything being returned from a function call than the error, in case an error is returned.

service/controller/v13/resource/instance/create_deployment_completed.go
tuommaki: I know, but just like many other things such as those log entries, I didn't want to change too much at once. There's definitely bunch of boyscout targets after this.

service/controller/v13/resource/instance/create_deployment_completed.go
tuommaki: I leaned towards this approach, because `State` for now is just plain `string` and any string is valid state in general and I didn't want to create some special reserved words for error cases. Go also lacks nice way to express `Option` type and I wanted to comply in all possible cases with restriction that `TransitionFunc` must return configured state.

service/controller/v13/resource/instance/create_deployment_completed.go
tuommaki: `FinalProvisioningState` refers to Azure deployment state which is either somehow running or pending or alternatively failed or successful. Latter ones are final in the sense that Azure won't take any further actions and the state cannot change.

At this point we know that it's failed because the earlier `if` didn't trigger as `SucceededProvisioningState`.

I found this state very complicated and while it's little bit better here, it's still very complicated. My current plan is to move the checksum check from here to first state which should then improve the situation quite a bit. Nevertheless, as earlier mentioned, I wanted to focus on the state machine conversion in this refactoring and do other changes separately.

service/controller/v13/resource/instance/create_deployment_completed.go
tuommaki: True. This should get simpler with above mentioned changes. :slightly_smiling_face:

service/controller/v13/resource/instance/create.go
tuommaki: Because this is about transitioning from one state into another within defined state machine which logic is driven by external side-effects. State machine kind of executes logic by transitioning from one state to another. It doesn't try to work towards some desired state. In other words one could maybe say that state machine itself never stops. It's never ready if that makes sense.

You may want to also check [Finite-state machine](https://en.wikipedia.org/wiki/Finite-state_machine).

tuommaki:
All e2e tests ran twice already green. Also my manual tests of cluster creation and upgrade into this version both succeeded without glitches.

service/controller/v13/resource/instance/create_deployment_completed.go
xh3b4sd: The state machine is not doing anything with the state in error cases when I got that right. So we should simply return an empty string here and make clear that the state in error cases has no effects whatsoever.

tuommaki: Ok. I changed that.

service/controller/v13/resource/instance/create.go
xh3b4sd: Disagree and commit.

service/controller/v13/resource/instance/create_deployment_uninitialized.go
```diff
@@ -0,0 +1,91 @@
...
+
+	computedDeployment, err := r.newDeployment(ctx, customObject, nil)
+	if controllercontext.IsInvalidContext(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "missing dispatched output values in controller context")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "did not ensure deployment")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return currentState, nil
+	} else if blobclient.IsBlobNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ignition blob not found")
+		resourcecanceledcontext.SetCanceled(ctx)
```
kopiczko: Sorry for the late review but I was busy with other stuff. I wanted to get to this eventually. And I like it. There is one but though. This isn't inline with this doc anymore: https://github.com/giantswarm/giantswarm/blob/895f0abcf3a97835c76d71d8967ef7c3a22a7e7b/sigs/operator/resource_control_flow.md. I'm fine with changing the doc accordingly.

tuommaki: Ack. I'll think about that and provide a PR for it. So far this approach is only used inside this single resource, but makes sense to document it in case it's useful somewhere else as well.

PR #633, opened 2020-01-23
E2e bastion

fiunchinho:
These changes would allow us to spin up an Azure VM on the bastion subnet and ssh into the tenant cluster nodes.

With this approach, the security group used for tenant cluster nodes will always allow SSH traffic coming from the bastion subnet, even if this subnet doesn't exist (another approach would be to allow all SSH traffic coming from the _whole_ virtual network, but the bastion VM still needs its own subnet).

This subnet is only created when running e2e tests. Not sure if this is the right approach: the virtual network during tests would be different than in production.

### DIFF
--- a/service/controller/v13/resource/deployment/template/bastion_e2e_vm.json
+++ b/service/controller/v13/resource/deployment/template/bastion_e2e_vm.json
@@ -1,0 +1,117 @@
+{
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "GiantSwarmTags":{
+      "type":"object",
+      "defaultValue":{
+        "provider":"F80D01C0-7AAC-4440-98F6-5061511962AD"
+      }
+    },
+    "subnetID": {
+      "type": "string"
+    },
+    "adminPublicKey": {
+      "type": "secureString"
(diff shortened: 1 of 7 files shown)


service/controller/v13/resource/deployment/template/bastion_e2e_vm.json
```diff
@@ -0,0 +1,117 @@
+{
```
fiunchinho: This specific template is not ready yet. You can review it when the PR is marked as `Ready`.

service/controller/v13/resource/deployment/template/bastion_e2e_vm.json
```diff
@@ -0,0 +1,117 @@
...
+            "createOption": "fromImage",
+            "managedDisk": {
+              "storageAccountType": "[parameters('osDiskType')]"
+            }
+          },
+          "imageReference": {
+            "publisher": "CoreOS",
+            "offer": "CoreOS",
+            "sku": "Stable",
+            "version": "2191.5.0"
```
whites11: Why are you using a core os version as old as the Dynosaurs? :joy:
Upgrade to `2135.6.0` maybe (a version known to be compatible with most azure regions).

fiunchinho:
Superseeded by https://github.com/giantswarm/azure-operator/pull/637

PR #634, opened 2020-01-24
Separate instance upgrade state between masters and workers

tuommaki:
As preparation for refactored worker upgrade process, separate masters and
workers upgrading as separate states. This allows more consistent refactoring
of workers upgrade process.


service/controller/v13/resource/instance/create_worker_instances_upgrading.go
```diff
@@ -0,0 +1,84 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) workerInstancesUpgradingTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	versionValue := map[string]string{}
```
fiunchinho: is this the kind of information that we could put in the context?

tuommaki: Theoretically, but when it's already available in `Status` of reconciled CR I feel that would be sort of duplicate information.

What kind of benefit do you see in that approach?

tuommaki:
E2E Green :muscle:

service/controller/v13/resource/instance/create_worker_instances_upgrading.go
fiunchinho: I believe I miss clicked. I was talking about the drainerConfigs. Instead of repeating the logic to fetch that information, maybe we could put it in the context, and just read it from there in here. That would avoid duplication, but I don't see it as a big problem as it currently is.

tuommaki: In `instance` resource's case I would prefer to not do that because those `DrainerConfigs` are only needed in two phases (`MasterInstancesUpgrading` and `WorkerInstancesUpgrading`) during upgrade and otherwise they aren't needed.

If they were needed during every reconciliation round, then that would be a good idea.

tuommaki:
That single refactoring branch kind of overtook everything. Dropping this here.

PR #638, opened 2020-01-25
Remove resource group when tests are finished

fiunchinho:
On Azure we have the concept of resource group that holds all the resources. When tearing down the tenant cluster created to run our e2e tests, we are currently deleting the `AzureConfig` object in the control plane, and waiting for it to be deleted.

Why don't we delete the resource group and forget about it? The tests on this PR took 15-20 minutes less than regular e2e tests executions. Is there a reason not to do it?

If we want to test that the cluster is deleted when deleting the `AzureConfig` there should be a specific test case for that, instead of doing it on all the tests.


fiunchinho:
Before this change

image


After this change

image

integration/setup/teardown.go
```diff
@@ -4,18 +4,15 @@ import (
...
 func Teardown(c Config) error {
 	ctx := context.Background()
 
-	// TODO the deletion detection should rather happen based on the cluster
-	// status or even based on the CR being gone.
-	//
-	//     https://github.com/giantswarm/giantswarm/issues/3839
-	//
-	err := c.Host.DeleteGuestCluster(ctx, "azure")
+	_, err := c.AzureClient.ResourceGroupsClient.Delete(ctx, env.ClusterID())
```
tuommaki: > Why don't we delete the resource group and forget about it?

Would it make sense to do both? My concern here is that when we just delete the resource group, is there a chance that `azure-operator` re-creates it when reconciling `AzureConfig` without `deleteTimestamp`?

fiunchinho: Good point. I've changed the code to first remove the Custom Resource.

PR #639, opened 2020-01-27
Refactor worker upgrade

tuommaki:
This change is a big overhaul to make upgrade process of worker nodes more
graceful. Problem with earlier implementation was that at times it happened
that multiple nodes might get reimaged before new nodes were already working
and this caused downtime to customer workload even when deployment had multiple
replicas scheduled.

Refactored upgrade process is now following:

- Upgrade master node[s] as before.
- Scale up number of workers to double.
- Cordon old nodes and wait until all new nodes are up.
- Drain old nodes.
- Once all old nodes are drained, terminate them.
- Scale down number of workers to desired amount.


tuommaki:
General tests on `godsmack` upgrading from `9.0.0` and `11.0.0` to this release have been successful on empty clusters. Next step is to put some workload on the clusters and monitor service reponses during upgrade.

service/controller/v13/resource/instance/create.go
```diff
@@ -4,21 +4,30 @@ import (
...
+		MasterInstancesUpgrading:    r.masterInstancesUpgradingTransition,
+		WaitForMastersToBecomeReady: r.waitForMastersToBecomeReadyTransition,
+		ScaleUpWorkerVMSS:           r.scaleUpWorkerVMSSTransition,
+		CordonOldWorkers:            r.cordonOldWorkersTransition,
+		WaitForWorkersToBecomeReady: r.waitForWorkersToBecomeReadyTransition,
+		CreateWorkerDrainerConfigs:  r.createWorkerDrainerConfigsTransition,
+		DeleteWorkerDrainerConfigs:  r.deleteWorkerDrainerConfigsTransition,
+		TerminateOldWorkerInstances: r.terminateOldWorkersTransition,
+		ScaleDownWorkerVMSS:         r.scaleDownWorkerVMSSTransition,
+		DeploymentCompleted:         r.deploymentCompletedTransition,
```
xh3b4sd: I just thought you can replicate the same with the operatorkit resources and the controller context. I see the need and the complexity but I think it is not a good practice to manage all of that logic in one resource package. We are pretty much replicating the reconciliation primitives inside the reconciliation primitives somehow. Just food for thought.

service/controller/v13/resource/instance/create_delete_worker_drainer_configs.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, dc := range drainerConfigs {
+		if !dc.Status.HasDrainedCondition() {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("node %s has not been drained yet", dc.Name))
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+			// Not all nodes have been drained yet. Resume later.
+			return currentState, nil
+		}
+
+		err = r.g8sClient.CoreV1alpha1().DrainerConfigs(dc.Namespace).Delete(dc.Name, &metav1.DeleteOptions{})
```
xh3b4sd: I don't follow the cleanup here. We delete some and then stop even though some could still be deleted. Shouldn't the filtering be done earlier so we delete all and then proceed by ignoring the rest of the drainer configs which are not "ready" yet?

service/controller/v13/resource/instance/create_drain_old_workers.go
```diff
@@ -0,0 +1,72 @@
...
+	if IsScaleSetNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the scale set '%s'", key.WorkerVMSSName(customObject)))
+	} else if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "processing worker VMSSs")
+
+	for _, i := range allWorkerInstances {
+		if *i.LatestModelApplied {
```
xh3b4sd: This here implies an instance got already updated, so we do not need to drain it again. A comment would be nice to have.

service/controller/v13/resource/instance/create_scale_workers.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (r *Resource) scaleDownWorkerVMSSTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	// Scale down to the desired number of nodes in worker VMSS.
+	nodeCountFunc := func(_ int64) int64 {
```
xh3b4sd: I don't follow why we need that callback. Why can't we simply configure the delta to `scaleVMSS`?

service/controller/v13/resource/instance/create_terminate_old_workers.go
```diff
@@ -0,0 +1,63 @@
...
+		return "", microerror.Mask(err)
+	}
+
+	g := key.ResourceGroupName(customObject)
+	s := key.WorkerVMSSName(customObject)
+	var ids compute.VirtualMachineScaleSetVMInstanceRequiredIDs
+	{
+		var strIds []string
+		for _, i := range allWorkerInstances {
+			if !*i.LatestModelApplied {
```
xh3b4sd: I this here then to delete all instances which are not updated? Does the update actually happen implicitly by scaling up in the first place? So instances do not get updated in place but rather created from scratch with the desired configuration and then we tear down the old instances we do not need anymore. Is that like a batched rolling upgrade?

service/controller/v13/resource/instance/create_wait_for_nodes_to_become_ready.go
```diff
@@ -0,0 +1,112 @@
...
+func isMaster(n corev1.Node) bool {
+	for k, v := range n.Labels {
+		switch k {
+		case "role":
+			return v == "master"
+		case "kubernetes.io/role":
+			return v == "master"
+		case "node-role.kubernetes.io/master":
+			return true
+		case "node.kubernetes.io/master":
```
xh3b4sd: Do Kubernetes they said. It will be fun they said. :trollface:

service/controller/v13/resource/instance/create_worker_instances_upgrading.go
```diff
@@ -0,0 +1,84 @@
...
+	var workerUpgradeInProgess bool
+	allWorkerInstances, err := r.allInstances(ctx, customObject, key.WorkerVMSSName)
+	if IsScaleSetNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the scale set '%s'", key.WorkerVMSSName(customObject)))
+	} else if err != nil {
+		return "", microerror.Mask(err)
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "processing worker VMSSs")
+
+		ws, err := r.nextInstance(ctx, customObject, allWorkerInstances, drainerConfigs, key.WorkerInstanceName, versionValue)
```
xh3b4sd: At this point I feel like I have a misunderstanding of the whole changeset. Why do we need to manage the single instances when we actually do the scale up magic?

tuommaki: No, you understood it right. I just forgot to delete the old code. Thanks for catching this! :laughing:

service/controller/v13/resource/instance/create.go
xh3b4sd: There could be one dedicated controller for this. Then it would be cohesive as well.

tuommaki: Do you have idea how that could synchronize with the "main" one if they reconcile the same CR? Specific condition flag in status?

xh3b4sd:
> would have brought big overhead

The tenant clients resource can just be added and merged without intensive testing for instance. In case you know which steps you are going you implement you could also have a PR for the layout and get the structure in place. But all good. Just saying we should always try to break it down.

service/controller/v13/resource/instance/create.go
xh3b4sd: I am not on top of all the internals and requirements. Can't make a good suggestion at this point.

service/controller/v13/resource/instance/create_cordon_old_workers.go
```diff
@@ -0,0 +1,161 @@
...
+		nodes = nodeList.Items
+	}
+
+	oldNodes, newNodes := sortNodesByTenantVMState(nodes, allWorkerInstances, customObject, key.WorkerInstanceName)
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found %d old and %d new nodes from tenant cluster", len(oldNodes), len(newNodes)))
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring old nodes are cordoned")
+
+	oldNodesCordoned, err := r.ensureNodesCordoned(ctx, oldNodes)
+	if tenant.IsAPINotAvailable(err) {
```
fiunchinho: How come we need to check if the API is available here?

service/controller/v13/resource/instance/create_cordon_old_workers.go
```diff
@@ -0,0 +1,161 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster API is not available")
+
+		return currentState, nil
+	} else if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	if len(newNodes) < len(oldNodes) {
```
fiunchinho: Wouldn't be safer to move this earlier and avoid cordoning the old nodes until the capacity has doubled?

service/controller/v13/resource/instance/create_cordon_old_workers.go
```diff
@@ -0,0 +1,161 @@
...
+			count++
+			continue
+		}
+
+		t := types.StrategicMergePatchType
+		p := []byte(UnschedulablePatch)
+
+		_, err = cc.Client.TenantCluster.K8s.CoreV1().Nodes().Patch(n.Name, t, p)
+		if apierrors.IsNotFound(err) {
+			// It might happen the node we want to drain got already removed. This
```
fiunchinho: Comment seems to be wrong

service/controller/v13/resource/instance/create_scale_workers.go
```diff
@@ -0,0 +1,80 @@
...
+		return microerror.Mask(err)
+	}
+
+	vmss, err := c.Get(ctx, key.ResourceGroupName(customObject), deploymentNameFunc(customObject))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	*vmss.Sku.Capacity = int64(nodeCount)
+	res, err := c.CreateOrUpdate(ctx, key.ResourceGroupName(customObject), deploymentNameFunc(customObject), vmss)
```
fiunchinho: Do we know if Azure has some kind of [optimistic concurrency control](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)? I don't see any other way of doing this than _getting_ and _updating_ the VMSS, but it sucks.

service/controller/v13/resource/instance/create_terminate_old_workers.go
```diff
@@ -0,0 +1,73 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found %d worker VMSS instances", len(allWorkerInstances)))
+
+	c, err := r.getScaleSetsClient(ctx)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "filtering instance IDs for old instances")
+
+	g := key.ResourceGroupName(customObject)
```
fiunchinho: Do we get charge by character count? **:stuck_out_tongue_winking_eye:**  `resourceGroupName` is more explicit, and makes it easier for me to understand

service/controller/v13/resource/instance/create_terminate_old_workers.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding all worker VMSS instances")
+	var allWorkerInstances []compute.VirtualMachineScaleSetVM
+	{
+		allWorkerInstances, err = r.allInstances(ctx, customObject, key.WorkerVMSSName)
+		if IsScaleSetNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the scale set '%s'", key.WorkerVMSSName(customObject)))
```
fiunchinho: Not sure whe scenario where this would happen, but would it make sense to move forward to the next state instead? I'm thinking that if we hit this scenario, the operator will get stuck on this state. We could print a warning but keep moving forward if the VMSS is not there? Just thinking out loud here.

service/controller/v13/controllercontext/client.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+type ContextClient struct {
+	TenantCluster ContextClientTenantCluster
+}
+
+type ContextClientTenantCluster struct {
+	G8s        versioned.Interface
+	K8s        kubernetes.Interface
+	CtrlClient client.Client
```
fiunchinho: is this used?

service/controller/v13/resource/instance/create_cordon_old_workers.go
tuommaki: Good point. Might just as well return wrapped error. I'll change.

service/controller/v13/controllercontext/client.go
tuommaki: Nope, not anymore. Good catch. Dropped.

service/controller/v13/resource/instance/create_cordon_old_workers.go
tuommaki: Yeah, that makes sense. I'll do that.

service/controller/v13/resource/instance/create_cordon_old_workers.go
tuommaki: :see_no_evil: True. I fixed this.

service/controller/v13/resource/instance/create_scale_workers.go
tuommaki: Not that I know of. I'm not sure if that would change much in this case. If external party changes VMSS size during update, it might do it even outside of this small scope and results would still be unknown.

service/controller/v13/resource/instance/create_terminate_old_workers.go
tuommaki: Would it even make sense to go to beginning?

service/controller/v13/resource/instance/create_terminate_old_workers.go
tuommaki: Practices on these tend to vary a bit depending on location, author and age of code. I think we generally lean towards shorter variable names when the scope of use is short or if it's about one specific type that we have documented. Other times then more descriptive ones are used.

I personally vary a bit on these. I'd definitely want to avoid the `superLongExplicitAllInclusiveVariableNamesThatAreUsedInJavaEECulture` but then again, single letter ones are not always good either. The code in here was actually copied / adapted from existing code in `azure-operator`.

Regarding the known variables we also have [fmt entry](https://github.com/giantswarm/fmt/blob/master/go/variable_names.md).

In this case these are just about avoiding long line with embedded function calls on function parameters so I'd like to keep these as-is.

service/controller/v13/resource/instance/create_terminate_old_workers.go
fiunchinho: You mean `DeploymentUninitialized`?

tuommaki: Yeah

fiunchinho: That could be a good idea yeah.

tuommaki:
Fishing for LGTM

tuommaki:
IMG_20200208_151635

PR #640, opened 2020-01-27
Check if tenant cluster resource group is deleted

fiunchinho:
Added new e2e test that makes sure all the resources are deleted when a `CustomResource` is deleted.

It would be good to also check that resources created on the control plane cluster (like DNS records) are also removed. But I can't focus on this task right now, so I think merging these changes already bring value, and we can always improve it later on.


.circleci/config.yml
```diff
@@ -191,15 +214,24 @@ workflows:
 
 
 
-#      - e2eTestCurPRMultiAZ:
-#          requires:
-#          - pr
+      - e2eTestCurPRMultiAZ:
```
fiunchinho: Since we made a new release, we can now run multi AZ tests with both `wip` and `current`.

integration/setup/teardown.go
```diff
@@ -19,7 +20,9 @@ func Teardown(c Config) error {
 	// We don't wait for the resources to be deleted, since we are going to delete them below.
 	err = c.Host.G8sClient().ProviderV1alpha1().AzureConfigs(c.Host.TargetNamespace()).Delete(clusterID, &metav1.DeleteOptions{})
 	if err != nil {
-		return microerror.Mask(err)
+		if !errors.IsNotFound(err) {
```
fiunchinho: This was needed because in the new test we delete the `CustomResource` before hand. I don't think this change affects anything else, and allow us to avoid erroring when the `CustomResource` is not there.

integration/setup/teardown.go
```diff
@@ -19,10 +19,10 @@ func Teardown(c Config) error {
 	// Delete AzureConfig from control plane to avoid reconciling while deleting the resources on Azure.
 	// We don't wait for the resources to be deleted, since we are going to delete them below.
 	err = c.Host.G8sClient().ProviderV1alpha1().AzureConfigs(c.Host.TargetNamespace()).Delete(clusterID, &metav1.DeleteOptions{})
-	if err != nil {
-		if !errors.IsNotFound(err) {
-			return microerror.Mask(err)
-		}
+	if errors.IsNotFound(err) {
+		// Fallthrough. This is fine.
```
tomahawk28: Do we still need to delete resource on Azure in this case? Just curious.

fiunchinho: Good point. I guess we don't. I'll return nil if the `CustomResource` is not found.

PR #641, opened 2020-01-29
Export logs from tests

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6334

This includes all the changes in vendor just for dicsussion here.

I've refactored the ignition structure a bit to make it possible to get service/logging script into nodes

If sig decides this approach is valid - I'll sory everything in separate PRs
If no - I'll abandon this before we will have ignition-operator and a better way to extend ignitions


main.go
```diff
@@ -146,6 +146,9 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CAFile, "", "Certificate authority file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Tenant.Ignition.Debug.Enabled, false, "Enable services which help debugging ignition.")
```
MarcelMue: I had the idea for `ignition-operator` to support fields like this in there and have debug units there. Then we could have e2e tests running with provider operators and ignition operator running. The provider operator wouldn't need modification while ignition operator injects the logging unit.

Not blocking here - just sharing my ideas.

.circleci/config.yml
```diff
@@ -70,6 +70,8 @@ jobs:
     <<: *e2eTest
   e2eTestWIPPRClusterState:
     environment:
+      IGNITION_DEBUG_ENABLED: "true"
+      IGNITION_DEBUG_LOGS_PREFIX: "wip-pr-clusterstate"
```
MarcelMue: Would everyone overwrite these or should we just have them enabled by default with a prefix like `azure-operator-wip-pr-clusterstate` so we don't need to worry about different operators and so on.

Also should inject a build number IMO.

corest: There is a clusterid included (might be changed to build number, but didn't find a clean way as there is no support for nested env variables) https://github.com/giantswarm/azure-operator/pull/641/files#diff-0a941e625c8bd762a7466d659e2664c2R43

corest: > Would everyone overwrite these or should we just have them enabled by default with a prefix like azure-operator-wip-pr-clusterstate so we don't need to worry about different operators and so on.

I thought about diffs here as for e.g. multiaz test you'll have a different prefix.

main.go
corest: I understand this.

Before this, I worked in the direction on support additions in form of separate ignition configs (https://github.com/giantswarm/azure-operator/tree/ignition_additions)
And it works fine, but in case of tests it is hard to pass my ignition config into container for tests. So I left that idea

corest:
So this works somehow unexpected a bit. With export being unabled - I'm getting debug service generated only into worker, not master (how is it possible when both have same template regaring debug services/files?). 

But what is more interesting - in logs I see only master logs, not worker logs. But master ignition doesn't even contain that service.

corest:
Ok, this is ready. All logs exported into logentries if you enabled those on test

corest:
@giantswarm/team-celestial feedback appreciated. Do want to keep this for another week

PR #642, opened 2020-02-03
Re-use already allocated Public IP for ingress LB

tuommaki:
This change picks up name of the already allocated public IP and assigns it as
annotation to ingress LB.

This is the complementary PR for https://github.com/giantswarm/azure-operator/pull/604
Towards https://github.com/giantswarm/giantswarm/issues/6636


tuommaki:
Tested both `11.1.0` and upgrade from `9.0.0` to `11.1.0` with this. Works with both.

PR #648, opened 2020-02-07
Added CircleCI step names

nprokopic:
Documented step names as suggested in https://github.com/giantswarm/fmt/pull/92

image


nprokopic:
Thanks folks!

PR #650, opened 2020-02-12
Ensure that scaling shortcut is active

tuommaki:
When tenant cluster workers are scaled, it's enough that VMSS deployment gets
updated and then we just consider it done. We don't want to re-image master[s]
nor replace worker nodes with new ones. We only want to update the VMSS
deployment and that configures the VMSS accordingly.


fiunchinho:
Nice catch @nprokopic

PR #653, opened 2020-02-13
Add check for mising tenant k8s client

tuommaki:
It can happen that tenant is not reachable and therefore the code should check
for missing client and keep the state until it's there.


tuommaki:
> Why this didn't happen with "pre state machine" code?

Do you mean in `instance` resource or in `tenantclients` resource?

fiunchinho:
Instance resource. It didn't got executed before `tenantclients` in the past (we didn't see this error before), right?.

tuommaki:
> Instance resource. It didn't got executed before tenantclients in the past (we didn't see this error before), right?.

We have had this "feature" before as well. The thing is that in some states it's expected that there's no tenant client (e.g. when creating cluster, during VMSS deployment upload) and it's ok. In general I think it should be per-state check and not block everything.

PR #656, opened 2020-02-14
Create tenant cluster k8s client when needed instead of saving it on the context for later

fiunchinho:
We use the controller context to store data that may be re-used by different resources along the same reconciliation loop.
The problem is that when resources go to the context to fetch data, they don't know whether or not the data will be there (we have nil checks to avoid panics). But for me, this makes the code harder to follow.
The k8s client for the tenant cluster is only used by one resource. But it was created and saved in the context by another resource. With the changes in this PR, it's more explicit that the k8s client for tenant cluster is a dependency of the resource.

What do you think?


service/controller/v12/cloudconfig/master_template.go
```diff
@@ -22,7 +22,7 @@ func (c CloudConfig) NewMasterCloudConfig(customObject providerv1alpha1.AzureCon
 	}
 
 	// On Azure only master nodes access etcd, so it is locked down.
-	customObject.Spec.Cluster.Etcd.Domain = "127.0.0.1"
+	customObject.Spec.Cluster.Etcd.Domain = "localhost"
```
fiunchinho: Forget about this change, it will be removed.

service/controller/v13/cloudconfig/master_template.go
```diff
@@ -22,7 +22,7 @@ func (c CloudConfig) NewMasterCloudConfig(customObject providerv1alpha1.AzureCon
 	}
 
 	// On Azure only master nodes access etcd, so it is locked down.
-	customObject.Spec.Cluster.Etcd.Domain = "127.0.0.1"
+	customObject.Spec.Cluster.Etcd.Domain = "localhost"
```
fiunchinho: Forget about this change, it will be removed.

nprokopic:
Some tests are failing, you might want to check that before merging.

PR #660, opened 2020-02-19
Gather ratelimiting metrics from VMSS client executed towards tenant worker VMSS

tuommaki:
In order to see different Azure service rate limits, invoke requests to
different resources to get specific rate limit headers.


tuommaki:
I'll drop this. I'm not happy with the approach. I think wrapping Azure SDK with decorator that can do metrics + potential caching is a lot better idea. I'd like to rather work towards that idea.

PR #662, opened 2020-02-19
Initial stage for state machine is the empty state

fiunchinho:
The initial state is the empty state, so we don't have to check if there is a state defined.
Also, every time we were returning the empty string as a state we use the constant.

### DIFF
--- a/service/controller/v14/resource/instance/create_deployment_completed.go
+++ b/service/controller/v14/resource/instance/create_deployment_completed.go
@@ -13,11 +13,11 @@ import (
 func (r *Resource) deploymentCompletedTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
 	customObject, err := key.ToCustomObject(obj)
 	if err != nil {
-		return "", microerror.Mask(err)
+		return DeploymentUninitialized, microerror.Mask(err)
 	}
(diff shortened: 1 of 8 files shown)


fiunchinho:
Moved changes to `v14`.

service/controller/v14/resource/instance/create_deployment_completed.go
```diff
@@ -13,11 +13,11 @@ import (
 func (r *Resource) deploymentCompletedTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
 	customObject, err := key.ToCustomObject(obj)
 	if err != nil {
-		return "", microerror.Mask(err)
+		return DeploymentUninitialized, microerror.Mask(err)
```
nprokopic: When you return a value from a transition function, it first goes through `Execute(...)`, which returns `""` in case of an error.
So you might want to modify this line here in state machine, to propagate retuned value, instead of returning `""` https://github.com/giantswarm/azure-operator/blob/6ca5d619838d159dfe92a6da7bd9236586b20d25/service/controller/v14/resource/instance/internal/state/funcs.go#L17

nprokopic: It's effectively `""` in both cases at the moment, but it might not be in the future :)

PR #665, opened 2020-02-20
Wire v14 version bundle

tuommaki:


service/collector/deployment.go
```diff
@@ -14,7 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/giantswarm/azure-operator/client"
-	"github.com/giantswarm/azure-operator/service/controller/v12/key"
+	"github.com/giantswarm/azure-operator/service/controller/v14/key"
```
tuommaki: I updated collector references to `key` package as a preparation for upcoming flattening of `azure-operator`.

service/controller/v14/version_bundle.go
```diff
@@ -47,6 +37,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "azure-operator",
-		Version: "2.9.0",
+		Version: "2.10.0",
```
tuommaki: I figured `2.10.0` is the next version we want to have? No major k8s version upgrades or so atm. so for me this made sense.

PR #666, opened 2020-02-20
Bumped coreOS before we forget

whites11:
Bumped the coreos version to 2303.4.0 (supported in westeurope, francecentral and germanywest) before we forget to do that.

### DIFF
--- a/service/controller/v14/version_bundle.go
+++ b/service/controller/v14/version_bundle.go
@@ -4,6 +4,8 @@ import (
 	"github.com/giantswarm/versionbundle"
 )
 
+const CoreosVersion = "2303.4.0"
+
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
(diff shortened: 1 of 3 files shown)


fiunchinho:
Would it make sense to define the CoreOS version as a constant at the v14 package level? And reference that constant from both places?

whites11:
> Would it make sense to define the CoreOS version as a constant at the v14 package level? And reference that constant from both places?

I did it, but not sure I like it

fiunchinho:
I think it's fine. What other people think?

service/controller/v14/version_bundle.go
```diff
@@ -4,6 +4,8 @@ import (
 	"github.com/giantswarm/versionbundle"
 )
 
+const CoreosVersion = "2303.4.0"
```
tuommaki: Minor note: `key` package could be better place in terms of "where is this configured?". That's where it is in `kvm-operator` as well: https://github.com/giantswarm/kvm-operator/blob/master/service/controller/key/key.go#L56

whites11: makes sense

whites11: done

PR #671, opened 2020-02-21
flattening: Removed current version bundle

fiunchinho:


nprokopic:
LGTM

fiunchinho:
I think I created way too many PR's :smile_cat: 

The `flatten` branch is a temporary branch used to keep track of all the changes that will end up in `master`. The `thiccc` branch should not be merged into flatten. This PR is not necessary.

PR #675, opened 2020-02-24
Use operator labels to tag resources using the operator version

fiunchinho:

Changed several things in this PR. The most important part is the switch from `VersionBundleVersion` to `OperatorVersion`. This change makes that the computed version is no longer coming from a field in the `AzureConfig` spec, but coming from the labels in the `AzureConfig` CR. A consequence of this is having to change the unit tests to set the version in the labels instead of using the `spec` field.


service/controller/resource_set.go
```diff
@@ -403,7 +403,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return false
 		}
 
-		if key.VersionBundleVersion(customObject) == project.BundleVersion() {
+		if key.OperatorVersion(customObject) == project.BundleVersion() {
```
fiunchinho: This is the most importante change. Now this operator will only handle `CustomResources` that contain a label with the version that matches the operator version. It won't use the `spec` field.

fiunchinho:
Adding @giantswarm/team-magic as reviewer, as I'm not sure if I'm doing the right thing adding the operator version as label to kubernetes resources created by the operator like `namespace`, `service` or `endpoints`.

When the operator tries to compute the desired state, [it only filters by name](https://github.com/giantswarm/azure-operator/blob/0e4c47ae0e631b2fad9b1620939dd7a62651cd3d/service/controller/resource/namespace/current.go#L27), so it should work normally. But maybe we don't want the label to be there anyway.

tuommaki:
>  labels should be updated when the tenant cluster is upgraded, right?

Yes, this is an important thing and should be covered before adding the label there.

service/controller/key/key.go
```diff
@@ -38,12 +38,12 @@ const (
...
 	AnnotationPrometheusCluster = "giantswarm.io/prometheus-cluster"
 
-	LabelApp           = "app"
-	LabelCluster       = "giantswarm.io/cluster"
-	LabelCustomer      = "customer"
-	LabelManagedBy     = "giantswarm.io/managed-by"
-	LabelOrganization  = "giantswarm.io/organization"
-	LabelVersionBundle = "giantswarm.io/version-bundle"
+	LabelApp             = "app"
+	LabelCluster         = "giantswarm.io/cluster"
```
kopiczko: I'm leaving the comment in the file so discussion is easier because of the github ui.

Why do you have the `flatten` branch? What we were doing so far was creating `legacy` branch out of `master` and switching draughstman deployment from there. Do you have different plan?

service/controller/key/key.go
```diff
@@ -398,6 +402,15 @@ func ToNodeCount(v interface{}) (int, error) {
 	return nodeCount, nil
 }
 
+func ToOperatorVersion(v interface{}) (string, error) {
+	customObject, err := ToCustomObject(v)
```
kopiczko: As a side note we should get the naming straight here https://github.com/giantswarm/fmt/blob/master/go/variable_names.md#custom-resource

service/controller/key/key.go
fiunchinho: I was following the same approach I saw [people used for kvm-operator](https://github.com/giantswarm/kvm-operator/pull/817). The final step will be a PR merging the `flatten` branch into `master`. Is there a problem with that approach?

kopiczko: I didn't notice how it was done in kvm-operator. If this approach is more convenient that's good.

fiunchinho:
Removed labels from k8s resources like namespace, endpoints or service.

service/controller/key/key.go
fiunchinho: Fixed here https://github.com/giantswarm/azure-operator/pull/680

service/controller/resource/endpoints/desired.go
```diff
@@ -26,12 +26,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
-				"cluster":                    key.ClusterID(cr),
-				"customer":                   key.ClusterCustomer(cr),
-				"giantswarm.io/cluster":      key.ClusterID(cr),
-				"giantswarm.io/organization": key.ClusterCustomer(cr),
-				"giantswarm.io/managed-by":   "azure-operator",
+				key.LabelApp:           "master",
+				key.LabelCluster:       key.ClusterID(cr),
+				key.LabelCustomer:      key.ClusterCustomer(cr),
+				key.LegacyLabelCluster: key.ClusterID(cr),
+				key.LabelManagedBy:     "azure-operator",
```
kopiczko: (if you apply that you'll also have to sort the import)

```suggestion
				key.LabelManagedBy:     project.Name(),
```

fiunchinho: Applied here and other several places as well.

fiunchinho:
Removed `Update`  e2e test since it won't work with the new flattened structure

service/controller/resource_set.go
```diff
@@ -403,7 +403,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return false
 		}
 
-		if key.VersionBundleVersion(cr) == project.BundleVersion() {
+		if key.OperatorVersion(cr) == project.BundleVersion() {
```
tuommaki: There is a separate task for this, but would it make sense to incorporate this change here as well:
https://github.com/giantswarm/giantswarm/issues/8637 ?

nprokopic:
> Removed labels from k8s resources like namespace, endpoints or service.

Just for the future reference also remove it then from the pull request text/description.

service/controller/resource_set.go
kopiczko: Separate. No need to bother within this PR.

PR #676, opened 2020-02-24
Fix thiccc deployment

fiunchinho:


helm/azure-operator-chart/templates/deployment.yaml
```diff
@@ -13,17 +13,16 @@ spec:
   selector:
     matchLabels:
       app: azure-operator
+      version: {{ .Chart.Version }}
```
fiunchinho: We didn't have this before, so it makes me wonder if it needs to be changed. Before deploying the flattened version, we first would need to re-deploy the `thiccc` branch on all installations so it starts using both labels right?

PR #679, opened 2020-02-25
Skip instances upgrade path during cluster creation

tuommaki:
When creating a new cluster it is enough to upload corresponding VMSS
deployments and be done with it. No need to upgrade master[s] or replace
workers.

Towards https://github.com/giantswarm/giantswarm/issues/9052


service/controller/resource/instance/create_cluster_upgrade_requirement_check.go
```diff
@@ -18,20 +18,34 @@ func (r *Resource) clusterUpgradeRequirementCheckTransition(ctx context.Context,
...
 		return DeploymentCompleted, nil
 	}
 
 	return MasterInstancesUpgrading, nil
 }
 
+func (r *Resource) isClusterCreating(cr providerv1alpha1.AzureConfig) bool {
+	// When cluster creation is in the beginning, it doesn't necessarily have
+	// any status conditions yet.
+	if len(cr.Status.Cluster.Conditions) == 0 {
```
xh3b4sd: I am making that up here now but it feels like there can be no conditions for different reasons. Would it be good enough to only take the `Creating` status condition into account? Not blocking here. Just throwing thoughts around.

tuommaki: I'm not sure if `Conditions` would be empty other than during creation. It seems that state transitions there use `replace` when patching.

Currently the logic is such that if it cannot pull `Nodes` from tenant API, it doesn't apply patches and this results on Azure often with behaviour that cluster is creating for few reconciliation rounds and then directly lands with `Created`.

Node Pool clusters don't use `statusresource` anymore so I was wondering if it makes sense to spend time fixing that? WDYT?

xh3b4sd: What about adding the `Creating` condition to `kubernetesd` as a workaround/fix for the time being and only rely on the proper condition here in the operator`

xh3b4sd: Otherwise just go with this. Your call.

tuommaki: That's a good shout. I'll do that. :+1:

service/controller/resource/instance/create_cluster_upgrade_requirement_check.go
```diff
@@ -18,20 +18,34 @@ func (r *Resource) clusterUpgradeRequirementCheckTransition(ctx context.Context,
...
+		return true
+	}
+	if cr.Status.Cluster.HasCreatingCondition() {
```
fiunchinho: who is/which part is removing the `Creating` condition?

tuommaki: `statusresource` - it's executed as a first resource in reconciliation loop.

PR #683, opened 2020-02-27
Export metric about expiration date of the access tokens

whites11:
See https://github.com/giantswarm/giantswarm/issues/8742

This PR adds a new Collector to the azure operator.
The collector provides metrics about the expiration timestamp for the secrets of the azure applications we use in the azure operator to connect to the subscriptions where the tenant clusters are deployed.

It requires additional permissions on the security principal to access the data (see https://github.com/Azure-Samples/azure-sdk-for-go-samples/issues/238).


tuommaki:
> It requires additional permissions on the security principal to access the data

This is the thing I was afraid. Can you pick the specific permissions needed for this and check with @T-Kukawka on our Azure customers if we could get the permissions? Reason to get them is obvious and extremely important for both sides (us and customers), but when looking the referenced PR and it's permissions screenshot, it raised concerns in me because for some dumb reason *read* and *write* was combined in the same checkbox so I'm not sure if we can have read-only permissions. In some sense I understand when customers don't want to give write permissions :/

service/collector/token_expiration.go
```diff
@@ -0,0 +1,109 @@
...
+			return microerror.Mask(err)
+		}
+
+		for applications.NotDone() {
+			app := applications.Value()
+
+			if *app.AppID == clientSet.ClientID {
+				for _, pc := range *app.PasswordCredentials {
+					ch <- prometheus.MustNewConstMetric(
+						tokenExpirationDesc,
```
fiunchinho: I am curious: is there a better way of modeling this metric? I know that a _date_ it's a weird metric. But it also feels weird to have a gauge to always be 1, labeled with a date. Will we be able to alert on this?

fiunchinho: Maybe the gauge can be the days until the certificate expires so that the alert will be triggered when the days are lower than certain threshold?

fiunchinho: https://www.robustperception.io/get-alerted-before-your-ssl-certificates-expire

whites11: We use the approach of storing the unix TS on similar things, such as the certificate expiration: https://prometheus.g8s.talos.westeurope.azure.gigantic.io/graph?g0.range_input=1h&g0.expr=cert_exporter_not_after&g0.tab=1

I'd stick with the same approach here as well, since we're talking about similar metrics

tuommaki: But I think in that other approach it's storing expiration date in unix timestamp as a value, not label, but here it seems to be label. My intuition would be to store value of `(pc.EndDate.Unix() - time.Now().Unix())` so you could simply alert when the value is close to zero. It makes alert function a lot more readable while the measurement is also pretty readable. Nevertheless, either way. Your call :slightly_smiling_face:

PR #689, opened 2020-03-02
Update Azure Service API versions

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9165

Update Azure SDK and Service API, since we were using old versions.


fiunchinho:
Superseeded by https://github.com/giantswarm/azure-operator/pull/692

PR #691, opened 2020-03-03
Use architect orbs instead of architect binary

fiunchinho:
Let's migrate to architect orbs.


fiunchinho:
In order to build the project, we need to pin the dependency on micrologger to `v0.1.0` , which is the version using `dep` instead of go modules.

But we need to fix other dependencies that also depend on micrologger:
- https://github.com/giantswarm/k8sclient/pull/30
- https://github.com/giantswarm/apprclient/pull/34

Gopkg.toml
```diff
@@ -90,19 +90,19 @@ required = [
...
-[[constraint]]
+[[override]]
   name = "github.com/giantswarm/microerror"
-  branch = "master"
+  version = "v0.1.0"
 
 [[constraint]]
   name = "github.com/giantswarm/microkit"
-  branch = "master"
+  branch = "micrologger-0-1-0"
```
tfussell: Why use `micrologger-0-1-0` branch instead of `v0.1.0` release?

.circleci/config.yml
```diff
@@ -21,37 +21,31 @@ e2eTest: &e2eTest
...
 version: 2.1
 jobs:
   build:
-    machine: true
+    executor: architect/architect
+    working_directory: /go/src/github.com/giantswarm/azure-operator
     steps:
-      - checkout
+      - checkout:
+          path: /go/src/github.com/giantswarm/azure-operator
```
tfussell: Is this still necessary? I think checkout checks out to the working directory by default and they are equal.

.circleci/config.yml
```diff
@@ -105,16 +85,24 @@ jobs:
...
   version: 2
   build_and_e2eTest:
     jobs:
       # build binary
+      - architect/go-test-legacy:
+          name: test
       - build:
+          requires:
+            - test
+      - architect/go-architect-legacy:
```
tfussell: I'm not sure what architect deploy does for azure-operator, but since there wasn't a `./architect deploy` step before, I think this isn't necessary here.

Gopkg.toml
fiunchinho: I was testing locally and I forgot this. Anyway, the microkit release `v0.1.0` depends on micrologger `master` branch. [I tried updating it but it fails](https://github.com/giantswarm/microkit/pull/131), I believe due to a similar issue with `microerror`. I guess I'd need to do all the same PR's I did yesterday, but for `microerror`?

.circleci/config.yml
fiunchinho: Removed

.circleci/config.yml
fiunchinho: Removed

Gopkg.toml
fiunchinho: Fixed with overrides

fiunchinho: Submitted several PR's to other projects. Working without overrides now

.circleci/config.yml
```diff
@@ -21,37 +21,30 @@ e2eTest: &e2eTest
...
     steps:
       - checkout
 
-      - run:
-          name: Install architect
-          command: |
-            wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
-            chmod +x ./architect
-            ./architect version
-            wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
```
tfussell: Strange that this was downloaded twice in the past.

.circleci/config.yml
```diff
@@ -21,37 +21,30 @@ e2eTest: &e2eTest
...
           keys:
             - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}-{{ checksum "/tmp/day" }}
             - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}
             - go-cache-v1-{{ checksum "/tmp/year" }}
 
       - run:
-          name: Build using architect
-          command: ./architect build
+          name: Build go project
+          command: go build
```
tfussell: `architect` injects a few globals into the binary. Did we confirm that these aren't necessary for deploying?

fiunchinho: It fails
```
panic: [{/go/src/github.com/giantswarm/azure-operator/main.go:70: } {/go/src/github.com/giantswarm/azure-operator/service/service.go:178: } {/go/src/github.com/giantswarm/azure-operator/service/controller/cluster.go:69: } {/go/src/github.com/giantswarm/azure-operator/service/controller/resource_set.go:232: } {/go/src/github.com/giantswarm/azure-operator/service/controller/resource/deployment/resource.go:57: deployment.Config.TemplateVersion must not be empty} {invalid config error}]

goroutine 1 [running]:
main.mainError.func1(0xc000430360, 0xc00018a900, 0xc00029ff00)
	/go/src/github.com/giantswarm/azure-operator/main.go:70 +0x493
github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/microkit/command/daemon.(*command).Execute(0xc000440210, 0xc0001daf00, 0xc0001866c0, 0x0, 0x4)
	/go/src/github.com/giantswarm/azure-operator/vendor/github.com/giantswarm/microkit/command/daemon/command.go:104 +0x1ac
github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra.(*Command).execute(0xc0001daf00, 0xc000186640, 0x4, 0x4, 0xc0001daf00, 0xc000186640)
	/go/src/github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra/command.go:830 +0x2aa
github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc0001db900, 0xc0001db900, 0xc00024d120, 0x1f)
	/go/src/github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra/command.go:914 +0x2fb
github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/giantswarm/azure-operator/vendor/github.com/spf13/cobra/command.go:864
main.mainError(0x17cf37d9, 0xc000481f50)
	/go/src/github.com/giantswarm/azure-operator/main.go:156 +0x1798
main.main()
	/go/src/github.com/giantswarm/azure-operator/main.go:35 +0x26
```

back to good old architect? or is there an alternative?

tfussell: I'm not sure what that specific variable is, but you can inject globals at build time using `go build -ldflags "-X <package>.<variable>=<value>"`. We definitely don't want to reimplement `architect` in a worse form here so we'll need to figure out how much work this will be to debug.

fiunchinho: This is the value that is missing https://github.com/giantswarm/architect/blob/e7baad9c841b6284164a3e7cb803810d98f0e76f/workflow/golang.go#L66

tfussell: Then something like `go build -ldflags "-X 'github.com/giantswarm/azure-operator/pkg/project.gitSHA=${CIRCLE_SHA1}'` might be sufficient.

fiunchinho: It fails, but it doesn't say why!? https://circleci.com/gh/giantswarm/azure-operator/8020

tfussell: It does, it's just interleaved with the other output:
```
# runtime/cgo
exec: "gcc": executable file not found in $PATH
```
I guess you need `CGO_ENABLED=0 go build ...`

fiunchinho: Superseeded by https://github.com/giantswarm/azure-operator/pull/701 (I needed a branch without underscores to be able to deploy it)

PR #692, opened 2020-03-03
Update Azure SDK & Service API versions

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9165

Update Azure SDK and Service API, since we were using old versions.


fiunchinho:
E2E tests passed correctly.

I also deployed this branch on a test installation and manually created a test to run some tests. Everything seems to work normally.

client/azure_client_set.go
```diff
@@ -3,9 +3,9 @@ package client
 import (
 	"fmt"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
```
nprokopic: What about other APIs, for example DNS here? Any specific reason not to update them?

There is a newer Service API version https://github.com/Azure/azure-sdk-for-go/tree/master/services/dns/mgmt/2018-05-01/dns

fiunchinho: I was focusing on the ones that could yield bigger impact. Should I update all of them? Maybe that's a good idea.

fiunchinho:
Done. I'll re-run e2e tests

client/azure_client_set.go
nprokopic: I would go for all of them :)

fiunchinho:
Manual test was ok. Sonobuoy tests were ok.

PR #697, opened 2020-03-05
Vmss rate limit

whites11:
Added a new exporter for the number of available calls we can make to the VMSS list VMs endpoint.

### DIFF
--- a/service/collector/vmss_rate_limit.go
+++ b/service/collector/vmss_rate_limit.go
@@ -1,0 +1,228 @@
+package collector
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


service/collector/vmss_rate_limit.go
```diff
@@ -0,0 +1,230 @@
...
+				continue
+			}
+			azureClients, err := client.NewAzureClientSet(*config)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			// VMSS List VMs specific limits.
+			{
+				// We need to find a resource group for an existing tenant cluster in this subscription.
```
fiunchinho: Comment is wrong.

whites11: Right, fixed

PR #701, opened 2020-03-06
Use architect orbs instead of the architect binary

fiunchinho:
Let's migrate to architect orbs.


fiunchinho:
Using this to build the binary seems to work.

```
GOCACHE=/go/cache GOPATH=/go CGO_ENABLED=0 go build -a -v -ldflags "-w -linkmode 'auto' -extldflags '-static' -X 'main.gitCommit=${CIRCLE_SHA1}' -X 'github.com/giantswarm/azure-operator/pkg/project.buildTimestamp=`date -I'seconds'`' -X 'github.com/giantswarm/azure-operator/pkg/project.gitSHA=${CIRCLE_SHA1}'"
```

I've deployed it to a test installation and it starts fine.

.circleci/config.yml
```diff
@@ -21,37 +21,30 @@ e2eTest: &e2eTest
...
           keys:
             - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}-{{ checksum "/tmp/day" }}
             - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}
             - go-cache-v1-{{ checksum "/tmp/year" }}
 
       - run:
-          name: Build using architect
-          command: ./architect build
+          name: Build go project
+          command: GOCACHE=/go/cache GOPATH=/go CGO_ENABLED=0 go build -a -v -ldflags "-w -linkmode 'auto' -extldflags '-static' -X 'main.gitCommit=${CIRCLE_SHA1}' -X 'github.com/giantswarm/azure-operator/pkg/project.buildTimestamp=`date -I'seconds'`' -X 'github.com/giantswarm/azure-operator/pkg/project.gitSHA=${CIRCLE_SHA1}'"
```
tfussell: Nice work getting this to build. My question would be: is this preferable to `architect build`? What do we get from running this directly? It's good to be explicit, but if we have to copy and paste this into several projects it might make sense to keep using `architect`. What do you think? If nothing else it would help to break it up into a few lines using backslashes.

fiunchinho: I don't think is worth it. I think it makes more sense to keep using architect build until we move to go modules and then we can start using all the new orbs, right?

tfussell: I guess so. It was worth a try. And we can still use the `go-architect-legacy` job to keep it nice and clean and still get the nice checks from `go-test-legacy`.

fiunchinho: I need your fix that stores the binary to be merged until this works, right?

tfussell: Yes, and it didn't make it into the 0.8.0 release unfortunately. I will try to make a patch release on Monday for it. Slow and steady...

PR #702, opened 2020-03-06
Update exposed info about using k8s 1.16.7

fiunchinho:
I forgot to update the version bundle info regarding the k8s version used.


fiunchinho:
@giantswarm/team-celestial Should I create `v3.0.1` just for this or overwrite existing `v3.0.0`? I know it's not good to overwrite tags that have been already pushed but... I'm pretty sure no one noticed it.

axbarsan:
I'd go with overriding v3.0.0, since it is needed by that version, so it could work properly

fiunchinho:
Overwritten `v3.0.0` with new commit.

PR #705, opened 2020-03-06
Cleanup reconciled version by updating microendpoint

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/8636 and https://github.com/giantswarm/giantswarm/issues/7391

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -92,17 +92,17 @@ required = [
   name = "github.com/giantswarm/microendpoint"
   branch = "master"
 
-[[constraint]]
+[[override]]
   name = "github.com/giantswarm/microerror"
-  branch = "master"
+  version = "~0.1.0"
 
 [[constraint]]
   name = "github.com/giantswarm/microkit"
   branch = "master"
 
-[[constraint]]
+[[override]]
   name = "github.com/giantswarm/micrologger"
-  branch = "master"
+  version = "~0.1.0"
 
 [[constraint]]
   name = "github.com/giantswarm/operatorkit"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -422,7 +422,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:568716a2e87ca3f4fb9bce7f65b8491a2ab92c97adee4b2b750b1d76c709b183"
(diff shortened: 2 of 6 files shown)


Gopkg.toml
```diff
@@ -92,17 +92,17 @@ required = [
   name = "github.com/giantswarm/microendpoint"
   branch = "master"
 
-[[constraint]]
+[[override]]
```
fiunchinho: why `override` instead of `constraint`?

wtty-fool: Because I need to override the following constraints in vendored repositories:
```
Solving failure: No versions of github.com/giantswarm/microendpoint met constraints:
        master: Could not introduce github.com/giantswarm/microendpoint@master, as it has a dependency on github.com/giantswarm/microerror with constraint ^0.1.0, which has no overlap with the following existing constraints:
        master from (root)
        master from github.com/giantswarm/apprclient@master
        master from github.com/giantswarm/helmclient@master
        master from github.com/giantswarm/e2e-harness@master
        master from github.com/giantswarm/errors@master
        master from github.com/giantswarm/operatorkit@master

        go-modules: Could not introduce github.com/giantswarm/microendpoint@go-modules, as it is not allowed by constraint master from project github.com/azure-operator.
```
@fiunchinho Is there a better way of doing this?

kopiczko: IMO it's ok as long as we don't have other libraries sorted.

kopiczko:
BTW @wtty-fool we have a policy we don't need to wait for approvals for vendor updates so you can go ahead and merge those with your admin rights unless you feel like you need an actual review.

Gopkg.toml
fiunchinho: But I believe I fixed all those other libraries to be able to build azure-operator:
- https://github.com/giantswarm/apprclient/pull/35/files#diff-836546cc53507f6b2d581088903b1785
- https://github.com/giantswarm/helmclient/pull/175/files#diff-836546cc53507f6b2d581088903b1785

etc

Maybe I missed one of them?

kopiczko: > Maybe I missed some of them?

Maybe 🤷‍♂

wtty-fool: Discussed on slack, marking this resolved for now.

PR #708, opened 2020-03-09
Already checked subscriptions should not be used again

fiunchinho:
Fixes https://github.com/giantswarm/giantswarm/issues/9298


fiunchinho:
Deployed on talos where the azure-operator was giving HTTP 500 errors, and now it works fine.

PR #710, opened 2020-03-10
Labels are immutable in v1, we need to go back so draughtsman can deploy

fiunchinho:
When we want to deploy the `thiccc` branch, the deployment fails unless we remove the helm release before deploying.

I believe the problem comes for the fact that we don't use `apiVersion: extensions/v1beta1` but `apiVersion: apps/v1` which made the field immutable.

### DIFF
--- a/helm/azure-operator-chart/templates/deployment.yaml
+++ b/helm/azure-operator-chart/templates/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1
+apiVersion: extensions/v1beta1
 kind: Deployment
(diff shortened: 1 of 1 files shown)


tuommaki:
I'm not sure about this. I have a vague memory we just fixed many non-stable CRD versions into stable ones and that makes me think if this is the right thing to do. Let's talk about this in Meet e.g. after standup.

fiunchinho:
After discussing with ross, he told me we need `apps/v1` to be able to update our CP clusters. I'll close this.

PR #714, opened 2020-03-11
Prevent infinite retries in general and on HTTP 429 completely

tuommaki:
Never retry on HTTP 429 (TooManyRequests) and in other cases retry only once in
Azure SDK clients.

### DIFF
--- a/client/azure_client_set.go
+++ b/client/azure_client_set.go
@@ -47,16 +47,7 @@ type AzureClientSet struct {
 
 func init() {
 	// ONE DOES NOT SIMPLY RETRY ON HTTP 429.
-	for i, sc := range autorest.StatusCodesForRetry {
-		if sc == http.StatusTooManyRequests {
-			// Shift end of slice to the left by one.
-			copy(autorest.StatusCodesForRetry[i:], autorest.StatusCodesForRetry[i+1:])
-			// Truncate the last element.
-			autorest.StatusCodesForRetry = autorest.StatusCodesForRetry[:len(autorest.StatusCodesForRetry)-1]
-			// Call it a day.
-			break
-		}
-	}
+	autorest.StatusCodesForRetry = removeElementFromSlice(autorest.StatusCodesForRetry, http.StatusTooManyRequests)
 }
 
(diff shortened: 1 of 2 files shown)


client/azure_client_set.go
```diff
@@ -44,6 +45,20 @@ type AzureClientSet struct {
 	VnetPeeringClient *network.VirtualNetworkPeeringsClient
 }
 
+func init() {
```
fiunchinho: I'm not a huge fan of using `init()`. I would move it to `NewAzureClientSet`, but I'm not blocking. Just a comment.

client/azure_client_set.go
```diff
@@ -44,6 +45,20 @@ type AzureClientSet struct {
...
+	// ONE DOES NOT SIMPLY RETRY ON HTTP 429.
+	for i, sc := range autorest.StatusCodesForRetry {
```
fiunchinho: It took me a couple of minutes to understand what this was doing. Can we wrap it in a function called `remove429FromRetries` or whatever? Or a code comment `Remove 429 from codes to retry`?

client/azure_client_set.go
```diff
@@ -44,6 +45,20 @@ type AzureClientSet struct {
...
+		if sc == http.StatusTooManyRequests {
+			// Shift end of slice to the left by one.
+			copy(autorest.StatusCodesForRetry[i:], autorest.StatusCodesForRetry[i+1:])
+			// Truncate the last element.
+			autorest.StatusCodesForRetry = autorest.StatusCodesForRetry[:len(autorest.StatusCodesForRetry)-1]
```
axbarsan: Nice 👍

client/azure_client_set.go
tuommaki: I'd like to keep it in `init()` because it's effectively modifying the global variable from `autorest` package. Therefore it should be only once, although the later iterations are no-ops.

client/azure_client_set.go
tuommaki: I'll write a `removeElementFromSlice()` function so it's more readable :slightly_smiling_face:

PR #717, opened 2020-03-11
Use new version of architect-orbs, make linter happy

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9511

Run e2e tests using new architect-orb.

Install azure-operator from catalog during tests.

Simplify tests configuration setting sane defaults.

Linting changes:
- Helm chart is now linted as part of CI.
- Code that was unused got removed
- sha1 got replaced with more secure algo
- unhandled errors got handled

I had to use annotations to disable the linting on some parts of the code, like when we don't check the error coming from the logger.


integration/setup/bastion.go
```diff
@@ -120,7 +120,7 @@ func CreatePublicIP(ctx context.Context, location, groupName string, ipClient ne
 func CreateVM(ctx context.Context, location, groupName, sshKeyData string, nic network.Interface, config Config) error {
 	vmName := "bastionE2EVirtualMachine"
 	username := "e2e"
-	config.Logger.LogCtx(ctx, "message", "Creating e2e bastion instance", "vmName", vmName)
+	config.Logger.LogCtx(ctx, "message", "Creating e2e bastion instance", "vmName", vmName) // nolint: errcheck
```
kopiczko: Why do we need all those `// nolint: errcheck`? Is this something we can fix in the micrologger?

fiunchinho: https://gigantic.slack.com/archives/C0251MXL5/p1584003641116900

fiunchinho: I'm not sure a logger should return an error.
Until a decision is made, I believe it's fine to ignore the lint errors in log calls.

integration/config/setup.sh
```diff
@@ -0,0 +1,12 @@
...
+shipyard_dir=".e2e-harness/workdir/.shipyard"
+mkdir -p "${shipyard_dir}"
+chmod -R 777 ".e2e-harness"
+kind get kubeconfig --name="kind" | sudo tee "${shipyard_dir}/config" > /dev/null
+
+
+shipyard_dir="/workdir/.shipyard"
+sudo mkdir -p "${shipyard_dir}"
+sudo chmod -R 777 "/workdir"
+kind get kubeconfig --name="kind" | sudo tee "${shipyard_dir}/config" > /dev/null
```
kopiczko: I'd skip `sudo` here. The parent dir has already 777 and we want this file to be readable by the program executed with `go test`.

```suggestion
kind get kubeconfig --name="kind" > "${shipyard_dir}/config"
```

integration/config/setup.sh
```diff
@@ -0,0 +1,20 @@
...
+
+shipyard_dir="/workdir/.shipyard"
+sudo mkdir -p "${shipyard_dir}"
+sudo chmod -R 777 "/workdir"
+kind get kubeconfig --name="kind" > "${shipyard_dir}/config"
+
+curl -L https://get.helm.sh/helm-v2.16.1-linux-amd64.tar.gz > ./helm.tar.gz
+tar xzvf helm.tar.gz
+chmod u+x linux-amd64/helm
+sudo mv linux-amd64/helm /usr/local/bin/
```
kopiczko: Installing helm binary should be moved to architect-orb. But it's good to have it here during prototyping.

I'd say that installing tiller should be also moved there because of the RBAC craziness.

integration/setup/setup.go
```diff
@@ -62,3 +80,248 @@ func Setup(ctx context.Context, c Config) error {
...
+		}
+
+		_, err := config.K8sClients.K8sClient().CoreV1().Secrets("default").Create(encryptionSecret)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	{
+		azureConfig := &providerv1alpha1.AzureConfig{
```
fiunchinho: This is one of the highlights: we create the `AzureConfig` using client-go instead of using the e2e test helm chart in the apiextensions repository.

helm/azure-operator-chart/templates/deployment.yaml
```diff
@@ -83,5 +83,3 @@ spec:
           limits:
             cpu: 250m
             memory: 250Mi
-      imagePullSecrets:
```
fiunchinho: Removed pull secret.

helm/azure-operator/ci/default-values.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
fiunchinho: This file is required by the Helm linting process in the orb.

fiunchinho:
Merging this for now.

PR #721, opened 2020-03-17
Helm preparation to get component versions from releases

tuommaki:
These changes prepare Helm deployment templates to get cluster component
versions from releases instead of k8scloudconfig.

Towards https://github.com/giantswarm/giantswarm/issues/8436

### DIFF
--- a/helm/azure-operator-chart/templates/deployment.yaml
+++ b/helm/azure-operator-chart/templates/deployment.yaml
@@ -1,10 +1,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: azure-operator
-  namespace: {{ .Values.namespace }}
+  name: {{ tpl .Values.resource.default.name .}}
+  namespace: {{ tpl .Values.resource.default.namespace . }}
   labels:
-    app: azure-operator
+    app: {{ tpl .Values.resource.default.name . }}
   annotations:
     checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
 spec:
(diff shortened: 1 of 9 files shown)


helm/azure-operator-chart/templates/rbac.yaml
```diff
@@ -58,14 +64,6 @@ rules:
       - get
       - list
       - watch
-  - apiGroups:
-      - ""
-    resources:
-      - configmaps
-    resourceNames:
-      - aws-operator-configmap
```
tuommaki: I'm pretty sure this wasn't needed :joy:

helm/azure-operator-chart/templates/deployment.yaml
```diff
@@ -21,37 +21,37 @@ spec:
       annotations:
         releasetime: {{ $.Release.Time }}
       labels:
-        app: azure-operator
+        app: {{ tpl .Values.resource.default.name . }}
         version: {{ .Chart.Version }}
     spec:
       volumes:
-      - name: azure-operator-configmap
+      - name: {{ .Values.project.name }}-configmap
```
fiunchinho: I had the same question when flattening: why do we sometimes use `.Values.project.name` and some other times `tpl .Values.resource.default.name`?

tuommaki: I asked about this this in the `aws-operator` PR and was thinking the reason is probably that you want to have multiple operators running in parallel so resource names are distinct per release/version, but the YAML content for manifests are still the same across different versions.

kopiczko: Stuff in `values.yaml` is not templated. So if you used `{{ .Values.resource.default.name . }}` the output would be `{{ .Release.Name | replace "." "-" | trunc 47 }}`. with `tpl` it actually templates the value output before rendering.

kopiczko: I misinterpreted the original question. I talked with @fiunchinho and the outcome is we shouldn't use `.Values.project.name` in general.

tuommaki:
Deploying this works now.

PR #722, opened 2020-03-17
No need to send to metrics channel already registered metrics

fiunchinho:
Fixes https://github.com/giantswarm/giantswarm/issues/9298

We were getting errors due to metrics being collected twice
```
An error has occurred while serving metrics:

collected metric "azure_operator_rate_limit_vmss_instance_list_parsing_errors" { counter:<value:150 > } was collected before with the same name and label values
```

After removing the lines in the changes, metrics seem to work normally.

image


fiunchinho:
The other approach I guess it would be not to use the built-in counter, and just increment a variable, but I don't see how's better.

whites11:
To be backported on master if it works

PR #724, opened 2020-03-17
Hardcode selector label since it can't be changed when using apps/v1 apiversion

fiunchinho:
Fixes https://github.com/giantswarm/giantswarm/issues/9440

When redeploying `thiccc` branch, kubernetes complained because the label `version` is immutable using `apps/v1` apiVersion. Instead of using the git sha as version just harcode the `thiccc` string, since there will be only one `thiccc` deployment.

### DIFF
--- a/helm/azure-operator-chart/templates/deployment.yaml
+++ b/helm/azure-operator-chart/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: {{ .Values.namespace }}
   labels:
     app: azure-operator
+    version: thiccc
   annotations:
     checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
 spec:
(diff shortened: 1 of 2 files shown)


fiunchinho:
```
opsctl -i godsmack deploy -m github azure-operator@hardcode-selector-labels 
Requesting the creation of deployment 

    repository      azure-operator
    ref             hardcode-selector-labels
    version         75b246d9ce96307d1eca780ca88fe4c83a1cf1b8
    installation    godsmack

Requested the creation of deployment

Waiting for deployment ..........................................................................
Success!

$ opsctl -i godsmack deploy -m github azure-operator@hardcode-selector-labels 
Requesting the creation of deployment 

    repository      azure-operator
    ref             hardcode-selector-labels
    version         75b246d9ce96307d1eca780ca88fe4c83a1cf1b8
    installation    godsmack

Requested the creation of deployment

Waiting for deployment .................................................
Success!
```

PR #726, opened 2020-03-23
Update vmss.json

T-Kukawka:
Removing copy part of disks, that unables upgrade to 11.X as we are adding additional disk there, that conflicts with PVCs.


T-Kukawka:
Need to test upgrade with this change

PR #727, opened 2020-03-23
Get components from releases

tuommaki:
Cluster components such as k8s hyperkube image, CNI version etc. are now
defined in Release CRs and extracted from there and configured into
k8scloudconfig. This allows easier k8s releases.


tuommaki:
On `ghost`:
```
$ kubectl --context giantswarm-3aqai get nodes
NAME                  STATUS   ROLES    AGE     VERSION
3aqai-master-000000   Ready    master   8m56s   v1.16.8
3aqai-worker-000002   Ready    worker   107s    v1.16.8
3aqai-worker-000004   Ready    worker   108s    v1.16.8
3aqai-worker-000005   Ready    worker   108s    v1.16.8
```

service/controller/cloudconfig/types.go
```diff
@@ -46,3 +49,10 @@ type certificateDecrypterUnitParams struct {
 type ingressLBFileParams struct {
 	ClusterDNSDomain string
 }
+
+type IgnitionTemplateData struct {
+	CustomObject providerv1alpha1.AzureConfig
```
whites11: Maybe sort the keys in alphabetical order?

pkg/project/version_bundle.go
```diff
@@ -15,27 +13,11 @@ func NewVersionBundle() versionbundle.Bundle {
...
-				Version: "18.06.1",
-			},
-			{
-				Name:    "etcd",
-				Version: "3.3.17",
-			},
-			{
-				Name:    "kubernetes",
-				Version: "1.16.7",
+				Component:   Name(),
```
MarcelMue: You don't need these changelogs at all anymore.

MarcelMue: See fancy changelog md in aws-operator: https://github.com/giantswarm/aws-operator/blob/master/CHANGELOG.md

pkg/label/label.go
```diff
@@ -0,0 +1,17 @@
+package label
+
+const (
+	App = "app"
+)
+
```
fiunchinho: Why separate blocks of constants?

service/service.go
```diff
@@ -169,6 +169,7 @@ func New(config Config) (*Service, error) {
 			OIDC:             OIDC,
 			InstallationName: config.Viper.GetString(config.Flag.Service.Installation.Name),
 			ProjectName:      config.ProjectName,
+			RegistryDomain:   config.Viper.GetString(config.Flag.Service.RegistryDomain),
```
fiunchinho: Would it be possible to have default values to avoid panicking later? I believe we will use `quay.io` in most cases.

pkg/label/label.go
tuommaki: No clue. I just copied this from `aws-operator` as that is SOTA. It seems like logical grouping.

pkg/project/version_bundle.go
tuommaki: Ah, right. :+1:

service/controller/cloudconfig/types.go
tuommaki: I'm not actually sure what's the right thing here. Above types are non-exported and this one is exported. In `aws-operator` it was like this as well. I'd maybe prefer this instead of mixing exported and non-exported types?

service/service.go
tuommaki: I think flag registration in `main.go` should provide the default, no?

fiunchinho: Yes, I missed that entirely. Never mind!

tuommaki:
I successfully now tested also upgrade from `9.0.0` in `ghost/trial-subscription`. Going with this.

service/controller/key/key.go
```diff
@@ -52,7 +53,21 @@ const (
...
+// Container image versions for k8scloudconfig.
+const (
+	// kubectl container version.
+	kubectlVersion = "1.16.4"
+
+	// k8s-api-healthz version.
+	kubernetesAPIHealthzVersion = "0999549a4c334b646288d08bd2c781c6aae2e12f"
+
+	// k8s-setup-network-environment.
+	kubernetesNetworkSetupDocker = "1f4ffc52095ac368847ce3428ea99b257003d9b9"
```
tfussell, line 67: @giantswarm/team-celestial I just noticed that this is a very old image corresponding to https://github.com/giantswarm/k8s-setup-network-environment/pull/2 I think. The latest is `68e90113331feca3b9ffe6a75a601b381ba8c1f7` (from https://quay.io/repository/giantswarm/k8s-setup-network-environment?tab=tags). I'm not sure the implications of this regression--just wanted to mention.

PR #728, opened 2020-03-24
Revert cloudconfig 5.1.1 added to use k8s 1.16.7

fiunchinho:
This doesn't make sense anymore since we are going with cloudconfig 6.x.x.

Removing also from cloudconfig https://github.com/giantswarm/k8scloudconfig/pull/669

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -407,7 +407,7 @@
   name = "github.com/giantswarm/k8scloudconfig"
   packages = [
     "ignition/v_2_2_0",
-    "v_5_1_1",
+    "v_5_1_0",
   ]
   pruneopts = "T"
   revision = "a952556a3a7ed07f8d126f095f996d5c0378a3a8"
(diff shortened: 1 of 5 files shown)


fiunchinho:
@giantswarm/team-celestial the tag `v3.0.0` was created but still never released. This tag contains the change that I'm reverting here. It's fine just releasing `v3.0.1`, or should we overwrite `v3.0.0` so all tags make sense?

tuommaki:
I'd maybe drop this because my upcoming PR will supersede that anyway: https://github.com/giantswarm/azure-operator/pull/727 

This would just create a merge conflict without any benefits. WDYT?

tuommaki:
...I wouldn't maybe worry too much about `3.0.0` when it was never used anywhere. `3.0.1` will become the next release and we can drop `3.0.0` completely from catalogs. The repository tag can stay there as we already agreed that we don't worry about mistakes.

fiunchinho:
Agree!

PR #729, opened 2020-03-25
Update ARM deployments' schema version

tuommaki:


fiunchinho:
Can you write a sentence or two on why are we trying this on the PR description?

tuommaki:
> Can you write a sentence or two on why are we trying this on the PR description?

This was an idea to fix the problems with LUN definitions in `copy` declaration, but this doesn't help there after all and I don't think we should do this now. Hence dropping.

PR #733, opened 2020-03-26
Use ephemeral storage for /var/lib

whites11:
Using ephemeral storage from VM for /var/lib;
Dropped kubelet and docker volumes formatting and mounting


service/controller/resource/instance/template/main.json
```diff
@@ -113,15 +113,8 @@
...
-              },
-              {
-                "name":"DockerDisk",
-                "sizeGB":"[if(greater(parameters('masterNodes')[0].dockerVolumeSizeGB, 0), parameters('masterNodes')[0].dockerVolumeSizeGB, 50)]"
-              },
-              {
-                "name":"KubeletDisk",
-                "sizeGB":"[if(greater(parameters('masterNodes')[0].kubeletVolumeSizeGB, 0), parameters('masterNodes')[0].kubeletVolumeSizeGB, 100)]"
+                "sizeGB":100,
+                "lun": 0
```
T-Kukawka: I guess we don't need that if we go through index?

PR #734, opened 2020-03-26
[Demo] Just set VMSS dataDisks, instead of using copy

nprokopic:
The idea here is to just set `dataDisks` field for VMSS `storageProfile`, instead of using ARM `copy` function.

Changes:
- `vmssVmDataDisks` in `main.json` has to be in exact same format as `virtualMachineProfile.storageProfile.dataDisks`.
- In `vmss.json`, `dataDisks` is just set to `vmssVmDataDisks` array.
- Setting `caching` to none, according to some advices found [here](https://stian.tech/disk-performance-on-aks-part-1/) and [here](https://github.com/jnoller/kubernaughty/issues/42)

**Just for demo purposes, some stuff is hardcoded.**

### DIFF
--- a/service/controller/resource/instance/template/main.json
+++ b/service/controller/resource/instance/template/main.json
@@ -113,15 +113,33 @@
             "value":[
               {
                 "name":"EtcdDisk",
-                "sizeGB":100
+                "diskSizeGB":100,
+                "lun": 0,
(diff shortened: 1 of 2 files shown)


whites11:
done in another PR

PR #736, opened 2020-03-27
Improve collectors

fiunchinho:
I tried to improve code a little bit to make it explicit, and make collectors look similar to each other, by extracting common pieces to a single place. Now collectors fetch k8s `Secret` by label instead of looping through all CR's.

Also, we expose some of our metrics using an Azure client for every subscription that is used in the control plane. We always expose _by subscription_ , but some metrics may need to be exposed by subscription id, clientid or tenant id.

I invested some time on this because even though we change to ClusterAPI implementation, the collectors can be extracted and re-used.


service/collector/rate_limit.go
```diff
@@ -121,61 +120,25 @@ func NewRateLimit(config RateLimitConfig) (*RateLimit, error) {
...
-			if err != nil {
-				return microerror.Mask(err)
-			}
-
-			crs = append(crs, list.Items...)
-
-			mark = list.Continue
-			page++
-		}
+	clientSets, err := credential.GetAzureClientSetsFromCredentialSecrets(u.k8sClient, u.environmentName)
```
fiunchinho: This collector will expose metrics not just by subscription. It will expose metric for every subscription id, client id and tenant id.

service/collector/resource_group.go
```diff
@@ -67,17 +72,26 @@ func NewResourceGroup(config ResourceGroupConfig) (*ResourceGroup, error) {
...
+		environmentName:        config.EnvironmentName,
+		cpAzureClientSetConfig: config.CPAzureClientSetConfig,
 	}
 
 	return r, nil
 }
 
 func (r *ResourceGroup) Collect(ch chan<- prometheus.Metric) error {
-	clientSets, err := getClientSets(r.k8sClient, r.environmentName)
+	clientSets, err := credential.GetAzureClientSetsFromCredentialSecretsBySubscription(r.k8sClient, r.environmentName)
```
fiunchinho: This collector will still expose metrics only by subscription. Resource groups are the same even using different client id or tenant id.

service/collector/usage.go
```diff
@@ -100,52 +98,33 @@ func NewUsage(config UsageConfig) (*Usage, error) {
...
-				return microerror.Mask(err)
-			}
-
-			crs = append(crs, list.Items...)
-
-			mark = list.Continue
-			page++
-		}
+	ctx := context.Background()
+	clientSets, err := credential.GetAzureClientSetsFromCredentialSecretsBySubscription(u.k8sClient, u.environmentName)
```
fiunchinho: This collector will still expose metrics only by subscription. Resource groups are the same even using different client id or tenant id.

service/credential/credential.go
```diff
@@ -64,6 +66,111 @@ func GetAzureConfig(k8sClient kubernetes.Interface, name string, namespace strin
 	return c, nil
 }
 
+func GetAzureConfigFromSecret(credential *v1.Secret) (*client.AzureClientSetConfig, error) {
```
nprokopic: Nice! :)

service/collector/rate_limit.go
```diff
@@ -121,61 +121,25 @@ func NewRateLimit(config RateLimitConfig) (*RateLimit, error) {
...
 	// We track RateLimit metrics for each client labeled by SubscriptionID and
 	// ClientID.
 	// That way we prevent duplicated metrics.
-	for _, clientConfig := range clientConfigBySubscription {
-		azureClients, err := client.NewAzureClientSet(clientConfig)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-
+	for clientConfig, clientSet := range clientSets {
```
tuommaki: Would it make sense to extract loop body into function and execute it concurrently in [sync.ErrGroup](https://godoc.org/golang.org/x/sync/errgroup)?

service/credential/credential.go
nprokopic: (I was referring to all changes in `credential.go`, nicely extracted functionality)

service/collector/rate_limit.go
fiunchinho: I will merge this and iterate on this on the [new azure-collector repository](https://github.com/giantswarm/azure-collector), if that's ok with you.

PR #737, opened 2020-03-30
Update k8scloudconfig 6.0.0 dependency

fiunchinho:


vendor/github.com/giantswarm/k8scloudconfig/v_6_0_0/types.go
```diff
@@ -43,9 +43,11 @@ type Params struct {
 	// cancelled if no progress has been made.
 	ImagePullProgressDeadline string
 	// Container images used in the cloud-config templates
-	Images       Images
-	Node         v1alpha1.ClusterNode
-	SSOPublicKey string
+	Images         Images
+	Node           v1alpha1.ClusterNode
+	RegistryDomain string
```
fiunchinho: I guess this was missing somehow?

whites11: Where do we set the value for those new variables? Can't find it :(

fiunchinho: It's one of the [default values](https://github.com/giantswarm/azure-operator/pull/737/files#diff-fa532a4f5909f5fae52d248c090acdbfR36).

tuommaki: It should actually come from `installation`, end up to configmap and from configmap with Viper get wired into `blobobject` resource where it's used here:
https://github.com/giantswarm/azure-operator/blob/master/service/controller/resource/blobobject/desired.go#L75

Due to current API in `k8scloudconfig`, the Default Values don't actually fully work as intuition would expect. If you check the line 72-74 above, you see field-by-field copy. Hence the registry from default values would never get anyway, but then again, I'm not sure if it should be used?

Does this help?

fiunchinho:
```
 k get no
NAME                  STATUS   ROLES    AGE     VERSION
e0tzs-master-000000   Ready    master   3m13s   v1.16.8
e0tzs-worker-000000   Ready    worker   25s     v1.16.8
e0tzs-worker-000001   Ready    worker   31s     v1.16.8
e0tzs-worker-000002   Ready    worker   27s     v1.16.8
```

PR #741, opened 2020-03-31
Check VMSS rate limit values before proceeding with VMSS changes.

whites11:
This PR adds a check to the VMSS API rate limit current value before scaling up one worker node in the cluster during the upgrade.
Adding a node is the most stressful operation from a VMSS API calls point of view.
If there are not at least 50% of the API calls available, it'll wait for next reconciliation loop.
Open to suggestions.

### DIFF
--- a/service/controller/resource/instance/instance_update_watchdog.go
+++ b/service/controller/resource/instance/instance_update_watchdog.go
@@ -3,10 +3,58 @@ package instance
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/giantswarm/microerror"
 )
+
+const (
+	vmssVMListHeaderName = "checkVMSSApiRateLimitThresholds"
+	headerKey3m          = "3m"
+	headerKey30m         = "30m"
+	max3m                = 190
+	max30m               = 900
+	threshold3m          = max3m * 0.5
+	threshold30m         = max30m * 0.5
+)
+
+func (r *Resource) checkVMSSApiRateLimitThresholds(response autorest.Response) (int64, int64) {
+	headers := response.Header[vmssVMListHeaderName]
+
+	rl3m := int64(-1)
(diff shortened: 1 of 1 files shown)


service/controller/resource/instance/instance_update_watchdog.go
```diff
@@ -3,11 +3,59 @@ package instance
...
+	vmssVMListHeaderName = "checkVMSSApiRateLimitThresholds"
+	headerKey3m          = "3m"
+	headerKey30m         = "30m"
+	max3m                = 190
+	max30m               = 900
+	threshold3m          = max3m * 0.5
+	threshold30m         = max30m * 0.5
+)
+
+func (r *Resource) checkVMSSApiRateLimitThresholds(response autorest.Response) (int64, int64) {
```
tuommaki:
```suggestion
func checkVMSSApiRateLimitThresholds(response autorest.Response) (int64, int64) {
```
Minor nitpick, but seems that this can be just a pure function.

service/controller/resource/instance/instance_update_watchdog.go
```diff
@@ -22,7 +70,12 @@ func (r *Resource) ensureWorkerInstancesAreAllRunning(ctx context.Context, rg st
 		return false, microerror.Mask(err)
 	}
 
-	// TODO check for rate limit.
+	// Check for rate limit. If current remaining API calls are less than the desider threshold, we don't proceed.
+	rl3m, rl30m := r.checkVMSSApiRateLimitThresholds(iterator.Response().Response)
```
tuommaki:
```suggestion
	rl3m, rl30m := checkVMSSApiRateLimitThresholds(iterator.Response().Response)
```

service/controller/resource/instance/instance_update_watchdog.go
whites11: done

service/controller/resource/instance/instance_update_watchdog.go
whites11: done

PR #742, opened 2020-03-31
Add initial structure for VMSS check worker pool

tuommaki:
This change provides initial structure that allows implementation of VMSS
instance state checks to be executed as jobs in worker pool. This mitigates
concerns of resource leaks and still allows frequent low-latency checks for
VMSS instances to quickly fix them in case they get into Failed state.


service/controller/resource/instance/internal/vmsscheck/instance_watchdog.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	return wd, nil
+}
+
+func (wd *concurrentInstanceWatchdog) GuardVMSS(ctx context.Context, resourceGroupName, vmssName string) {
+	guardName := vmssGuardName(resourceGroupName, vmssName)
+	_, exists := wd.vmssGuards[guardName]
+	if exists {
+		return
```
tuommaki: One check per resource group / VMSS is enough.

tuommaki:
This was tested with additional changes in chained PR and review was also done there in addition to this. Going with this.

PR #744, opened 2020-03-31
Moved instance watchdog implementation to new vmsscheck.guardJob

nprokopic:
Moved instance watchdog implementation to new vmsscheck.guardJob

Slight issue:

One of the functions that was moved (actually copied) is `ensureWorkerInstancesAreAllRunning`, now it's here (with the rest of functionality):
https://github.com/giantswarm/azure-operator/blob/5053352ecc45c65bdfea7f28c79135dd4edfe7a6/service/controller/resource/instance/internal/vmsscheck/guard_job.go#L91

Here where it is originally implemented (not deleted from there):
https://github.com/giantswarm/azure-operator/blob/5053352ecc45c65bdfea7f28c79135dd4edfe7a6/service/controller/resource/instance/instance_update_watchdog.go#L61

The original `ensureWorkerInstancesAreAllRunning` function is also called here:
https://github.com/giantswarm/azure-operator/blob/5053352ecc45c65bdfea7f28c79135dd4edfe7a6/service/controller/resource/instance/create_scale_workers.go#L39

And according to the new design @tuommaki implemented, when I moved `ensureWorkerInstancesAreAllRunning` (and other pieces of functionality) into `vmsscheck` package, it got hidden from the instance resource, since the resource is now using `InstanceWatchdog` interface, and `vmsscheck.guardJob` is internal implementation of the `InstanceWatchdog`.

So some ideas:
- Extend `vmsscheck` public API so we can check the state of instances from the outside. This info is already available in `guardJob.allInstancesSucceeded` field, so exposing that to `InstanceWatchdog` and then from there further to the outside (to resource in this case) could maybe do the trick in an even almost non-hacky nice way :) E.g. adding something like `func AllGuardsSuccessful() bool` to `InstanceWatchdog`, or more fine-grained check if successful per job ID, and then call that in `create_scale_workers.go`.
- Move check functionality into a shared package where it would be called from both `vmsscheck` package and instance resource
- Refactor checking in `create_scale_workers.go` and possibly parts of `vmsscheck` packages so that check implementation (guardJob and/or instanceWatchdog implementation) is accessible from the outside. (not really liking this)


tuommaki:
I tested an upgrade and this works.

service/controller/resource/instance/internal/vmsscheck/guard_job.go
```diff
@@ -13,13 +55,93 @@ func (gj *guardJob) ID() string {
...
+			retries := 3
+			for retries > 0 {
+				_, err := c.Reimage(ctx, rg, vmssName, *instance.InstanceID, nil)
+				if err != nil {
+					gj.logger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("Error reimaging instance %s: %s", *instance.Name, err.Error()))
+					if retries == 0 {
+						return false, microerror.Mask(err)
+					}
+
+					retries = retries - 1
```
whites11, line 125: This was giving errors to @fiunchinho right?

fiunchinho: Yes, I fixed it deleting the `break` below. I believe it's not necessary?

tuommaki:
Going with this now. e2e failed on reaching expected number of nodes. Despite of that, I already span several clusters today with this so I'm confident enough to go with this. We still need to fix scaling check and test one more round so this is good.

PR #748, opened 2020-04-01
Only roll nodes when they aren't in sync with provider operator

tuommaki:
It is difficult to detect different change scenarious with Azure APIs /
infrastructure and that has invoked full upgrade process e.g. when creating
cluster or scaling despite of many checks that try to test for such situation.

On principal rolling of node should not be needed other than when clusters are
upgrading or when certificates need renewal. This change introduces first of
these so upgrade process only executes through when there are nodes with old
azure-operator version labels. Similar case for expiring certificates should be
added later if needed.

### DIFF
--- a/service/controller/resource/instance/create_cluster_upgrade_requirement_check.go
+++ b/service/controller/resource/instance/create_cluster_upgrade_requirement_check.go
@@ -3,9 +3,14 @@ package instance
 import (
 	"context"
 
+	"github.com/coreos/go-semver/semver"
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/giantswarm/azure-operator/pkg/label"
+	"github.com/giantswarm/azure-operator/pkg/project"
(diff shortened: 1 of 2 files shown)


service/controller/resource/instance/create_cluster_upgrade_requirement_check.go
```diff
@@ -46,17 +50,38 @@ func (r *Resource) isClusterCreating(cr providerv1alpha1.AzureConfig) bool {
 	return false
 }
 
-func (r *Resource) isClusterScaling(ctx context.Context, cr providerv1alpha1.AzureConfig) (bool, error) {
-	c, err := r.getScaleSetsClient(ctx)
+// anyNodesOutOfDate iterates over all nodes in tenant cluster and finds
+// corresponding azure-operator version from node labels. If node doesn't have
+// this label or was created with older version than currently reconciling one,
+// then this function returns true. Otherwise (including on error) false.
+func (r *Resource) anyNodesOutOfDate(ctx context.Context, cr providerv1alpha1.AzureConfig) (bool, error) {
```
nprokopic, line 57: CircleCI build is failing due to unused parameter `cr`.

tuommaki: Ah, right. Of course. One sec.

tuommaki:
Tested. Works superb. Going.

PR #756, opened 2020-04-03
Use flatcar as distro

MarcelMue:
Switching to flatcar together with using the `containerlinux` version from releases.


service/controller/resource/instance/deployment.go
```diff
@@ -38,6 +38,16 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
 		workerBlobName,
 	}
 
+	var distroVersion string
+	{
+		for _, component := range cc.Release.Components {
+			if component.Name == key.ContainerLinuxComponentName {
+				distroVersion = component.Version
```
MarcelMue, line 45: Getting the version out of the components.

service/controller/key/key.go
```diff
@@ -52,7 +52,7 @@ const (
 	CertificateEncryptionKeyName   = "encryptionkey"
 	CertificateEncryptionIVName    = "encryptioniv"
 
-	CoreosVersion = "2303.4.0"
+	ContainerLinuxComponentName = "containerlinux"
```
omegas27, line 55: Maybe for the future but should that key not be operator system?

service/controller/resource/instance/types.go
```diff
@@ -35,24 +33,24 @@ type nodeOSImage struct {
...
 
 // newNodeOSImage provides OS information for Container Linux
-func newNodeOSImageCoreOS() nodeOSImage {
+func newNodeOSImage(distroVersion string) nodeOSImage {
 	return nodeOSImage{
-		Offer:     "CoreOS",
-		Publisher: "CoreOS",
-		SKU:       "Stable",
-		Version:   key.CoreosVersion,
+		Offer:     "flatcar-container-linux",
```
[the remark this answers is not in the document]
omegas27: Could we add metadata in the release CR so the publisher, sku and offer can be extracted from the CR instead of hardcoded?

integration/setup/common.go
```diff
@@ -132,7 +132,7 @@ func common(ctx context.Context, config Config) error {
 					},
 					{
 						Name:    "containerlinux",
```
omegas27, line 134: is flatcar still considered containerlinux?

service/controller/resource/instance/deployment.go
omegas27: Why not call it osversion?

service/controller/resource/instance/types.go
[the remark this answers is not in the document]
MarcelMue: I thought about this but I don't want to include provider specific information in the releases somehow. We would then also need to do this for AWS and so on.

integration/setup/common.go
MarcelMue: AFAIK yes - it's just a fork/continuation.

service/controller/key/key.go
MarcelMue: True when we move away from `containerlinux` eventually - I didn't want to overdo it here :D

service/controller/resource/instance/deployment.go
MarcelMue: @stone-z called it the same in `kvm-operator ` AFAIK - so I wanted to be consistent. Maybe we change it everywhere?
I have no strong opinion here.

service/controller/resource/instance/types.go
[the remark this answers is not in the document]
omegas27: I understand, that is why I was thinking of some sort of metadata solution but maybe we can remove this hardcoded solution in another way

service/controller/resource/instance/deployment.go
omegas27: If that is ok with you, sure :) I just wanted to avoid discrepancies between OS and Distro

PR #757, opened 2020-04-07
New node detection

whites11:
See https://github.com/giantswarm/giantswarm/issues/9921

We try to use a different approach to distinguish between old nodes and new nodes during an upgrade.
Instead of relying on the `Latest model` field provided by azure, we use a tag we set to the VMSS containing the release version.
We hope that the tag is propagated to the instances of the VMSS only when they are upgraded to the new model.

### DIFF
--- a/service/controller/resource/instance/create_cordon_old_workers.go
+++ b/service/controller/resource/instance/create_cordon_old_workers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/coreos/go-semver/semver"
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/microerror"
 	corev1 "k8s.io/api/core/v1"
@@ -12,6 +13,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/giantswarm/azure-operator/pkg/label"
+	"github.com/giantswarm/azure-operator/pkg/project"
 	"github.com/giantswarm/azure-operator/service/controller/controllercontext"
 	"github.com/giantswarm/azure-operator/service/controller/key"
 	"github.com/giantswarm/azure-operator/service/controller/resource/instance/internal/state"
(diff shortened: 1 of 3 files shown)


whites11:
The tag gets propagated from the VMSS to the instance immediately.
We ended up using the `azure-operator.giantswarm.io/version` label from the kubernetes node resource.

whites11:
e2e tests are failing because github pages is currently down

service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -131,22 +134,31 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
...
+
+		v, exists := n.GetLabels()[label.OperatorVersion]
+		if !exists {
+			// Label does not exist, we consider node as old.
+			oldNodes = append(oldNodes, n)
+			continue
+		}
+
+		nodeVersion := semver.New(v)
+		if nodeVersion.LessThan(*myVersion) {
```
tuommaki, line 158: I've been thinking about this. I think I used `LessThan()` as well, but I wonder if just difference (`nodeVersion != myVersion`) would be enough or better? Our code in general doesn't care that much about semantics of versions, but rather that all nodes would be same version.

service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -131,22 +134,31 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
...
 			// found from node list.
-			if found {
-				newNodes = append(newNodes, n)
-			}
+			continue
+		}
+
+		v, exists := n.GetLabels()[label.OperatorVersion]
+		if !exists {
+			// Label does not exist, we consider node as old.
```
tuommaki, line 152: Actually, this might be wrong. I think old nodes always have the operator label (we added it quite a while back) but new nodes don't necessarily have. There's time period when the node has registered but systemd unit that applies labels haven't run yet. You can observe this during upgrade by seeing new node without assigned role in the node list.

whites11: So what do you suggest to do? Consider the node not new nor old as long as it has the label?
In theory, the azure operator will wait to cordon until he finds the number of old workers equals the number of new workers, so not counting one worker means we would wait for the following reconciliation loop.

service/controller/resource/instance/create_cordon_old_workers.go
whites11: That would make sense if we supported downgrading of nodes. But we don't so I guess there's no difference in using `!=` or `LessThan`, right?

service/controller/resource/instance/create_cordon_old_workers.go
whites11: I made a change to ignore the node from the count if it doesn't have the version label

service/controller/resource/instance/create_cordon_old_workers.go
tuommaki: Yes, both work so no difference. All good.

service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -131,22 +134,32 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
...
-				newNodes = append(newNodes, n)
-			}
+			continue
+		}
+
+		v, exists := n.GetLabels()[label.OperatorVersion]
+		if !exists {
+			// Label does not exist, this normally happens when a new node is coming up but did not finish
+			// its kubernetes bootstrap yet and thus doesn't have all the needed labels.
+			// We'll ignore this node for now and wait for it to bootstrap correctly.
```
tuommaki, line 154: Please add it to `newNodes` because otherwise this function doesn't work as expected (`len(input nodes) > (len(oldNodes) + len(newNodes))`) and that might cause problems one day when some maintenance changes are done.

whites11: This is the check we do
```
if len(newNodes) < len(oldNodes) {
```
The only problem I see with the current "ignore" approach is that the upgrade could be stuck if we remove the label in the future for any reason.

tuommaki: Sure, it works now, but theoretically there could be some point in time in the future where the client-side logic expects sum of return values to match input length.

Nevertheless, for now this works correctly.

PR #759, opened 2020-04-07
Add simple gatekeeper to provide backoff-like functionality

tuommaki:
This is super simple API to make it easy/readable to hold off from making
additional request in case threshold has been hit.

Towards https://github.com/giantswarm/giantswarm/issues/9912


pkg/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,19 @@
+package gatekeeper
```
fiunchinho, line 1: looks fine, but I'm not a big fan of the name. It's not clear what it does, and it can be confused with OPA gatekeeper. What about `backpressure` or something along those lines?

tuommaki: I had the same problem as well, but couldn't come up with any good name and wanted to just proceed with the implementation in general.

I'll go with `backpressure`. I don't fully buy in that one either, but I just can't come up with anything reasonable. At least that's not confused with that OPA thingie :man_shrugging:

tuommaki:
Fishing for LGTM

PR #760, opened 2020-04-07
Add RateLimitCircuitBreaker for go-autorest

tuommaki:
This is a simple `autorest.SendDecorator` implementation to hold back when
Azure API responds with HTTP 429 Too Many Requests.

Towards https://github.com/giantswarm/giantswarm/issues/9912


tuommaki:
CI should get green here once I merge the two required PRs:
- https://github.com/giantswarm/azure-operator/pull/758
- https://github.com/giantswarm/azure-operator/pull/759

client/azure_client_set.go
```diff
@@ -152,6 +147,7 @@ func newDeploymentsClient(config *clientConfig) (*resources.DeploymentsClient, e
 	c := resources.NewDeploymentsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
 	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
 	c.RetryAttempts = 1
+	senddecorator.ConfigureClient(&gatekeeper.Gatekeeper{}, &c)
```
fiunchinho, line 150: Just want to make sure: the Azure rate limit is per client? If that's the case, it makes sense to create a gatekeeper instance for every client, otherwise sharing a single instance would make more sense.

client/senddecorator/ratelimit_circuitbreaker.go
```diff
@@ -17,6 +17,11 @@ const (
 	defaultWaitAfterTooManyRequests = 6 * time.Minute
 )
 
+func init() {
+	// ONE DOES NOT SIMPLY RETRY ON HTTP 429.
+	autorest.StatusCodesForRetry = removeElementFromSlice(autorest.StatusCodesForRetry, http.StatusTooManyRequests)
+}
```
fiunchinho, line 23: I know this is personal, but I still it's _magical_ to use `init()`. What do you say about moving `autorest.StatusCodesForRetry = removeElementFromSlice(autorest.StatusCodesForRetry, http.StatusTooManyRequests)` to the first line of the `RateLimitCircuitBreaker` function, right above the return?

client/azure_client_set.go
tuommaki: No, rate limit is per subscription and ideally the `gatekeeper` should be shared. That's why I raised it to `ConfigureClient()` level. I still want to figure though how to best wire it from the highest level.

The problem here is that different services have different rate-limit quotas but there's only single `client.NewAzureClientset()` call in `resource_set.go`.

client/senddecorator/ratelimit_circuitbreaker.go
tuommaki: My reasoning for `init()` is that it's guaranteed to run once before actual business logic whenever this package is imported. It's a global variable in library and changing that repeatedly on function call feels semantically wrong to me. `init()` makes more sense since that knob should be turned once and be left at that.

fiunchinho: okok

tuommaki:
Ok. This is good now, but I'll hold off until we get the patch release out.

PR #761, opened 2020-04-08
Add pre-commit hook config file

fiunchinho:
Added pre-commit hook configuration to run some checks locally before comitting to the repository

image


README.md
```diff
@@ -64,6 +64,14 @@ Follow [this guide][examples-local].
 See [CONTRIBUTING](CONTRIBUTING.md) for details on submitting patches, the
 contribution workflow as well as reporting bugs.
 
+## Pre-commit hook
+
+- [Install pre-commit](https://pre-commit.com/#installation).
+- [Install the git hook scripts](https://pre-commit.com/#3-install-the-git-hook-scripts)
```
axbarsan, line 70: Can we automate this with a `make` command?

fiunchinho: You mean the installation of the hooks? Like having a `Makefile` with a target called `precommit-hooks` that installs the hooks?
Installing the framework itself it's trickier, because it's platform dependent maybe?

axbarsan: I now saw that... We could implement a platform check and based on that check run a different command, or even have different `make` commands for different platforms, but I think that's just over-complicating everything

PR #762, opened 2020-04-08
add azure operator version to ARM parameters

whites11:
to force an upgrade even when the arm template is not changed.

### DIFF
--- a/service/controller/resource/instance/template/vmss.json
+++ b/service/controller/resource/instance/template/vmss.json
@@ -5,6 +5,9 @@
     "location":{
       "type":"string"
     },
+    "azureOperatorVersion":{
+      "type":"string"
+    },
     "GiantSwarmTags":{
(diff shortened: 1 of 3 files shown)


service/controller/resource/instance/template/vmss.json
```diff
@@ -111,7 +114,8 @@
       "tags":{
         "provider":"[toUpper(parameters('GiantSwarmTags').provider)]",
         "cluster-autoscaler-enabled":"true",
-        "cluster-autoscaler-name":"[parameters('vmssName')]"
+        "cluster-autoscaler-name":"[parameters('vmssName')]",
+        "gs-azure-operator.giantswarm.io-version":"[parameters('azureOperatorVersion')]"
```
whites11, line 118: tags beginning with "azure" are not allowed.

PR #765, opened 2020-04-08
Add Metrics SendDecorator for Azure SDK clients

tuommaki:
Add send decorator to Azure SDK clients that records API call's latency and
status code.

### DIFF
--- a/client/azure_client_set.go
+++ b/client/azure_client_set.go
@@ -11,11 +11,193 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/giantswarm/exporterkit/collector"
 	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/giantswarm/azure-operator/client/senddecorator"
 	"github.com/giantswarm/azure-operator/pkg/backpressure"
 )
+
+var (
+	// Prometheus metrics for clients
+	deploymentsClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_deployments",
+			Help:       "HTTP metrics for Azure SDK Deployments client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	groupsClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_groups",
+			Help:       "HTTP metrics for Azure SDK ARM Resource Groups client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	dnsRecordSetsClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_dnsrecordsets",
+			Help:       "HTTP metrics for Azure SDK DNS Record Sets client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	dnsZonesClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_dnszones",
+			Help:       "HTTP metrics for Azure SDK DNS Zones client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	interfacesClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_interfaces",
+			Help:       "HTTP metrics for Azure SDK Virtual Network Interfaces client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	storageAccountsClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_storageaccounts",
+			Help:       "HTTP metrics for Azure SDK Storage Accounts client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	usageClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_usage",
+			Help:       "HTTP metrics for Azure SDK Usage client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	virtualNetworkMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_virtualnetwork",
+			Help:       "HTTP metrics for Azure SDK Virtual Network client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+
+	virtualNetworkGatewaysMetricsDesc = prometheus.NewSummaryVec(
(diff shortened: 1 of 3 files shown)


client/azure_client_set.go
```diff
@@ -11,12 +11,194 @@ import (
...
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"status_code"},
+	)
+)
+
+func init() {
+	// Register prometheus metrics.
+	err := prometheus.Register(deploymentsClientMetricsDesc)
+	if err != nil && !collector.IsAlreadyRegisteredError(err) {
```
tuommaki, line 145: `IsAlreadyRegisteredError()` is needed for unit tests.

tuommaki:
This is my initial take on decorating Azure SDK clients with metrics collection. I'm not very happy with the approach I took here, but so far I was unable to think about anything better.

There are few challenges here:
- New clients are created on `initCtx()`. Theoretically they could be cached, but I wonder if that's sane.
- Collector from `service/collector` must not be wired into clients because dependencies go already other way round and I feel it would be also semantically wrong.
- Creating some intermediate component that collects arbitrary metrics, caches them and lets `Collector` implementation to collect them feels unnecessarily redundant. 

If you feel that we can continue with this approach, I'll then further add at least `subscription_id` label and possibly some other as well, but I welcome alternative design proposals here.

tuommaki:
Ping @xh3b4sd @kopiczko @JosephSalisbury for alternatives ideas.

client/azure_client_set.go
```diff
@@ -11,12 +11,194 @@ import (
...
 	"github.com/giantswarm/azure-operator/pkg/backpressure"
 )
 
+var (
+	// Prometheus metrics for clients
+	deploymentsClientMetricsDesc = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "azure_sdk_client_deployments",
+			Help:       "HTTP metrics for Azure SDK Deployments client.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
```
xh3b4sd, line 28: I need an ELI5 on that one. What is the mapping achieving? I can't really interpret it. Godoc of prometheus library was not really helpful on a brief glance.

client/azure_client_set.go
xh3b4sd: I thought this is what we have the collector set thingy for. So that you do not have to do the crappy init magic. Why doing in that way here?

fiunchinho: why don't we register the metrics in the `MetricsCollector` factory method so that they only get registered once?

client/azure_client_set.go
```diff
@@ -11,12 +11,194 @@ import (
...
 	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/giantswarm/azure-operator/client/senddecorator"
 	"github.com/giantswarm/azure-operator/pkg/backpressure"
 )
 
+var (
+	// Prometheus metrics for clients
+	deploymentsClientMetricsDesc = prometheus.NewSummaryVec(
```
fiunchinho, line 24: noob question: are there advantages to use different metrics instead of a single metric with different labels? Like having a single metric but using the client name as label?

tuommaki: I'm not fully sure. Potentially both could work. Metric per client just made logically sense.

client/azure_client_set.go
tuommaki: From https://godoc.org/github.com/prometheus/client_golang/prometheus#SummaryOpts:
```
    // Objectives defines the quantile rank estimates with their respective
    // absolute error. If Objectives[q] = e, then the value reported for q
    // will be the φ-quantile value for some φ between q-e and q+e.  The
    // default value is an empty map, resulting in a summary without
    // quantiles.
```

client/azure_client_set.go
tuommaki: > I thought this is what we have the collector set thingy for. So that you do not have to do the crappy init magic. Why doing in that way here?

`client` package is used from `service/collector` so we can't import anything from there (import loop).

> why don't we register the metrics in the MetricsCollector factory method so that they only get registered once?

Because currently `MetricsCollector` gets called on every reconciliation loop because clients are constructed in `initCtx`.

xh3b4sd: Sounds like a use case to extend `exporterkit` maybe.

client/azure_client_set.go
xh3b4sd: I found that already, and then came here to ask for clarification. An example and reasoning would help. Neither the numbers nor the map tell me any story. I don't follow why they are selected as they are now.

client/azure_client_set.go
JosephSalisbury: I would have a label here, per client. Makes the alerting easier.

JosephSalisbury: I would use a label here, makes the alert simpler.

client/azure_client_set.go
JosephSalisbury: Why a summary? I'd have used a histogram

client/senddecorator/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+		return autorest.SenderFunc(func(r *http.Request) (*http.Response, error) {
+
+			start := time.Now()
+
+			// Pass the request to next SendDecorator.
+			resp, err := s.Do(r)
+
+			elapsed := time.Since(start)
+
+			// Record latency as milliseconds.
```
JosephSalisbury, line 23: Let's use seconds as the unit please, standards recommend SI base units (see https://prometheus.io/docs/practices/naming/)

client/azure_client_set.go
tuommaki: Ack

client/azure_client_set.go
tuommaki: > An example and reasoning would help. Neither the numbers nor the map tell me any story. I don't follow why they are selected as they are now.

50%, 90% and 99% percentiles are common defaults for these kind (like request latencies) of metrics. You can also grep for `quantile` from `g8s-grafana` - see e.g. https://github.com/giantswarm/g8s-grafana/blob/master/helm/g8s-grafana-chart/dashboards/shared/dns.json#L274-L294

> Why a summary? I'd have used a histogram.

Yeah, histogram would work as well. I guess we usually compute the quantiles on server-side. I was just looking into these metric types and since this readily calculates the quantile on client-side, it sort of made sense to me to do it that way.

client/senddecorator/metrics.go
tuommaki: Hmm... That's what documentation seems to recommend. Feels super counter-intuitive when expected values are in granularity of 10-100ms. I can only imagine the amount of error when doing statistics with huge number of very small floats. I guess it still works in practice?

client/azure_client_set.go
JosephSalisbury: The fact that summaries calculates quantiles on client-side is why I tend towards histograms :D

client/senddecorator/metrics.go
JosephSalisbury: I think in most cases it doesn't matter, and being able to treat all 'times' as the same pays off a lot

client/azure_client_set.go
kopiczko: You can define descs in collector package, init then there and import them here.

client/azure_client_set.go
```diff
@@ -11,12 +11,194 @@ import (
...
+
+func init() {
+	// Register prometheus metrics.
+	err := prometheus.Register(deploymentsClientMetricsDesc)
+	if err != nil && !collector.IsAlreadyRegisteredError(err) {
+		panic(err)
+	}
+	err = prometheus.Register(groupsClientMetricsDesc)
+	if err != nil && !collector.IsAlreadyRegisteredError(err) {
+		panic(err)
```
kopiczko, line 150: Let’s add microerror.JSON/microerror.Stack here.

client/senddecorator/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"net/http"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// MetricsCollector returns decorator that collects API call metrics.
+func MetricsCollector(m *prometheus.SummaryVec) autorest.SendDecorator {
```
kopiczko, line 12: Shouldn’t that be private?

client/azure_client_set.go
xh3b4sd: What do you mean with `microerror.Stack`? That does not exist (anymore).

tuommaki:
Thanks for all review comments here. I'll take them into account when I work on this further. We are currently separating metrics collection in `azure-operator` to external service and this has consequences on this client instrumentation as well.

As written above, this was Hackathon project that I tried and as such, I gained learning from here, but will close now this specific PR and come up with a better one in the future.

client/azure_client_set.go
kopiczko: I think this codebase still uses v0.1.0

PR #766, opened 2020-04-08
Get rid of LastModelApplied checks

whites11:
This PR removes all usages of the LastModelApplied field of the VMSS Instance type because it proved to be unreliable, at least on germany west central.

There is only one usage for it at the moment (in the update phase) but I guess we can't remove it from there.

### DIFF
--- a/service/controller/resource/instance/create_cordon_old_workers.go
+++ b/service/controller/resource/instance/create_cordon_old_workers.go
@@ -165,3 +165,54 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
 
 	return
 }
+
+func (r *Resource) getK8sWorkerNodeForInstance(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance compute.VirtualMachineScaleSetVM) (*corev1.Node, error) {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	name := key.WorkerInstanceName(customObject, *instance.InstanceID)
+
+	nodeList, err := cc.Client.TenantCluster.K8s.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	nodes := nodeList.Items
+
+	for _, n := range nodes {
+		if n.GetName() == name {
+			return &n, nil
+		}
+	}
+
+	// Node related to this instance was not found.
+	return nil, nil
+}
+
+func (r *Resource) isWorkerInstanceFromPreviousRelease(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance compute.VirtualMachineScaleSetVM) (*bool, error) {
+	t := true
+	f := false
+
+	n, err := r.getK8sWorkerNodeForInstance(ctx, customObject, instance)
+	if err != nil {
+		return nil, err
+	}
+
+	myVersion := semver.New(project.Version())
+
+	v, exists := n.GetLabels()[label.OperatorVersion]
+	if !exists {
+		// Label does not exist, this normally happens when a new node is coming up but did not finish
+		// its kubernetes bootstrap yet and thus doesn't have all the needed labels.
+		// We'll ignore this node for now and wait for it to bootstrap correctly.
+		return nil, nil
+	}
+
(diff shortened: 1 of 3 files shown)


service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -165,3 +165,59 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
 
 	return
 }
+
+func (r *Resource) getK8sWorkerNodeForInstance(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance compute.VirtualMachineScaleSetVM) (*corev1.Node, error) {
```
tuommaki, line 169: Can I propose to move these functions to e.g. `worker.go`? Now they are in one of the state transition files and then used in the others. When looking for code it's confusing even though editor / IDE helps finding it. :slightly_smiling_face:

service/controller/resource/instance/create_drain_old_worker_nodes.go
```diff
@@ -53,7 +53,13 @@ func (r *Resource) drainOldWorkerNodesTransition(ctx context.Context, obj interf
 
 	var nodesPendingDraining int
 	for _, i := range allWorkerInstances {
-		if *i.LatestModelApplied {
+		old, err := r.isWorkerInstanceFromPreviousRelease(ctx, cr, i)
```
tuommaki, line 56: I'd propose to list all nodes in the beginning and then either sort them to old and new or alternatively create a function that finds the given one from already fetched nodes. Now in here this lists all nodes from k8s api for every instance that exists in Azure VMSS. Even for small cluster it's already quite a lot of data to transfer from etcd via apiserver to client and in bigger clusters it gets exponentially worse.

service/controller/resource/instance/create_drain_old_worker_nodes.go
```diff
@@ -53,7 +53,13 @@ func (r *Resource) drainOldWorkerNodesTransition(ctx context.Context, obj interf
...
+		if err != nil {
+			return DeploymentUninitialized, nil
```
nprokopic, line 58: Just ignore error and return reconciliation back to the start when an error happens here? (just to check if it was intended)

service/controller/resource/instance/create_cordon_old_workers.go
nprokopic: What about having _worker node_ <-> _VMSS instance_ mapping saved (in status?) so we can look it up when we need it?
(maybe too much/unnecessary for this fix)

service/controller/resource/instance/create_drain_old_worker_nodes.go
nprokopic: What about having worker node <-> VMSS instance mapping saved (in status?) so we can look it up when we need it?
(maybe too much/unnecessary for this fix)

tuommaki: The node list in status wouldn't be guaranteed to be up-to-date-enough, I think. Consider that upgrade is started, `instance` resource proceeds to some extend, then Azure works on the background while CP gets disconnected from it. I'm not sure if that would work correctly in all cases.

service/controller/resource/instance/create_drain_old_worker_nodes.go
whites11: probably wrong and coming from a copy paste from another part. It was late guys :(

whites11: but we can't postpone IMHO, let's merge.

service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -165,3 +165,59 @@ func sortNodesByTenantVMState(nodes []corev1.Node, instances []compute.VirtualMa
...
+		}
+	}
+
+	// Node related to this instance was not found.
+	return nil, nil
+}
+
+func (r *Resource) isWorkerInstanceFromPreviousRelease(ctx context.Context, customObject providerv1alpha1.AzureConfig, instance compute.VirtualMachineScaleSetVM) (*bool, error) {
+	t := true
+	f := false
```
fiunchinho, line 195: What's with the boolean pointers? :smile:  Just returning true or false wasn't enough?

PR #770, opened 2020-04-09
Default parameters so it's easy to run the operator locally

fiunchinho:
I'm trying to make it easier to run the operator locally, to make testing easier.
This shouldn't change the operator behavior: it's just setting defaults in case parameters are not passed when starting the operator.


main.go
```diff
@@ -126,21 +126,21 @@ func mainError() error {
...
 	daemonCommand.PersistentFlags().String(f.Service.Azure.ClientSecret, "", "Secret of the Active Directory Service Principal.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.EnvironmentName, "AZUREPUBLICCLOUD", "Azure Cloud Environment identifier.")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.CIDR, "", "CIDR of the host cluster virtual network used to create a peering.")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.ResourceGroup, "", "Host cluster resource group name.")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.VirtualNetwork, "", "Host cluster virtual network name.")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.VirtualNetworkGateway, "", "Host cluster virtual network gateway name.")
-	daemonCommand.PersistentFlags().Bool(f.Service.Azure.MSI.Enabled, true, "Whether to enabled Managed Service Identity (MSI).")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.Location, "", "Location of the host and guset clusters.")
+	daemonCommand.PersistentFlags().String(f.Service.Azure.Location, "westeurope", "Location of the host and guset clusters.")
```
fiunchinho, line 129: New default added to `westeurope`

main.go
```diff
@@ -126,21 +126,21 @@ func mainError() error {
...
 	daemonCommand.PersistentFlags().String(f.Service.Azure.SubscriptionID, "", "ID of the Azure Subscription.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.TenantID, "", "ID of the Active Directory Tenant.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Azure.MSI.Enabled, true, "Whether to enabled Managed Service Identity (MSI).")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.Template.URI.Version, defaultTemplateVersion, "URI version for ARM template links.")
 	daemonCommand.PersistentFlags().Int(f.Service.Azure.VMSSCheckWorkers, 5, "Number of workers in VMSS check worker pool.")
+	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.CIDR, "10.0.0.0/16", "CIDR of the host cluster virtual network used to create a peering.")
```
fiunchinho, line 135: New default added to `10.0.0.0/16`. This is different during e2e tests.

main.go
```diff
@@ -126,21 +126,21 @@ func mainError() error {
...
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Name, "", "Installation name for tagging Azure resources.")
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.ClientID, "", "OIDC authorization provider ClientID.")
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.IssuerURL, "", "OIDC authorization provider IssuerURL.")
-	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.UsernameClaim, "", "OIDC authorization provider UsernameClaim.")
-	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.GroupsClaim, "", "OIDC authorization provider GroupsClaim.")
+	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.UsernameClaim, "email", "OIDC authorization provider UsernameClaim.")
+	daemonCommand.PersistentFlags().String(f.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.GroupsClaim, "groups", "OIDC authorization provider GroupsClaim.")
```
fiunchinho, line 143: New default values added
- UsernameClaim: "email"
- GroupsClaim: "groups"

I don't think we use different values anywhere.

service/service.go
```diff
@@ -73,6 +74,21 @@ func New(config Config) (*Service, error) {
...
+	virtualNetwork := config.Viper.GetString(config.Flag.Service.Azure.HostCluster.VirtualNetwork)
+	if virtualNetwork == "" {
+		virtualNetwork = resourceGroup
+	}
+
+	virtualNetworkGateway := config.Viper.GetString(config.Flag.Service.Azure.HostCluster.VirtualNetworkGateway)
+	if virtualNetworkGateway == "" {
+		virtualNetworkGateway = fmt.Sprintf("%s-%s", resourceGroup, "vpn-gateway")
+	}
+
```
fiunchinho, line 91: In case the resource group parameter is empty, we default it to the installation name.
The virtual network resource in that resource group is also named after the installation name if I'm not mistaken.

The virtual network gateway is called ${installation_name}-vpn-gateway, so we default to that.

service/service.go
```diff
@@ -102,9 +118,19 @@ func New(config Config) (*Service, error) {
...
 
+	oidcClientId := config.Viper.GetString(config.Flag.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.ClientID)
+	if oidcClientId == "" {
+		oidcClientId = config.Viper.GetString(config.Flag.Service.Azure.ClientID)
+	}
+
+	oidcIssuerUrl := config.Viper.GetString(config.Flag.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.IssuerURL)
+	if oidcIssuerUrl == "" {
+		oidcIssuerUrl = fmt.Sprintf("https://login.microsoftonline.com/%s/v2.0", config.Viper.GetString(config.Flag.Service.Azure.TenantID))
+	}
```
fiunchinho, line 129: This is the part I'm less sure about. I'm configuring OIDC using the same clientid and tenantid than the operator is using. That's how we do it during tests. Does it make sense?

T-Kukawka: Yeah we can do that, i mean we are not authenticating with OIDC anyway

T-Kukawka: I mean we are not testing here that we authenticate but  that the params are passed correctly right?

fiunchinho: Yes. We don't change anything if params are passed, which is the case in our installations. We just change how it works when params are not passed. In that case I use the same credentials than the operator is using.

main.go
tuommaki: I'm not sure we should default OIDC values to anything? Didn't check, but I'd imagine that if these are empty strings, then OIDC configuration is omitted. Otherwise the values are configured into ignition template rendering and that would most probably be wrong.

fiunchinho: [I think you are right](https://github.com/giantswarm/azure-operator/blob/3bf42310288f14e730ea63b641bfde98b20b25f7/service/controller/cloudconfig/master_template.go#L28-L39). I'll remove the defaults for OIDC then.

PR #773, opened 2020-04-14
Cancel full reconciliation if ARM deployment for deployment resource has not finished

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9862

### DIFF
--- a/service/controller/resource/deployment/checksum.go
+++ b/service/controller/resource/deployment/checksum.go
@@ -1,0 +1,55 @@
+package deployment
+
+import (
+	"crypto/sha256"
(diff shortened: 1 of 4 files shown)


fiunchinho:
With this change we end up with a `kubernetes` load balancer which has none health probes and none load balancing rules.
I still don't know which component creates those, so not sure why this fails.

fiunchinho:
Trying now executing the `dns` resource before the `instance` resource. Otherwise the `instance` resource was canceling the reconciliation before reaching the `dns` resource and the cluster was never marked as ready.

PR #776, opened 2020-04-15
Use release.Revision in Helm chart for Helm 3 support

rossf7:
Towards giantswarm/giantswarm#10098


rossf7:
I was a bit stealthy here. This is needed because `Release.Time` and `Release.Revision` were removed from Helm 3. Release.Revision got added back in https://github.com/helm/helm/pull/6688.

The purpose of the annotation is so that the operator pod is replaced if another resource changes. So we can switch to Release.Revision. I'll merge once the CI is green.

PR #780, opened 2020-04-17
CHeck for VMSS rates before scaling up worker nodes

whites11:
I moved the VMSS rate limit check from inside the guard job to just before the upscaling of worker vmss.
The most critical moment during an upgrade in terms of VMSS API calls is when a new node is added.
That's the moment when we need to check for the API to be in a safe state.

### DIFF
--- a/service/controller/resource/instance/create_scale_workers.go
+++ b/service/controller/resource/instance/create_scale_workers.go
@@ -39,6 +39,12 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
 
 	allReady, err := vmsscheck.InstancesAreRunning(ctx, r.logger, key.ResourceGroupName(cr), key.WorkerVMSSName(cr))
 	if err != nil {
+		// VMSS rate limits are not safe, let's wait for next reconciliation loop.
+		if vmsscheck.IsVMSSUnsafeError(err) {
+			return ScaleUpWorkerVMSS, nil
+		}
+
+		// Unexpected error.
 		return "", microerror.Mask(err)
 	}
 	// Not all workers are Running in Azure, wait for next reconciliation loop.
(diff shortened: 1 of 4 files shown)


service/controller/resource/instance/internal/vmsscheck/vmss.go
```diff
@@ -48,6 +73,16 @@ func InstancesAreRunning(ctx context.Context, logger micrologger.Logger, rg stri
 		}
 	}
 
+	if allSucceeded {
+		// All instances are succeeded, let's check the VMSS rate is safe.
+		// If current remaining API calls are less than the desired threshold, we don't proceed.
```
tuommaki, line 78: I'm not very fond of this. This gets easily forgotten and hidden and is not obvious from function name nor godoc comment.

I'd propose to introduce an error & matcher and then handle the error correspondingly. To me it also seems that it might make sense to do the same thing for few other functions such as `r.getInstancesCount()`.

That way the situation is lot easier to understand when following the higher level control flow and it doesn't come as a surprise that if VMSS rate limit quotas are low that suddenly the state transitioning doesn't take place even though all instances are clearly `Succeeded` in portal.

whites11: I agree with the custom error idea, but I don't get why we should do that for the `r.getInstancesCount` function. That is a read only function that does not cause the VMSS rate to increase at all.

service/controller/resource/instance/create_scale_workers.go
```diff
@@ -39,6 +39,12 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
 
 	allReady, err := vmsscheck.InstancesAreRunning(ctx, r.logger, key.ResourceGroupName(cr), key.WorkerVMSSName(cr))
 	if err != nil {
```
fiunchinho, line 41: I think we use [a different code style](https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors)

```
if vmsscheck.IsVMSSUnsafeError(err) {
	return ScaleUpWorkerVMSS, nil
} else if err != nil {
}
```

whites11: thanks, fixed

PR #783, opened 2020-04-17
Version bump to v3.0.7 for release 11.2.5

tuommaki:


nprokopic:
Few questions:
- It should be for 11.2.6 now?
- When is the release planned? We should do the bumping after the testing and at the moment we are releasing it, right?

tuommaki:
> It should be for 11.2.6 now?

Yes

> When is the release planned? We should do the bumping after the testing and at the moment we are releasing it, right?

Yeah, I think this PR should probably get dropped, right? Then `master` can be tested and eventually released.

nprokopic:
Yes, I'd agree to drop this PR, and go with version bump after testing.

PR #784, opened 2020-04-17
Remove linter ignore keywords

fiunchinho:
Updates micrologger and removes `// nolint: errcheck` from logging code


Gopkg.toml
```diff
@@ -94,7 +94,7 @@ required = [
 
 [[override]]
   name = "github.com/giantswarm/micrologger"
-  version = "~0.1.0"
+  version = "~0.3.0"
```
fiunchinho, line 97: This version doesn't return an error when logging.

PR #785, opened 2020-04-21
Use full template content instead of links

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/10309

We use [ARM linked templates](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates), which is a way to break up a big template into several pieces. There are two ways of doing this:

- using links to public templates
- using nested templates, with embedded content

We are currently using the first method, linking to Github which has a couple of problems, mainly:

- The template won't be read if Github is down
- The operator can't be tested locally, since it depends on Github links.

With these changes we use the second method, using embedded templates which are built locally.


fiunchinho:
Tests passed. I'm going to merge master now and see what happens.

.gitignore
```diff
@@ -9,3 +9,4 @@ TODO
 !/examples/local/*.tmpl.yaml
 
 integration/test/**/*-e2e
+integration/test/**/logs/*.txt
```
fiunchinho, line 12: running tests locally generate logs on that folder.

pkg/checksum/checksum_test.go
```diff
@@ -18,55 +16,24 @@ import (
...
-			expectedChecksum:    "",
-			errorMatcher:        IsNilTemplateLinkError,
-		},
-		{
-			name:                "case 2: Error downloading template from external URI",
-			templateLinkPresent: true,
-			expectedChecksum:    "",
-			statusCode:          http.StatusInternalServerError,
-			responseBody:        `{"error": "500 - Internal server error"}`,
-			errorMatcher:        IsUnableToGetTemplateError,
```
fiunchinho, line 47: No need for these test cases anymore

service/controller/resource/deployment/templates.go
```diff
@@ -0,0 +1,1523 @@
+package deployment
```
fiunchinho, line 1: There is no built-in way in Go to read `resources` in the same package. So I decided to inline the templates using constants.

service/controller/resource/deployment/templates.go
```diff
@@ -0,0 +1,1523 @@
...
+func getARMTemplate() (map[string]interface{}, error) {
+	contents := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(getARMTemplateAsString()), &contents); err != nil {
+		return nil, err
+	}
+	return contents, nil
+}
+
+func getARMTemplateAsString() string {
+	return fmt.Sprintf(main, securityGroups, routeTable, virtualNetwork, publicLoadBalancer, privateLoadBalancer, kubernetesLoadBalanacer, dnsA, dnsA, containerSetup, dnsCname)
```
fiunchinho, line 1522: Here I build the whole template.

whites11: Well, of course it was better to have actual json files, expecially from a code completion/IDE point of view.
Not a blocker but maybe in the future we could check solutions like https://github.com/markbates/pkger ?

fiunchinho: If we get go modules working we can try that, I think it will be a better solution.

whites11:
What about this approach?

https://dev.to/koddr/the-easiest-way-to-embed-static-files-into-a-binary-file-in-your-golang-app-no-external-dependencies-43pc

integration/setup/common.go
```diff
@@ -70,6 +70,17 @@ func common(ctx context.Context, config Config) error {
 		}
 	}
 
+	{
+		config.Logger.LogCtx(ctx, "level", "debug", "message", "ensuring drainerconfig CRD exists")
```
fiunchinho, line 74: The drainerconfig CRD was being created by node-controller, but that got removed as part of the CRD graduation effort.

main.go
```diff
@@ -130,7 +119,6 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Azure.SubscriptionID, "", "ID of the Azure Subscription.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.TenantID, "", "ID of the Active Directory Tenant.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Azure.MSI.Enabled, true, "Whether to enabled Managed Service Identity (MSI).")
-	daemonCommand.PersistentFlags().String(f.Service.Azure.Template.URI.Version, defaultTemplateVersion, "URI version for ARM template links.")
```
fiunchinho, line 133: No need for this anymore, since it was used for ARM template links.

pkg/checksum/checksum.go
```diff
@@ -5,37 +5,20 @@ import (
...
 )
 
 func GetDeploymentTemplateChecksum(deployment resources.Deployment) (string, error) {
-	templateLink := deployment.Properties.TemplateLink
-	if templateLink == nil {
-		return "", microerror.Mask(nilTemplateLinkError)
-	}
-
-	resp, err := http.Get(*templateLink.URI)
+	template := deployment.Properties.Template.(map[string]interface{})
```
fiunchinho, line 14: Instead of downloading the template link, we calculate the checksum from the template property of the `Deployment`.

service/controller/resource/deployment/template/template.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+
+	"github.com/giantswarm/microerror"
+	"github.com/markbates/pkger"
+)
+
+// GetARMTemplate returns the ARM template reading a json file locally using pkger.
+func GetARMTemplate() (map[string]interface{}, error) {
```
fiunchinho, line 12: This is the function I created for reading files locally using `pkger`. It's copied and pasted in different resources. I didn't spent much time on code design since node pools is around the corner.

service/controller/resource/instance/deployment.go
```diff
@@ -98,14 +98,16 @@ func (r Resource) newDeployment(ctx context.Context, obj providerv1alpha1.AzureC
 		"zones":                 key.AvailabilityZones(obj, location),
 	}
 
+	armTemplate, err := template.GetARMTemplate()
```
fiunchinho, line 101: We read the template using the function and set it in the `Deployment`.

service/controller/resource/instance/template/main.json
```diff
@@ -158,10 +150,219 @@
       "type":"Microsoft.Resources/deployments",
       "name":"worker-vmss-deploy",
       "properties":{
+        "expressionEvaluationOptions":{
+          "scope": "inner"
+        },
```
fiunchinho, line 155: This is required when we want references (parameters or variables) in the nested template to reference variables/parameters defined in the template. Otherwise it will reference variables/parameters from parent template.

If this works fine, we can later see if we can remove some parameters from nested templates.

service/controller/resource/instance/template/template.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+
+	"github.com/giantswarm/microerror"
+	"github.com/markbates/pkger"
+)
+
+// GetARMTemplate returns the ARM template reading a json file locally using pkger.
+func GetARMTemplate() (map[string]interface{}, error) {
```
fiunchinho, line 12: Same as other package/resources.

service/controller/resource/deployment/template/template.go
```diff
@@ -0,0 +1,30 @@
...
+	defer f.Close()
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return contents, microerror.Mask(err)
+	}
+
+	if err := json.Unmarshal(b, &contents); err != nil {
+		return nil, err
+	}
```
tuommaki, line 28: Not super important, but in general it's nice to do stuff by streaming when original interface is streaming. Therefore this would be roughly as follows:
``` go
    d := json.NewDecoder(f)
    if err := d.Decode(&contents); err != nil {
    }
```

**NOTE**: Masking missing for unmarshalling error.

fiunchinho: Please, take a look now.

PR #789, opened 2020-04-22
Split master worker

whites11:
Towards giantswarm/giantswarm#10055

Moved internal functions from inside the instance resource to the resource package to be reused when I'll add the masters resource


whites11:
> Would it make sense to call the `helpers` package `vmss` to have a more explicit name?

it's the wrong branch once again ...  good god I need some rest

PR #791, opened 2020-04-22
Remove unused states

whites11:
Towards giantswarm/giantswarm#10055

This PR removes from the `instance` resource the states that are only related to the master instances.
Those states will be added back in the `masters` resource in a future PR.


service/controller/resource/instance/create_cordon_old_workers.go
```diff
@@ -94,6 +94,39 @@ func (r *Resource) cordonOldWorkersTransition(ctx context.Context, obj interface
 	return WaitForWorkersToBecomeReady, nil
 }
 
+func (r *Resource) allInstances(ctx context.Context, customObject providerv1alpha1.AzureConfig, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string) ([]compute.VirtualMachineScaleSetVM, error) {
```
whites11, line 97: This function is copied 1:1 from one of the masters state transition files.

service/controller/resource/instance/create_drain_old_worker_nodes.go
```diff
@@ -123,3 +124,45 @@ func (r *Resource) drainOldWorkerNodesTransition(ctx context.Context, obj interf
 
 	return TerminateOldWorkerInstances, nil
 }
+
+func (r *Resource) createDrainerConfig(ctx context.Context, customObject providerv1alpha1.AzureConfig, nodeName string) error {
```
whites11, line 128: This function is copied 1:1 from one of the masters transitions file

service/controller/resource/instance/create_test.go
```diff
@@ -1,552 +0,0 @@
-package instance
```
whites11, line 1: This test checks the functioning of instance selection.
This feature is not used by the workers anymore, so the test is unuseful.

PR #798, opened 2020-04-23
Added two more transition functions for the new `masters` resource

whites11:
Towards: giantswarm/giantswarm#10055

Added imlementation of a few more transition functions for the new masters resource.
Mostly copy-paste from the instance resource.

### DIFF
--- a/service/controller/resource/masters/create_wait_for_masters_to_become_ready.go
+++ b/service/controller/resource/masters/create_wait_for_masters_to_become_ready.go
@@ -1,0 +1,96 @@
+package masters
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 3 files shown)


service/controller/resource/masters/create_wait_for_masters_to_become_ready.go
```diff
@@ -0,0 +1,96 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "found out that all tenant cluster master nodes are not Ready")
+		return currentState, nil
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "found out that all tenant cluster master nodes are Ready")
+
+	return DeploymentCompleted, nil
+}
+
+func areNodesReadyForTransitioning(ctx context.Context, nodeRoleMatchFunc func(corev1.Node) bool) (bool, error) {
```
nprokopic, line 36: Maybe just use `isMaster` directly if this is the only invoked with `isMaster` function (since masters and workers are being split).
KISS principle :)

PR #799, opened 2020-04-23
Added final transition function implementation

whites11:
Towards: giantswarm/giantswarm#10055

This the last transition function implementation and the last PR for the whole feature.
Thanks for bearing with me team.

### DIFF
--- a/service/controller/resource/masters/create_master_instances_upgrading.go
+++ b/service/controller/resource/masters/create_master_instances_upgrading.go
@@ -1,0 +1,449 @@
+package masters
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/azure-operator/service/controller/internal/state"
+	"github.com/giantswarm/azure-operator/service/controller/key"
+)
+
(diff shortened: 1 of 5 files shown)


service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -0,0 +1,449 @@
...
+	if !masterUpgradeInProgress {
+		// When masters are upgraded, proceed to workers.
+		return WaitForMastersToBecomeReady, nil
+	}
+
+	// Upgrade still in progress. Keep current state.
+	return currentState, nil
+}
+
+func (r *Resource) allInstances(ctx context.Context, customObject providerv1alpha1.AzureConfig, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string) ([]compute.VirtualMachineScaleSetVM, error) {
```
nprokopic, line 97: We should have these functions that are common for masters and workers in some common place. Not a blocker, it can be done in follow up refactoring PR.

service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -0,0 +1,449 @@
...
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found the scale set '%s'", deploymentNameFunc(customObject)))
+
+	return instances, nil
+}
+
+func (r *Resource) createDrainerConfig(ctx context.Context, customObject providerv1alpha1.AzureConfig, nodeName string) error {
```
nprokopic, line 130: Same as for `r.allInstances(...)`.

service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -0,0 +1,449 @@
...
+	"github.com/giantswarm/azure-operator/service/controller/key"
+)
+
+func (r *Resource) masterInstancesUpgradingTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	versionValue := map[string]string{}
```
fiunchinho, line 26: is it ok if this is `nil` because `Status.Cluster.Nodes` is empty?

whites11: If the map is empty, we simply not reimage any master node.
We might end up in an inconsistent state, but I am not sure how to deal with that, because if the information is missing from the CR I can't automatically reconstruct it from elsewhere.
There shouldn't be panics, though

PR #801, opened 2020-04-23
DO NOT MERGE Ported flatcar modifications to the splitter ARM template branch

whites11:
DO NOT MERGE

This PR is the adaptation of Marcel's after the split of ARM templates.
This is tested and working for new cluster creation only.
We should use this as a parachute in case we can't get to a working upgrade for the master nodes.


service/controller/resource/instance/template/main.json
```diff
@@ -156,7 +156,7 @@
     {
       "apiVersion":"[variables('apiVersion')]",
       "type":"Microsoft.Resources/deployments",
-      "name":"worker-vmss-deploy",
+      "name":"workers-vmss-template",
```
tuommaki, line 159: I think we need to add the IDs here as well, for the coming node pools?

pkg/helpers/vmss/types.go
```diff
@@ -35,24 +33,24 @@ type NodeOSImage struct {
...
+		OSImage:             newNodeOSImageCoreOS(distroVersion),
 		VMSize:              vmSize,
 		DockerVolumeSizeGB:  dockerVolumeSizeGB,
 		KubeletVolumeSizeGB: kubeletVolumeSizeGB,
 	}
 }
 
 // newNodeOSImage provides OS information for Container Linux
-func newNodeOSImageCoreOS() NodeOSImage {
+func newNodeOSImageCoreOS(distroVersion string) NodeOSImage {
```
tuommaki, line 48:
```suggestion
func newNodeOSImageFlatcar(distroVersion string) NodeOSImage {
```
:question:

PR #802, opened 2020-04-24
Override tested operator

fiunchinho:
With these changes we are able to override the chart package used to install the operator, which allows us to pass a local path so we don't have to push and wait for CI to run tests locally.

I also added documentation on how to run e2e tests locally from dev machine.

### DIFF
--- a/docs/running_e2etests_locally.md
+++ b/docs/running_e2etests_locally.md
@@ -1,4 +1,4 @@
-# Running e2e tests locally
+# Running integration tests locally
 
 Build the application, which means compiling the Go code, building the docker image and the helm chart package.
 
--- a/docs/running_e2etests_locally.md
+++ b/docs/running_e2etests_locally.md
@@ -1,0 +1,25 @@
+# Running e2e tests locally
+
+Build the application, which means compiling the Go code, building the docker image and the helm chart package.
+
+```bash
+CGO_ENABLED=0 go build -a -v -ldflags "-w -linkmode 'auto' -extldflags '-static' -X 'main.gitCommit=`git rev-parse HEAD`'" \
+  && docker build -t "quay.io/giantswarm/azure-operator:`architect project version`" . \
+  && architect helm template --dir ./helm/azure-operator \
+  && helm package helm/azure-operator \
+  && git checkout HEAD -- ./helm/azure-operator/Chart.yaml ./helm/azure-operator/values.yaml \
+```
+
+Prepare the environment using kind
+
+```bash
+(kind delete cluster || true) && kind create cluster && kind load docker-image quay.io/giantswarm/azure-operator:`architect project version` \
+  && kind get kubeconfig > kubeconfig.yaml && sudo ln -fs "${PWD}/kubeconfig.yaml" /workdir/.shipyard/config \
+  && helm init --wait && k create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:default
+```
+
+
+You can then run the tests passing the helm chart package that we just generated
+```bash
+OPERATOR_HELM_TARBALL_PATH="${PWD}/azure-operator-`architect project version`.tgz" AZURE_SUBSCRIPTIONID="${AZURE_SUBSCRIPTIONID}" AZURE_TENANTID="${AZURE_TENANTID}" AZURE_CLIENTID="${AZURE_CLIENTID}" AZURE_CLIENTSECRET="${AZURE_CLIENTSECRET}" CIRCLE_SHA1="`git rev-parse HEAD`" REGISTRY_PULL_SECRET="${REGISTRY_PULL_SECRET}" TEST_DIR="integration/test/multiaz" AZURE_AZS=1 go test -v -tags=k8srequired ./integration/test/multiaz
+```
--- a/integration/env/common.go
+++ b/integration/env/common.go
@@ -14,16 +14,18 @@ import (
 const (
 	DefaultTestedVersion = "wip"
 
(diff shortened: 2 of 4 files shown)


integration/env/common.go
```diff
@@ -14,16 +14,18 @@ import (
...
-	EnvVarCircleSHA            = "CIRCLE_SHA1" // #nosec
-	EnvVarKeepResources        = "KEEP_RESOURCES"
-	EnvVarRegistryPullSecret   = "REGISTRY_PULL_SECRET" // #nosec
-	EnvVarTestedVersion        = "TESTED_VERSION"
-	EnvVarTestDir              = "TEST_DIR"
-	EnvVarVersionBundleVersion = "VERSION_BUNDLE_VERSION"
+	EnvVarCircleSHA               = "CIRCLE_SHA1" // #nosec
+	EnvVarKeepResources           = "KEEP_RESOURCES"
+	EnvVarRegistryPullSecret      = "REGISTRY_PULL_SECRET" // #nosec
+	EnvVarOperatorHelmTarballPath = "OPERATOR_HELM_TARBALL_PATH"
```
fiunchinho, line 20: Added new env var to be able to override the chart tarball used

integration/setup/setup.go
```diff
@@ -148,8 +148,8 @@ func GetLatestOperatorRelease() string {
 func installResources(ctx context.Context, config Config) error {
 	var err error
 
-	var operatorTarballPath string
-	{
+	operatorTarballPath := env.OperatorHelmTarballPath()
+	if operatorTarballPath == "" {
```
fiunchinho, line 152: Only download chart tarball if env var is empty.

integration/setup/setup.go
```diff
@@ -274,6 +274,18 @@ func installResources(ctx context.Context, config Config) error {
 		}
 	}
 
+	var sshUserList []providerv1alpha1.ClusterKubernetesSSHUser
+	{
+		if env.SSHPublicKey() != "" {
```
fiunchinho, line 279: There was a problem when not passing the SSH public key. Now we only set the SSH public key if something is passed.

docs/running_e2etests_locally.md
```diff
@@ -0,0 +1,25 @@
+# Running e2e tests locally
+
+Build the application, which means compiling the Go code, building the docker image and the helm chart package.
+
+```bash
+CGO_ENABLED=0 go build -a -v -ldflags "-w -linkmode 'auto' -extldflags '-static' -X 'main.gitCommit=`git rev-parse HEAD`'" \
```
fiunchinho, line 6: It would be cool to have an `architect` command to run these commands, something like `architect build --docker --chart`

fiunchinho: what do you think @giantswarm/wg-integration-testing ?

docs/running_e2etests_locally.md
```diff
@@ -0,0 +1,25 @@
+# Running e2e tests locally
```
rossf7, line 1:
```suggestion
# Running integration tests locally
```

Not a blocker but I think this is clearer.

docs/running_e2etests_locally.md
rossf7: Using a binary makes sense to me.

I just get confused what should be in architect vs devctl. But this is using architect helm template so maybe its the right place.

But I'd like to hear others opinion.

docs/running_e2etests_locally.md
```diff
@@ -0,0 +1,25 @@
...
+  && docker build -t "quay.io/giantswarm/azure-operator:`architect project version`" . \
+  && architect helm template --dir ./helm/azure-operator \
+  && helm package helm/azure-operator \
+  && git checkout HEAD -- ./helm/azure-operator/Chart.yaml ./helm/azure-operator/values.yaml \
+```
+
+Prepare the environment using kind
+
+```bash
+(kind delete cluster || true) && kind create cluster && kind load docker-image quay.io/giantswarm/azure-operator:`architect project version` \
```
rossf7, line 16: TIL `kind load` 

<3

docs/running_e2etests_locally.md
```diff
@@ -0,0 +1,25 @@
...
+  && kind get kubeconfig > kubeconfig.yaml && sudo ln -fs "${PWD}/kubeconfig.yaml" /workdir/.shipyard/config \
+  && helm init --wait && k create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:default
+```
+
+
+You can then run the tests passing the helm chart package that we just generated
+```bash
+OPERATOR_HELM_TARBALL_PATH="${PWD}/azure-operator-`architect project version`.tgz" AZURE_SUBSCRIPTIONID="${AZURE_SUBSCRIPTIONID}" AZURE_TENANTID="${AZURE_TENANTID}" AZURE_CLIENTID="${AZURE_CLIENTID}" AZURE_CLIENTSECRET="${AZURE_CLIENTSECRET}" CIRCLE_SHA1="`git rev-parse HEAD`" REGISTRY_PULL_SECRET="${REGISTRY_PULL_SECRET}" TEST_DIR="integration/test/multiaz" AZURE_AZS=1 go test -v -tags=k8srequired ./integration/test/multiaz
+```
```
rossf7, line 25: This is where we talked about having a Keybase git repo with a script. I think that would help but can come later.

docs/running_e2etests_locally.md
fiunchinho: I figured `architect` would be best since it contained `architect build` and `architect helm template`. But open to suggestions.

rossf7: OK cool that makes sense to me.

PR #803, opened 2020-04-27
Workers flatcar

whites11:
This PR is about bringing migration from coreOS to flatcar possible.

the amount of modifications is quite high, and the process is not fully automated. The ETCD data migration is in fact manual, aided by two scripts (attached to this PR).

The state machine flow chart for the master instance is available here: https://drive.google.com/file/d/1alsOOChl39-LsRDPbWlv5LwnHt4b7A-O/view?usp=sharing

Open for questions/clarifications

[scripts.zip](https://github.com/giantswarm/azure-operator/files/4538334/scripts.zip)


scripts/backup_etcd_disk_tenant_cluster.sh
```diff
@@ -0,0 +1,51 @@
...
+    exit 1
+fi
+
+read -p "Installation: $installation, cluster: $cluster. Hit enter to continue, or ctrl+c to abort. "
+
+echo -n "Creating kubeconfig for $installation: "
+opsctl create kubeconfig -i $installation >/dev/null
+echo "OK"
+echo -n "Creating kubeconfig for $cluster @ $installation: "
+gsctl create kubeconfig --cluster=$cluster --certificate-organizations system:masters -e $installation >/dev/null
```
MarcelMue, line 20:
```suggestion
gsctl create kubeconfig --cluster=$cluster --certificate-organizations system:masters -e $installation --ttl 1d >/dev/null
```

scripts/backup_etcd_disk_tenant_cluster.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "OK"
+
+echo -n "Getting master node name: "
+master="`kubectl --context=giantswarm-$cluster get no -l role=master|grep -v NAME|awk '{print $1}'`"
+
+if [ "$master" == "" ]
+then
+    echo "ERR"
```
MarcelMue, line 28:
```suggestion
    echo "ERR: Master not found"
```

scripts/restore_etcd_disk_tenant_cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+    exit 1
+fi
+
+read -p "Installation: $installation, cluster: $cluster. Hit enter to continue, or ctrl+c to abort. "
+
+echo -n "Creating kubeconfig for $installation: "
+opsctl create kubeconfig -i $installation >/dev/null
+echo "OK"
+echo -n "Creating kubeconfig for $cluster @ $installation: "
+gsctl create kubeconfig --cluster=$cluster --certificate-organizations system:masters -e $installation >/dev/null
```
MarcelMue, line 20:
```suggestion
gsctl create kubeconfig --cluster=$cluster --certificate-organizations system:masters -e $installation --ttl 1d >/dev/null
```

scripts/restore_etcd_disk_tenant_cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+echo "OK"
+
+echo -n "Getting master node name: "
+master="`kubectl --context=giantswarm-$cluster get no -l role=master|grep -v NAME|awk '{print $1}'`"
+
+# TODO check there is only one line in the above result
+
+if [ "$master" == "" ]
+then
+    echo "ERR"
```
MarcelMue, line 30:
```suggestion
    echo "ERR: Master not found"
```

scripts/restore_etcd_disk_tenant_cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+fi
+
+echo "OK"
+
+echo -n "Stopping API server: "
+opsctl ssh $installation $master --cmd "sudo mv /etc/kubernetes/manifests/k8s-api-server.yaml /root || true" 2>&1 >/dev/null
+echo "OK"
+
+echo -n "Waiting 10 seconds to ensure the api server goes down: "
+sleep 10
```
MarcelMue, line 41: TODO: Actually check if it's down.

scripts/backup_etcd_disk_tenant_cluster.sh
whites11: done

scripts/backup_etcd_disk_tenant_cluster.sh
whites11: this part of the script changed a lot already, it is more verbose now

scripts/restore_etcd_disk_tenant_cluster.sh
whites11: this part of the script changed a lot already, it is more verbose now

scripts/restore_etcd_disk_tenant_cluster.sh
whites11: done

service/controller/resource/instance/cordon_old_vmss_transition.go
```diff
@@ -0,0 +1,96 @@
...
+	if IsScaleSetNotFound(err) {
+		return "", microerror.Mask(err)
+	} else if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("The legacy VMSS %s is still present", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
+
+	// The legacy VMSS was found, check the scaling.
+	if *vmss.Sku.Capacity > 0 {
```
fiunchinho, line 44:
```suggestion
        legacyVmssHasInstancesRunning := *vmss.Sku.Capacity > 0
	if legacyVmssHasInstancesRunning {
```

service/controller/resource/instance/cordon_old_vmss_transition.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if cc.Client.TenantCluster.K8s == nil {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster client not available yet") // nolint: errcheck
+		return currentState, nil
+	}
+
+	// If the legacy VMSS still exists with at least one replica, we want to cordon its replicas.
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Checking if the legacy VMSS %s is still present", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
+	vmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
```
fiunchinho, line 34: What do you think about?

```suggestion
	legacyVmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
```

service/controller/resource/blobobject/desired.go
```diff
@@ -63,7 +49,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	var ignitionTemplateData cloudconfig.IgnitionTemplateData
 	{
-		versions, err := v_6_0_0.ExtractComponentVersions(release.Spec.Components)
+		versions, err := v_6_0_0.ExtractComponentVersions(cc.Release.Components)
```
fiunchinho, line 52: why did we move it to the controller context?

service/controller/resource/instance/create_drain_old_vmss.go
```diff
@@ -0,0 +1,125 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding all worker VMSS instances")                         // nolint: errcheck
+
+	allWorkerInstances, err := r.allInstances(ctx, cr, key.LegacyWorkerVMSSName)
+	if IsScaleSetNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the scale set '%s'", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
+	} else if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found %d worker VMSS instances", len(allWorkerInstances))) // nolint: errcheck
```
fiunchinho, line 51: no need for `nolint: errcheck` anymore

service/controller/resource/instance/create_scale_workers.go
```diff
@@ -28,6 +29,28 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
 		return "", microerror.Mask(err)
 	}
 
+	// If the old VMSS is still present, we should skip this step.
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Checking if the legacy VMSS %s is still present", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
+	vmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
```
fiunchinho, line 34: same as above, what do you think?
```suggestion
	legacyVmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
```

service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
```diff
@@ -29,6 +31,31 @@ func (r *Resource) waitForWorkersToBecomeReadyTransition(ctx context.Context, ob
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", "found out that all tenant cluster worker nodes are Ready")
 
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return "", microerror.Mask(err)
```
fiunchinho, line 36: I believe it would help understand the flow if we return `DeploymentUnintialized` instead of "".

service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
```diff
@@ -29,6 +31,31 @@ func (r *Resource) waitForWorkersToBecomeReadyTransition(ctx context.Context, ob
...
+	}
+
+	// If the old VMSS still exists, we want to go to a different state.
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Checking if the legacy VMSS %s is still present", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
+	vmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
+	if IsScaleSetNotFound(err) {
+		return "", microerror.Mask(err)
```
fiunchinho, line 43: just making sure (you know best): this will restart the state machine , is that really what we want?

service/controller/resource/masters/create.go
```diff
@@ -63,10 +72,5 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "no state change")
 	}
 
-	// We should wait and avoid further resources to reconciliate as long as the masters are not successfully deployed.
```
fiunchinho, line 66: Not sure I understand how this changes the semantics. It will continue to next resource (instance) if there hasn't been a state change?

service/controller/resource/instance/cordon_old_vmss_transition.go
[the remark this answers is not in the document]
nprokopic: Also `r.getScaleSet(...)` never returns `scaleSetNotFound` error.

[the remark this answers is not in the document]
nprokopic: It should return something like this https://github.com/giantswarm/azure-operator/blob/a1f7bc1e8e3977066c7a3268874919afea955059/service/controller/resource/instance/create_cordon_old_workers.go#L109

service/controller/resource/masters/create.go
tuommaki: This is from my change. With this code the reconciliation didn't go further until master was upgraded. If there's a problem with VPN connection during that, it prevents master upgrade but operator cannot reconcile the broken connection. Therefore I changed to logic so that `instance` resource checks for `masters` resource's status condition and skips execution until it's `DeploymentCompleted`. For intended purpose it works the same but allows reconciliation of e.g. vpn resources (which don't change with flatcar but is crucial for whole operations).

service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
tuommaki: Huh? No, it's returning error so that won't change the state.

service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
tuommaki: No, this returns error. See https://github.com/giantswarm/fmt/blob/master/go/errors.md#zero-values-when-returning-errors

service/controller/resource/masters/create_delete_legacy_vmss.go
```diff
@@ -0,0 +1,42 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return Empty, microerror.Mask(err)
+	}
+
+	// Delete the scaleset
+	err = r.deleteScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyMasterVMSSName(cr))
+	if IsScaleSetNotFound(err) {
+		// Scale set not found, all good.
+		return DeploymentCompleted, nil
```
tuommaki, line 22: Is there any chance that scale set deletion went through but somehow still returned an error and on next reconciliation round we skip `UnblockAPICalls` even though it would need to be done?

service/controller/resource/masters/create_manual_intervention_needed.go
```diff
@@ -0,0 +1,13 @@
+package masters
+
+import (
+	"context"
+
+	"github.com/giantswarm/azure-operator/service/controller/internal/state"
+)
+
+// This transition function aims at detecting if the master VMSS needs to be migrated from CoreOS to flatcar.
```
tuommaki, line 9: I believe this comment is copy&paste? Zap? :slightly_smiling_face: 
```suggestion
```

service/controller/resource/masters/create_restart_kubelets.go
```diff
@@ -0,0 +1,99 @@
...
+	up, err := r.isApiServerUP(ctx)
+	if err != nil || !up {
+		msg := "API server is NOT up."
+		if err != nil {
+			msg = fmt.Sprintf("%s Original error was %s", msg, err.Error())
+		}
+		r.logger.LogCtx(ctx, "level", "debug", "message", msg)
+
+		return currentState, nil
+	}
```
tuommaki, line 28: I think combining checks here is unnecessary. It would be easier to read and probably shorter to write if they are checked separately :slightly_smiling_face:

service/controller/resource/masters/create_restart_kubelets.go
```diff
@@ -0,0 +1,99 @@
...
+		"sudo systemctl restart k8s-kubelet",
+	}
+	runCommandInput := compute.RunCommandInput{
+		CommandID: &commandId,
+		Script:    &script,
+	}
+
+	allMasterInstances, err := r.allInstances(ctx, cr, key.LegacyWorkerVMSSName)
+	if err != nil {
+		return currentState, microerror.Mask(err)
```
tuommaki, line 62:
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_restart_kubelets.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	for _, instance := range allMasterInstances {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Sending restart kubelet command to %s", *instance.Name))
+		runCommandFuture, err := vmssVMsClient.RunCommand(ctx, *group.Name, key.LegacyWorkerVMSSName(cr), *instance.InstanceID, runCommandInput)
+		if err != nil {
+			return currentState, microerror.Mask(err)
```
tuommaki, line 69:
```suggestion
			return "", microerror.Mask(err)
```

service/controller/resource/masters/create_restart_kubelets.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		_, err = vmssVMsClient.RunCommandResponder(runCommandFuture.Response())
+		if err != nil {
+			return currentState, microerror.Mask(err)
```
tuommaki, line 74:
```suggestion
			return "", microerror.Mask(err)
```

service/controller/resource/masters/create_wait_for_backup_confirmation.go
```diff
@@ -0,0 +1,14 @@
+package masters
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/azure-operator/service/controller/internal/state"
+)
+
+// This transition function aims at detecting if the master VMSS needs to be migrated from CoreOS to flatcar.
```
tuommaki, line 10: Foobar
```suggestion
```

service/controller/resource/masters/error.go
```diff
@@ -72,6 +72,25 @@ func IsMissingLabel(err error) bool {
...
 }
 
+func IsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	c := microerror.Cause(err)
+
+	{
```
tuommaki, line 82: Not sure about the block here, but it looks somehow weird. Alternative proposal:
```suggestion
	{
		c := microerror.Cause(err)
```

service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
fiunchinho: Sorry, I don't know what I was thinking.

service/controller/resource/instance/create_scale_workers.go
```diff
@@ -82,6 +105,20 @@ func (r *Resource) getInstancesCount(ctx context.Context, customObject providerv
...
+func (r *Resource) getScaleSet(ctx context.Context, resourceGroup string, scaleSetName string) (*compute.VirtualMachineScaleSet, error) {
+	c, err := r.getScaleSetsClient(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	vmss, err := c.Get(ctx, resourceGroup, scaleSetName)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
nprokopic, line 118:
```suggestion
	if IsScaleSetNotFound(err) {
		return nil, microerror.Mask(scaleSetNotFoundError)
	} else if err != nil {
		return nil, microerror.Mask(err)
	}
```

scripts/restore_etcd_disk_tenant_cluster.sh
whites11: done

service/controller/internal/vmsscheck/deleter_job.go
```diff
@@ -0,0 +1,113 @@
...
+		instance := iterator.Value()
+
+		gj.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Instance %s has state %s", *instance.Name, *instance.ProvisioningState))
+
+		switch *instance.ProvisioningState {
+		case provisioningStateFailed:
+			// Reimage the instance.
+			gj.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Deleting instance %s", *instance.Name))
+
+			retries := 3
```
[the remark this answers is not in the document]
whites11: removed the retry

service/controller/internal/vmsscheck/deleter_job.go
```diff
@@ -0,0 +1,113 @@
...
+					retries = retries - 1
+					time.Sleep(5 * time.Second)
+					continue
+				}
+
+				break
+			}
+
+			gj.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Deleted instance %s", *instance.Name))
+			allSucceeded = false
```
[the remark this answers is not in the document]
whites11: yeah, in a single iteration if it gets to false it will never get back to true. Will ask for deletion and wait for next iteration. Resolving

service/controller/resource/instance/cordon_old_vmss_transition.go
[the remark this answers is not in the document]
whites11: removed

service/controller/resource/masters/create_deallocate_legacy_instance.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	instancesRunning, err := r.getRunningInstances(ctx, resourceGroup, vmssName)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if len(instancesRunning) > 0 {
```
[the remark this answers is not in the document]
whites11: It's just for the log message. It's not harming so I'll leave it

service/controller/resource/blobobject/desired.go
whites11: @MarcelMue did this, would you like to answer?

service/controller/resource/instance/create_drain_old_vmss.go
whites11: OK this PR is already huge, I will not remove all those.

service/controller/resource/instance/create_terminate_old_vmss.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	_, err = dc.Delete(ctx, key.ResourceGroupName(cr), legacyVMSSDeploymentName)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted the legacy VMSS deployment %s", legacyVMSSDeploymentName)) // nolint: errcheck
+
+	return DeploymentCompleted, nil
```
nprokopic, line 50: Is it ok here that we don't wait for the VMSS and the deployment to actually be deleted?

When you call `c.Delete(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))` and `dc.Delete(ctx, key.ResourceGroupName(cr), legacyVMSSDeploymentName)` above to delete legacy VMSS and its deployment, they are not still deleted, the functions return Futures.

So if you're not waiting for the Future to complete, then maybe:
- check if the VMSS exist, and if it exist and it's running (or just not being deleted), delete it; check if the deployment is there, and if it exist and not being deleted, delete it; return same state
- if both VMSS and deployment are deleted, set the state to `DeploymentCompleted`

Or just wait for the Futures to complete with something like this:
```golang
vmssDeleteFuture, err := c.Delete(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
if err != nil {
}

_, err = c.DeleteResponder(vmssDeleteFuture.Response())
if err != nil {
}
```

service/controller/resource/instance/create_scale_workers.go
whites11: done

service/controller/resource/masters/create_block_api_calls.go
```diff
@@ -0,0 +1,174 @@
...
+			Protocol:                 network.SecurityRuleProtocolAsterisk,
+			Priority:                 to.Int32Ptr(3000),
+			SourceAddressPrefix:      to.StringPtr("*"),
+			SourcePortRange:          to.StringPtr("*"),
+		},
+	}
+
+	workerFound, err := r.ensureSecurityRule(ctx, key.ResourceGroupName(cr), key.WorkerSecurityGroupName(cr), temporarySecurityRuleName, workersRule)
+	if err != nil {
+		return currentState, microerror.Mask(err)
```
nprokopic, line 55: nitpick :) according to another comment from @tuommaki, we should return `""` here (and in few more other places). See https://github.com/giantswarm/fmt/blob/master/go/errors.md#zero-values-when-returning-errors.
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_block_api_calls.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	masterFound, err := r.ensureSecurityRule(ctx, key.ResourceGroupName(cr), key.MasterSecurityGroupName(cr), temporarySecurityRuleName, mastersRule)
+	if err != nil {
+		return currentState, microerror.Mask(err)
```
nprokopic, line 60:
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_block_api_calls.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Security rule in place
+	return true, nil
+}
+
+func (r *Resource) unblockAPICallsTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return currentState, microerror.Mask(err)
```
nprokopic, line 102:
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_block_api_calls.go
```diff
@@ -0,0 +1,174 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting security rule %s from security group %s", temporarySecurityRuleName, key.WorkerSecurityGroupName(cr)))
+
+	// Delete security rule for workers
+	err = r.deleteSecurityRule(ctx, key.ResourceGroupName(cr), key.WorkerSecurityGroupName(cr), temporarySecurityRuleName)
+	if IsNotFound(err) {
+		// Rule not exists, ok to continue.
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("security rule %s from security group %s was not found", temporarySecurityRuleName, key.WorkerSecurityGroupName(cr)))
+	} else if err != nil {
+		// In case of error just retry.
+		return currentState, microerror.Mask(err)
```
nprokopic, line 114:
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_block_api_calls.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Delete security rule for masters
+	err = r.deleteSecurityRule(ctx, key.ResourceGroupName(cr), key.MasterSecurityGroupName(cr), temporarySecurityRuleName)
+	if IsNotFound(err) {
+		// Rule not exists, ok to continue.
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("security rule %s from security group %s was not found", temporarySecurityRuleName, key.MasterSecurityGroupName(cr)))
+	} else if err != nil {
+		// In case of error just retry.
+		return currentState, microerror.Mask(err)
```
nprokopic, line 124:
```suggestion
		return "", microerror.Mask(err)
```

service/controller/resource/masters/create_check_flatcar_migration_needed.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/giantswarm/azure-operator/service/controller/internal/state"
+	"github.com/giantswarm/azure-operator/service/controller/key"
+)
+
+// This transition function aims at detecting if the master VMSS needs to be migrated from CoreOS to flatcar.
+func (r *Resource) checkFlatcarMigrationNeededTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return Empty, microerror.Mask(err)
```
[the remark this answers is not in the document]
nprokopic: It doesn't matter, right? As we do not save the state in case of an error.

service/controller/resource/blobobject/desired.go
MarcelMue: It was nicer for me to have a dedicated resource to get the Releases once and then take them from the context when I needed them.

Just personal preference I guess.

service/controller/internal/vmsscheck/deleter_job.go
```diff
@@ -0,0 +1,113 @@
...
+// If any of the instances is not Succeeded, returns false.
+// It deletes instances that are in "Failed" state.
+func (gj *deleterJob) deleteFailedInstances(ctx context.Context, rg string, vmssName string) (bool, error) {
+	c, err := getVMsClient(ctx)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	// Get a list of instances in the VMSS.
+	iterator, err := c.ListComplete(ctx, rg, vmssName, "", "", "")
```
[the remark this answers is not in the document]
whites11: yeah but we can't avoid it

service/controller/resource/masters/create_cluster_upgrade_requirement_check.go
```diff
@@ -23,7 +23,15 @@ func (r *Resource) clusterUpgradeRequirementCheckTransition(ctx context.Context,
...
 	isCreating := r.isClusterCreating(cr)
 	anyOldNodes, err := r.anyNodesOutOfDate(ctx)
-	if err != nil {
+	if IsClientNotFound(err) {
+		// The kubernetes API is down.
+		// We check if the Legacy Master VMSS exists and in that case
+		// we assume this is because we're migrating to Flatcar.
+		exists, err := r.vmssExists(ctx, key.ResourceGroupName(cr), key.LegacyMasterVMSSName(cr))
+		if err != nil || !exists {
+			return "", microerror.Mask(err)
```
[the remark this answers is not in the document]
nprokopic: > this return "" will take state machine back to beginning and it gets stuck

Will it take it back? We are not saving the state if the error is returned, right?
https://github.com/giantswarm/azure-operator/blob/e57695f86b29d2488d7e1bdfce8c490c0458f025/service/controller/resource/masters/create.go#L48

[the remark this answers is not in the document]
whites11: Not sure what to do when `! exists` is true. That means that the old VMSS is not existing but the new one is either not existing or not working/responding.
We could wait for it to come up (stay in the same state) but that could lead to an infinite loop (if the new VMSS does not exist at all it will never come up).
How do you suggest to handle this error?

service/controller/resource/masters/create_delete_legacy_vmss.go
whites11: Maybe there is, but I never witnessed it.
I am not sure how to handle this differently.

service/controller/resource/masters/create_manual_intervention_needed.go
whites11: removed

service/controller/resource/masters/create_restart_kubelets.go
whites11: done

service/controller/resource/masters/create_restart_kubelets.go
whites11: done

service/controller/resource/masters/create_restart_kubelets.go
whites11: done

service/controller/resource/masters/create_wait_for_backup_confirmation.go
whites11: lol

service/controller/resource/masters/error.go
whites11: agreed

service/controller/resource/instance/cordon_old_vmss_transition.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/azure-operator/service/controller/controllercontext"
+	"github.com/giantswarm/azure-operator/service/controller/internal/state"
+	"github.com/giantswarm/azure-operator/service/controller/key"
+)
+
+func (r *Resource) cordonOldVMSSTransition(ctx context.Context, obj interface{}, currentState state.State) (state.State, error) {
```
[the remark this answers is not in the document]
whites11: done

service/controller/resource/instance/create_terminate_old_vmss.go
whites11: We can't wait because the reconciliation loop would be too long.
You are right, we should check if the VMSS and deployments gets delete correctly.
What will happen if they don't? The VMSS simply stays there and we can get rid of it manually.
I know this doesn't sound like very "SRE", but adding new checks and complicating the code even more is not worth it IMHO.

service/controller/resource/masters/create_cluster_upgrade_requirement_check.go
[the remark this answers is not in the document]
nprokopic: > We could wait for it to come up (stay in the same state) but that could lead to an infinite loop

By returning the error, I believe that we are staying in the same state, because on error we do not save the returned state (`""` in this case).

service/controller/resource/masters/create_block_api_calls.go
whites11: done

whites11:
> For me I think it's enough if tests have yield successful results. Good job man, seriously.

I will do some more tests now, after most of the comments have been addressed.

service/controller/resource/masters/create_cluster_upgrade_requirement_check.go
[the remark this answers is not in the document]
tuommaki: > By returning the error, I believe that we are staying in the same state, because on error we do not save the returned state

Yes. On error there's no transitioning.

service/controller/resource/instance/create_terminate_old_vmss.go
nprokopic: (late post-merge and post-vacation comment :smile:  )

> but adding new checks and complicating the code even more is not worth it IMHO

Agreed for this, as it was a one time thing for upgrading to Flatcar.

PR #804, opened 2020-04-27
Added another VM size

whites11:


nprokopic:
Should we closed this PR, as this is now done here https://github.com/giantswarm/azure-operator/pull/805?

PR #805, opened 2020-04-27
Add Standard_E8a_v4 and Standard_E8as_v4 VM sizes

nprokopic:
Added `Standard_E8a_v4` and `Standard_E8as_v4` VM sizes for masters and workers.

### DIFF
--- a/service/controller/resource/instance/template/main.json
+++ b/service/controller/resource/instance/template/main.json
@@ -127,6 +127,8 @@
       "Standard_E4_v3",
       "Standard_E64_v3",
       "Standard_E8_v3",
+      "Standard_E8a_v4",
+      "Standard_E8as_v4",
       "Standard_F1",
       "Standard_F16",
       "Standard_F2",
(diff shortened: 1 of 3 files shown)


whites11:
This PR is not enough.
You need to add the `Standard_E8a_v4` instance type to the list of types that don't support premium storage

service/controller/resource/instance/template/main.json
```diff
@@ -127,6 +127,8 @@
       "Standard_E4_v3",
       "Standard_E64_v3",
       "Standard_E8_v3",
+      "Standard_E8a_v4",
```
nprokopic, line 130: @whites11 you meant this?

service/controller/resource/masters/template/main.json
```diff
@@ -127,6 +127,8 @@
       "Standard_E4_v3",
       "Standard_E64_v3",
       "Standard_E8_v3",
+      "Standard_E8a_v4",
```
nprokopic, line 130: @whites11 and this?

PR #808, opened 2020-04-29
Added CHANGELOG

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/10250


fiunchinho:
I may be missing some bits but it's better than nothing.

PR #809, opened 2020-04-30
Remove ensuring Azure CRD exists

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9963

From now on, [CRD's are ensured using opsctl](https://github.com/giantswarm/opsctl/pull/633) instead of ensuring them from the operators.
This is part of the work required for the [graduation of Azure CRD's to v1](https://github.com/giantswarm/apiextensions/pull/439).

Already released versions of the operator have been replaced with this patch as well.


fiunchinho:
I will wait until go modules is merged, since I need to upgade to apiextensions `v0.3.4`. [The branch using go modules](https://github.com/giantswarm/azure-operator/pull/785/files#diff-37aff102a57d3d7b797f152915a6dc16R15) is using apiextensions `v0.3.2`, so it should be easy to update.

PR #815, opened 2020-05-05
Bump to new wip version

whites11:


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "4.0.0"
+	version            = "4.0.0-dev"
```
tuommaki, line 8:
```suggestion
	version            = "4.1.0-dev"
```

tuommaki: ...we already had `4.0.0-dev` and that's now released, so either `4.0.1` or `4.1.0`. Given that we bump up go modules & stuff, I'd maybe opt for `4.1.0` ?

PR #819, opened 2020-05-06
Nat gateway for workers

whites11:
Towards #10657

This PR is about making the worker instances get access to the internet using a NAT gateway instead of the ingress controller load balancer.


service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go
```diff
@@ -40,7 +40,8 @@ func (r *Resource) waitForWorkersToBecomeReadyTransition(ctx context.Context, ob
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Checking if the legacy VMSS %s is still present", key.LegacyWorkerVMSSName(cr))) // nolint: errcheck
 	vmss, err := r.getScaleSet(ctx, key.ResourceGroupName(cr), key.LegacyWorkerVMSSName(cr))
 	if IsScaleSetNotFound(err) {
-		return "", microerror.Mask(err)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("The legacy VMSS %s was not found", key.LegacyWorkerVMSSName(cr)))
+		return DrainOldWorkerNodes, nil
```
whites11, line 44: This is a fix unrelated to this task. It prevents updating from 11.3.0 to anything newer.

service/controller/resource/cleanup/resource.go
```diff
@@ -0,0 +1,132 @@
...
+const (
+	// Name is the identifier of the resource.
+	Name = "cleanup"
+)
+
+type Config struct {
+	Logger micrologger.Logger
+}
+
+// Resource manages Azure resource groups.
```
fiunchinho, line 23: Copy pasta

service/controller/resource/cleanup/resource.go
```diff
@@ -0,0 +1,132 @@
+package cleanup
```
fiunchinho, line 1: Would it make sense adding some information about cleaned resources in the `Status` field so we know if we need to try to delete them again? Otherwise we are going to try to find and delete these resources on every reconciliation loop. Not sure if important, but doesn't seem very efficient.

I was thinking about adding a `resource` to the `Status` field whenever we delete the dummy ip or whatever. That way we only run the cleanup logic if this is missing from the `Status` field.

Next release could delete this `cleanup` resource, or change it so that it removes the `cleanup` information from the `Status` field.

tuommaki: We might want to think about alternative solutions here. We used to do cleanup/migration this way, but given how frequently we have been doing releases for Azure and when this is even desired, I'm afraid there will be too often jumps over some versions in some clusters and those would then miss the cleanup.

Example:
- Customer is running cluster `11.2.4`.
- We change something for `11.3.0`. This one has cleanup for old way.
- Customer can't upgrade.
- We change something for `11.3.1`. This one gets cleanup deleted? Doesn't have to.
- Customer can't upgrade.
- We change something bigger for `11.4.0`. This one definitely gets cleanup deleted.
- Customer finally upgrades...but misses the cleanup.

We could maybe think about the general resource management architecture as we have been before. We already moved metrics collections to separate service. Maybe we could have some kind of migration operator with own CR. Then `azure-operator` could create the CR for each release it's been reconciling. It could then separately delete all CRs that have been marked as done and which have version from past, so deletion wouldn't be version specific but creation would be.

Meanwhile the migration operator could act the same: If version of CR is from the past of some given migration, then perform it and mark status.

For cleanup this should work easily, but could we also somehow leverage this for migrations?

whites11: Hey guys, thanks for your comments.
I don't like the cleaning resource as well, but I personally don't feel like it would be efficient if I work on a whole new operator for this.
Can we find a solution to move forward with the nat gateway (to unlock internal load balancers) and work on the cleaner separately?
The solution could be to simply not deleting those two resources (they don't hurt after all)

PR #821, opened 2020-05-07
Centralize methods to gather info from Release

fiunchinho:
We want an easy way to fetch info from a `Release`.

This also allows us to use it during e2e tests. I'm not changing the logic much, but with these changes we are reading components versions from the `Release` that we install during e2e tests, so that all the info about components is self-contained in the `Release` CR.
I create the `Release` as the first step of the setup and then pass it to other steps, so every step can fetch information from the `Release`.

### DIFF
--- a/service/controller/key/key.go
+++ b/service/controller/key/key.go
@@ -273,7 +273,11 @@ func MasterSecurityGroupName(customObject providerv1alpha1.AzureConfig) string {
 
 // OSVersion returns the version of the operating system.
 func OSVersion(release releasev1alpha1.Release) (string, error) {
-	return ComponentVersion(release, ContainerLinuxComponentName)
+	v, err := ComponentVersion(release, ContainerLinuxComponentName)
+	if err != nil {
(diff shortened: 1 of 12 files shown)


service/controller/key/key.go
```diff
@@ -260,6 +271,11 @@ func MasterSecurityGroupName(customObject providerv1alpha1.AzureConfig) string {
 	return fmt.Sprintf("%s-%s", ClusterID(customObject), masterSecurityGroupSuffix)
 }
 
+// OSVersion returns the version of the operating system.
+func OSVersion(release releasev1alpha1.Release) (string, error) {
+	return ComponentVersion(release, ContainerLinuxComponentName)
```
tuommaki, line 276: While I personally do like these for conciseness, we need to have the additional blocks to not swallow steps from stack trace:
```suggestion
	v, err := ComponentVersion(release, ContainerLinuxComponentName)
	if err != nil {
		return "", microerror.Mask(err)
	}
	return v, nil
```

PR #823, opened 2020-05-07
Vnet peering

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/10667

Uses vnet peering instead of VPN for cp to tc communication


service/controller/resource/vnetpeering/create.go
```diff
@@ -0,0 +1,229 @@
...
+			return microerror.Mask(err)
+		}
+		return nil
+	}
+
+	return nil
+}
+
+func (r *Resource) getCPVnetPeering(vnetId string) network.VirtualNetworkPeering {
+	t := true
```
fiunchinho, line 196: You can now use [the new functions](https://github.com/giantswarm/to/pull/7) instead of that.

fiunchinho:
@whites11 should we remove these? https://github.com/giantswarm/g8s-prometheus/blob/71e14d68e25022eccbaaf3672aaa503bee87b78c/helm/g8s-prometheus/prometheus-rules/azure.rules.yml#L72-L98

whites11:
> @whites11 should we remove these? https://github.com/giantswarm/g8s-prometheus/blob/71e14d68e25022eccbaaf3672aaa503bee87b78c/helm/g8s-prometheus/prometheus-rules/azure.rules.yml#L72-L98

Only when we'll have upgraded all clusters

service/controller/resource/vnetpeering/create.go
whites11: you did it, right?

fiunchinho: Yes

service/controller/resource/vnetpeering/create.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+
+		err = results.NextWithContext(ctx)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	if !found {
+		// No connections have been found, safe to delete the VPN Gateway.
```
tuommaki, line 183: What's the reasoning behind deleting connections first and then the GW? Is there any chance that cleanup gets stuck here because something reconciles deleted connection before GW gets deleted?

fiunchinho: Do we all agree on removing this cleaning logic and leave it until we create the vaccum cleaner? @tuommaki @whites11 @nprokopic

fiunchinho:
# How it works

With the changes in this PR a new multi-tenant service principal needs to be created (or update the current one) for every installation. `azure-operator` will stop using the `clientid` and `clientsecret` from the credential secret referenced in the `AzureConfig` CR: it will use the `clientid` and `clientsecret` given as parameters. It will still use the `subscriptionid` and `tenantid` from the credential secret, though.

This service principal is used to create both the control plane and tenant clusters. In order to do that, the service principal needs to assigned the `Owner` role in the subscription where the control plane or tenant cluster will run.

If the subscription is linked to a different tenant AD than the control plane subscription, the service principal needs to be invited to the tenant AD and assigned the `Owner` role in the subscription.

# Migration

Before deploying this version, the Service Principal used for the control plane needs to be updated to be multi-tenant. Everything keeps working.

## Upgrading cluster to this version

Before upgrading a cluster to this version, the customer needs to follow the docs on how to prepare an Azure subscription for tenant clusters, as if they were creating a cluster for the first time. In those docs we explain required steps for

- tenant cluster in same subscription as control plane: customer doesn't need to to anything.
- tenant cluster in different subscription, but same Tenant AD than control plane: customer needs to assign `Owner` role to GiantSwarm's Service Principal.
- tenant cluster in different subscription and different Tenant AD: customer needs to invite GiantSwarm's Service Principal to the Tenant AD and assign `Owner` role.

I don't see any changes between upgrading existing cluster vs creating new cluster.

service/controller/resource/vnetpeering/create.go
tuommaki: What's the cost? I guess that's the question mainly. I'm just wondering if you could just plain delete the GW? I mean, I'd expect the connections to get dropped at that point.

tuommaki:
> azure-operator will stop using the clientid and clientsecret from the credential secret referenced in the AzureConfig CR: it will use the clientid and clientsecret given as parameters. It will still use the subscriptionid and tenantid from the credential secret, though.

How does it find the right credential secret if it doesn't use the one referenced in AzureConfig?

> I'm not a security expert. But even though we could use the same Service Principal on all installations for all customers, it probably would be better to create a Service Principal for every installation.

Could we make this even more fine grained without usability suffering too much? The more there are clusters behind single principal, the more goes down when it expires and on the other hand the more it hurts if it somehow leaks. Neither of these should ever happen but...

fiunchinho:
> > azure-operator will stop using the clientid and clientsecret from the credential secret referenced in the AzureConfig CR: it will use the clientid and clientsecret given as parameters. It will still use the subscriptionid and tenantid from the credential secret, though.
> 
> How does it find the right credential secret if it doesn't use the one referenced in AzureConfig?

There is only one service principal: the one that's passed as parameter when starting the operator. There is no need to find it.

> > I'm not a security expert. But even though we could use the same Service Principal on all installations for all customers, it probably would be better to create a Service Principal for every installation.
> 
> Could we make this even more fine grained without usability suffering too much? The more there are clusters behind single principal, the more goes down when it expires and on the other hand the more it hurts if it somehow leaks. Neither of these should ever happen but...

I don't think it's possible. The whole point is to use a single service principal that we control.

nprokopic:
> There is only one service principal: the one that's passed as parameter when starting the operator. There is no need to find it.

I'm not sure I like the idea of passing credentials as parameters (not saying it's an issue, just a gut feeing).
Why not have it as a secret?

> > > I'm not a security expert. But even though we could use the same Service Principal on all installations for all customers, it probably would be better to create a Service Principal for every installation.
> > 
> > 
> > Could we make this even more fine grained without usability suffering too much? The more there are clusters behind single principal, the more goes down when it expires and on the other hand the more it hurts if it somehow leaks. Neither of these should ever happen but...
> 
> I don't think it's possible. The whole point is to use a single service principal that we control.

@tuommaki how much more fine-grained than installation-level? Per organization would maybe be a good idea.
@fiunchinho I agree with Tuomas here that we should limit the number of cluster behind a single service principal as much as possible (without hurting/breaking user experience). Currently service principals are scoped to the organization, so I would say we should keep it at least on that level.

fiunchinho:
> > There is only one service principal: the one that's passed as parameter when starting the operator. There is no need to find it.
> 
> I'm not sure I like the idea of passing credentials as parameters (not saying it's an issue, just a gut feeing).
> Why not have it as a secret?

Notice that when I say "as parameters", I mean a `Secret` that is used to populate environment variables.

tuommaki:
> > There is only one service principal: the one that's passed as parameter when starting the operator. There is no need to find it.
> 
> I'm not sure I like the idea of passing credentials as parameters (not saying it's an issue, just a gut feeing).
> Why not have it as a secret?
> 
> > > > I'm not a security expert. But even though we could use the same Service Principal on all installations for all customers, it probably would be better to create a Service Principal for every installation.
> > > 
> > > 
> > > Could we make this even more fine grained without usability suffering too much? The more there are clusters behind single principal, the more goes down when it expires and on the other hand the more it hurts if it somehow leaks. Neither of these should ever happen but...
> > 
> > 
> > I don't think it's possible. The whole point is to use a single service principal that we control.
> 
> @tuommaki how much more fine-grained than installation-level? Per organization would maybe be a good idea.
> @fiunchinho I agree with Tuomas here that we should limit the number of cluster behind a single service principal as much as possible (without hurting/breaking user experience). Currently service principals are scoped to the organization, so I would say we should keep it at least on that level.

Ideally per-cluster, but because customer has to configure the SP on their side, I think per-organization is good. That allows prod, pre-prod and dev to be separated properly.

fiunchinho:
After discussing during meeting, we decided to create tenant clusters using credentials found in organizations' secret. These credentials will belong to the GiantSwarm Active Directory.

Preparation work is being done https://github.com/giantswarm/azure-operator/pull/840

service/controller/resource/vnetpeering/create.go
tuommaki: We discussed. Difficult to test for sure. Minor thing. All good here.

PR #825, opened 2020-05-08
Unit test release resource

fiunchinho:


service/controller/resource/release/create.go
```diff
@@ -24,20 +24,20 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-	var release *v1alpha1.Release
+	var release v1alpha1.Release
 	{
 		releaseVersion := cr.Labels[label.ReleaseVersion]
 		if !strings.HasPrefix(releaseVersion, "v") {
 			releaseVersion = fmt.Sprintf("v%s", releaseVersion)
 		}
 
-		release, err = r.g8sClient.ReleaseV1alpha1().Releases().Get(releaseVersion, metav1.GetOptions{})
+		err = r.k8sClient.CtrlClient().Get(ctx, client.ObjectKey{Namespace: "", Name: releaseVersion}, &release)
```
fiunchinho, line 34: Use controller client, since we agreed we would be using it.

service/controller/resource/release/default_k8sclient.go
```diff
@@ -0,0 +1,79 @@
+package release
```
fiunchinho, line 1: should this be somewhere else?

.circleci/config.yml
```diff
@@ -59,8 +59,6 @@ workflows:
       # manual approval to run e2e tests.
       - hold:
           type: approval
-          requires:
```
fiunchinho, line 62: This was preventing us from approving the e2e tests execution until `push-azure-operator-to-control-plane-app-catalog` finished. Now they can be approved from the beginning of the workflow.

service/controller/resource/release/default_k8sclient.go
axbarsan: I would put it in a folder like `service/controller/resource/release/test` or `service/controller/resource/release/testing`

fiunchinho: Using similar location to [aws operator](https://github.com/giantswarm/aws-operator/blob/ha-ma/service/internal/unittest/default_k8sclient.go).

PR #831, opened 2020-05-14
Centralized state machine change logging

nprokopic:
Added centralized logging to the internal state machine. State machine change log entries will then be used to create state machine flowchart with [mermaid](https://mermaid-js.github.io/mermaid/#/).


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "4.1.0-dev"
+	version            = "4.1.0-nikola-flowcharts"
```
whites11, line 8: Better remove this from the PR

nprokopic: Yeah, wanted to remove it, thanks for the reminder :)

PR #833, opened 2020-05-15
Decouple auth from subscription

fiunchinho:
This PR tries to decouple the authentication method from the subscription where we will create/update/delete resources. **There should be no changes in behavior**. But it will help if we want to change how we handle Service Principals in the future (i.e having only one SP).

There are a couple of refactorings as well:
- use native azure/autorest objects instead of our own azure client set config struct
- When control plane azure clients are needed, they are passed as arguments when creating the operatorkit `resource`
- logic to read credentials from secret referenced by CR moved to `key` package.


service/controller/resource/dnsrecord/resource.go
```diff
@@ -17,27 +16,24 @@ const (
...
-	Logger                   micrologger.Logger
+	CPRecordSetsClient dns.RecordSetsClient
+	Logger             micrologger.Logger
 }
 
 // Resource manages Azure resource groups.
 type Resource struct {
-	hostAzureClientSetConfig client.AzureClientSetConfig
-	logger                   micrologger.Logger
+	cpRecordSetsClient dns.RecordSetsClient
```
fiunchinho, line 25: The control plane azure clients can be created once, instead of once for every reconciliation loop. In those cases we inject the client when creating the resource.

service/controller/resource/vpnconnection/resource.go
```diff
@@ -18,37 +17,33 @@ const (
 
 // Config is the configuration required by Resource.
 type Config struct {
-	Logger micrologger.Logger
-
-	Azure                    setting.Azure
-	HostAzureClientSetConfig client.AzureClientSetConfig
+	Azure                                    setting.Azure
+	Logger                                   micrologger.Logger
+	CPVirtualNetworkGatewaysClient           network.VirtualNetworkGatewaysClient
```
fiunchinho, line 22: [Same](https://github.com/giantswarm/azure-operator/pull/833/files#r425759089)

service/controller/resource_set.go
```diff
@@ -475,14 +484,17 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
 		}
 
-		guestAzureClientSetConfig.EnvironmentName = config.Azure.EnvironmentName
+		subscriptionID, partnerID, err := key.GetSubscriptionAndPartnerID(config.K8sClient.K8sClient(), key.CredentialName(cr), key.CredentialNamespace(cr))
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
 
-		azureClients, err := client.NewAzureClientSet(*guestAzureClientSetConfig)
+		azureClients, err := client.NewAzureClientSet(tenantAzureClientCredentialsConfig, subscriptionID, partnerID)
```
fiunchinho, line 497: On every reconciliation loop we create the azure client set with the tenant credentials. Same as before.

service/service.go
```diff
@@ -179,15 +173,16 @@ func New(config Config) (*Service, error) {
...
-			Ignition:         Ignition,
-			OIDC:             OIDC,
-			InstallationName: config.Viper.GetString(config.Flag.Service.Installation.Name),
-			ProjectName:      config.ProjectName,
-			RegistryDomain:   config.Viper.GetString(config.Flag.Service.RegistryDomain),
-			SSOPublicKey:     config.Viper.GetString(config.Flag.Service.Tenant.SSH.SSOPublicKey),
-			VMSSCheckWorkers: config.Viper.GetInt(config.Flag.Service.Azure.VMSSCheckWorkers),
+			Azure:                          azure,
+			CPAzureClientCredentialsConfig: cpAzureClientCredentialsConfig,
+			CPSubscriptionID:               config.Viper.GetString(config.Flag.Service.Azure.SubscriptionID),
```
fiunchinho, line 178: This is the key. We decouple the credentials used from the subscription where we will manage resources.

client/azure_client_set.go
```diff
@@ -152,211 +129,108 @@ func NewAzureClientSet(config AzureClientSetConfig) (*AzureClientSet, error) {
 	return clientSet, nil
 }
 
-func newDeploymentsClient(config *clientConfig) (*resources.DeploymentsClient, error) {
-	c := resources.NewDeploymentsClientWithBaseURI(config.resourceManagerEndpoint, config.subscriptionID)
-	c.Authorizer = autorest.NewBearerAuthorizer(config.servicePrincipalToken)
-	c.RetryAttempts = 1
```
fiunchinho, line 158: I removed this line from all the clients. I don't think it's necessary anymore?

service/controller/cloudconfig/base_extension.go
```diff
@@ -1,23 +1,24 @@
...
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/certs"
 
-	"github.com/giantswarm/azure-operator/v4/client"
 	"github.com/giantswarm/azure-operator/v4/service/controller/encrypter"
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 	"github.com/giantswarm/azure-operator/v4/service/controller/setting"
 )
 
 type baseExtension struct {
```
fiunchinho, line 13: Pass both `ClientCredentialsConfig` and `SubscriptionID`

service/controller/cloudconfig/base_extension.go
```diff
@@ -1,23 +1,24 @@
...
-	azure        setting.Azure
-	azureConfig  client.AzureClientSetConfig
-	calicoCIDR   string
-	clusterCerts certs.Cluster
-	customObject providerv1alpha1.AzureConfig
-	encrypter    encrypter.Interface
-	vnetCIDR     string
+	azure                        setting.Azure
```
fiunchinho, line 14: Pass both `ClientCredentialsConfig` and `SubscriptionID`

service/controller/cloudconfig/cloud_config.go
```diff
@@ -63,20 +64,26 @@ func New(config Config) (*CloudConfig, error) {
...
 		logger:             config.Logger,
 		randomkeysSearcher: config.RandomkeysSearcher,
 
-		azure:        config.Azure,
-		azureConfig:  config.AzureConfig,
-		azureNetwork: config.AzureNetwork,
-		ignition:     config.Ignition,
-		OIDC:         config.OIDC,
-		ssoPublicKey: config.SSOPublicKey,
+		azure:                  config.Azure,
```
fiunchinho, line 80: Pass both `ClientCredentialsConfig` and `SubscriptionID`

service/controller/resource_set.go
```diff
@@ -475,14 +473,21 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return nil, microerror.Mask(err)
 		}
 
-		guestAzureClientSetConfig, err := credential.GetAzureConfig(config.K8sClient.K8sClient(), key.CredentialName(cr), key.CredentialNamespace(cr))
+		tenantAzureClientCredentialsConfig, err := key.GetTenantAzureClientCredentialsConfig(config.K8sClient.K8sClient(), cr)
```
fiunchinho, line 476: On every reconciliation loop we still create the credentials for the reconciled CR.

service/controller/key/key.go
```diff
@@ -566,3 +570,68 @@ func vmssInstanceIDBase36(instanceID string) (string, error) {
...
+const (
+	ClientIDKey       = "azure.azureoperator.clientid"
+	ClientSecretKey   = "azure.azureoperator.clientsecret"
+	SubscriptionIDKey = "azure.azureoperator.subscriptionid"
+	TenantIDKey       = "azure.azureoperator.tenantid"
+	PartnerIDKey      = "azure.azureoperator.partnerid"
+	defaultAzureGUID  = "37f13270-5c7a-56ff-9211-8426baaeaabd"
+)
+
+func GetTenantAzureClientCredentialsConfig(k8sClient kubernetes.Interface, cr providerv1alpha1.AzureConfig) (auth.ClientCredentialsConfig, error) {
```
tuommaki, line 583: I think we solely use key functions as similar abstraction as entity class' getters. Usually we don't have any kind of client logic in key package, but I think it would be better to encapsulate it in own package somewhere. Maybe something like `pkg/credential` so these could become `credential.GetTenantAzureClientConfig` and `credential.GetSubscriptionAndPartnerID()` ?

client/azure_client_set.go
fiunchinho: Can I get a thumbs up on this comment @giantswarm/team-celestial ?

service/controller/key/key.go
fiunchinho: Done

client/azure_client_set.go
tuommaki: Shouldn't be a problem I think.

PR #836, opened 2020-05-18
Ensure certconfig CRD is created

fiunchinho:
`cert-operator` version `0.1.0` [got retagged](https://github.com/giantswarm/cert-operator/pull/278), and it no longer ensure the `CertConfig` CRD is created in the cluster, so we need to do it ourselves.


integration/setup/provider.go
```diff
@@ -109,17 +108,6 @@ func provider(ctx context.Context, config Config, giantSwarmRelease releasev1alp
 		}
 	}
 
-	{
```
fiunchinho, line 112: This wasn't needed.

integration/setup/release.go
```diff
@@ -28,17 +28,6 @@ func createGSReleaseContainingOperatorVersion(ctx context.Context, config Config
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "ensured Release CRD exists")
 	}
 
-	{
```
fiunchinho, line 31: This has been deprecated.

go.mod
```diff
@@ -26,6 +26,8 @@ require (
 	github.com/giantswarm/helmclient v0.2.0
 	github.com/giantswarm/ipam v0.2.0
 	github.com/giantswarm/k8sclient v0.2.0
+	github.com/giantswarm/k8sclient/v2 v2.0.0
+	github.com/giantswarm/k8sclient/v3 v3.0.0
```
fiunchinho, line 30: I need to use two different versions of the same library: we need to different versions until all the CRDs are graduated.

integration/setup/common.go
```diff
@@ -155,6 +155,17 @@ func installCertOperator(ctx context.Context, config Config, version string) err
         GuestAPI:
           Public: false
 `
+	{
```
fiunchinho, line 158: This is the most important change.

PR #837, opened 2020-05-18
Initial implementation to map CAPI CRs into AzureConfig

tuommaki:
This is initial implementation to provide migration path from legacy Giant
Swarm types towards Cluster API types using CAPZ types as source of truth.

This helps with development of Node Pools as tenant cluster control plane can
still reconcile AzureConfig while node pools are created from MachinePool and
AzureMachinePool.


tuommaki:
Ping @giantswarm/team-celestial - here is my preliminary work towards mapping the CRs. Don't spend time on review just yet as I need to test this (and probably do some fixes) but you can take a look at the approach here and give notes if you see anything outrageously stupid.

My thinking here is that this mapping resource lives only for migration period and once we are cool with Cluster API types (and customers have upgraded) we'll then drop this.

tuommaki:
This is more or less ready for review, but few fixes here & there arrive from `AzureClusterConfig` mapping PR which lands here.

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+
+		if len(masterMachines) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no control plane AzureMachines found for cluster %q", key.ClusterID(&cluster)))
+			r.logger.LogCtx(ctx, "level", "debug", "message", "cancelling reconciliation")
+			reconciliationcanceledcontext.SetCanceled(ctx)
+			return nil
+		}
+
+		if len(workerMachines) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("worker AzureMachines found for cluster %q", key.ClusterID(&cluster)))
```
fiunchinho, line 92:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no worker AzureMachines found for cluster %q", key.ClusterID(&cluster)))
```

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+
+func (r *Resource) buildAzureConfig(ctx context.Context, cluster capiv1alpha3.Cluster, azureCluster capzv1alpha3.AzureCluster, masters, workers []capzv1alpha3.AzureMachine) (providerv1alpha1.AzureConfig, error) {
+	var err error
+
+	azureConfig := providerv1alpha1.AzureConfig{}
+	azureConfig.Labels = make(map[string]string)
+
+	{
+		azureConfig.TypeMeta.APIVersion = azureAPIVersion
+		azureConfig.TypeMeta.Kind = "AzureConfig"
```
fiunchinho, line 169: It's the first time I see we set those fields. How come is required?

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+	{
+		azureConfig.Labels[label.ReleaseVersion] = key.ReleaseVersion(&cluster)
+	}
+
+	{
+		azureConfig.Labels[label.OperatorVersion] = key.OperatorVersion(&azureCluster)
+
+		azureConfig.Spec.Cluster.Kubernetes.Kubelet.Labels = ensureLabel(azureConfig.Spec.Cluster.Kubernetes.Kubelet.Labels, label.OperatorVersion, key.OperatorVersion(&azureCluster))
+		azureConfig.Spec.Cluster.Kubernetes.Kubelet.Labels = ensureLabel(azureConfig.Spec.Cluster.Kubernetes.Kubelet.Labels, "giantswarm.io/provider", ProviderAzure)
+		azureConfig.Spec.VersionBundle.Version = key.OperatorVersion(&azureCluster)
```
fiunchinho, line 197: Do we still need to fill this field?

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+				return nil, microerror.Mask(err)
+			}
+
+			azureConfig.Status.Cluster = clusterStatus
+		}
+	*/
+
+	var (
+		// hostDNSZone is created by stripping 3 first components from
+		// API domain are stipped, e.g.
```
fiunchinho, line 219:
```suggestion
		// hostDNSZone is created by stripping 3 first components from
		// API domain, e.g.
```

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	{
+		commonCluster.Masters = newSpecClusterMasterNodes()
+		commonCluster.Workers = newSpecClusterWorkerNodes()
+	}
+
+	{
+		commonCluster.Scaling.Max = len(workers)
+		commonCluster.Scaling.Min = len(workers)
```
fiunchinho, line 379: I remember you saying that there was not clear way of using auto scaling with ClusterAPI types? How will this look like if we start using cluster autoscaler?

service/controller/resource/azureconfig/create.go
```diff
@@ -0,0 +1,472 @@
...
+		if err != nil {
+			return providerv1alpha1.Cluster{}, microerror.Mask(err)
+		}
+
+		commonCluster.Kubernetes.SSH.UserList = userList
+	}
+
+	{
+		commonCluster.Masters = newSpecClusterMasterNodes()
+		commonCluster.Workers = newSpecClusterWorkerNodes()
```
fiunchinho, line 374: Wouldn't this always create 3 worker nodes independently of what the customer has configured? We could pass `len(workers)` as parameters to `newSpecClusterWorkerNodes` so that we know how many to create.

service/controller/resource/azureconfig/credential.go
```diff
@@ -0,0 +1,70 @@
...
+		return credentialSecret, nil
+	}
+
+	// If no credential secrets are found, we use the default.
+	credentialSecret := &v1alpha1.CredentialSecret{
+		Namespace: credentialNamespace,
+		Name:      credentialDefaultName,
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "did not find credential secret, using default secret")
```
fiunchinho, line 67: I believe it would less confusing if the operator would fail when it couldn't find the desired secret. If a non existing secret is chosen we want to know it as soon as possible, instead of having a fallback, that may not even be suitable for the customer subscription (wrong tenant AD).

service/controller/resource/azureconfig/ingress.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/giantswarm/microerror"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+)
+
+const (
+	ingressControllerDomainPrefix = "ingress"
+	ingressWildcardDomainPrefix   = "*"
+)
+
+func (r *Resource) newIngressDomain(cr capzv1alpha3.AzureCluster, domainPrefix string) (string, error) {
```
fiunchinho, line 15: Just by reading their names, it's hard to tell which one is which. With a sentence explaining what they do we don't have to go through the code. I believe it would benefitial to add docs to these functions.

service/controller/resource_set.go
```diff
@@ -537,6 +538,29 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+		tenantAzureClientCredentialsConfig, err := credential.GetTenantAzureClientCredentialsConfig(config.K8sClient.K8sClient(), cr)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		authorizer, err := tenantAzureClientCredentialsConfig.Authorizer()
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		subscriptionID, partnerID, err := credential.GetSubscriptionAndPartnerID(config.K8sClient.K8sClient(), cr)
```
fiunchinho, line 563: I think this is not needed? The next line already gets that information, and I think we don't need the `authorizer`.

main.go
```diff
@@ -128,6 +128,36 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.Tenant.PartnerID, "", "Partner ID used for the Control Plane cluster.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.VirtualNetwork, "", "Host cluster virtual network name.")
 	daemonCommand.PersistentFlags().String(f.Service.Azure.HostCluster.VirtualNetworkGateway, "", "Host cluster virtual network gateway name.")
+
+	daemonCommand.PersistentFlags().String(f.Service.Cluster.BaseDomain, "", "Cluster base domain without k8s/g8s prefixes.")
```
fiunchinho, line 132: It's probably impossible to have a good default for all of them, but having sane defaults for as many parameters as possible will help both e2e tests and running the operator locally.

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "4.1.0-dev"
+	version            = "4.1.0-tuomas"
```
fiunchinho, line 8: Not important. But this will probably be released as 5.0.0 right?

service/service.go
```diff
@@ -189,6 +194,32 @@ func New(config Config) (*Service, error) {
...
 	}
 
+	var azureClusterController *controller.AzureCluster
+	{
+		c := controller.AzureClusterConfig{
+			K8sClient: k8sClient,
+			Logger:    config.Logger,
+
+			Flag:  config.Flag,
+			Viper: config.Viper,
```
fiunchinho, line 204: I know there are a lot of new parameters, but I think it's good to reduce the dependency on `viper` by passing the required parameters instead of the whole thing. It allows us to isolate the `viper` dependency, making it easier to change in the future. And making the controller easier to test, because we'd need just the string parameters.

fiunchinho:
I was thinking that maybe I should [revert this one](https://github.com/giantswarm/azure-operator/pull/855) and open PR into your branch?

service/controller/resource/azureconfig/credential.go
tuommaki: A bulk of this mapping code is copied from `kubernetesd` which hasn't really been maintained for a long time. I started to fix this but quickly realized that this is part of the core logic from there. In case where a given organization doesn't have specifically defined credential secret for BYOC account, there are no credentials for that organization and it is normal then to use the default credentials.

What I try to say is that it's not an error per se in current design to not find a credential secret. It's just an indication that BYOC hasn't been configured for the given organization.

tuommaki:
As I proceeded with the other PR, this one got stale. Closing.

PR #838, opened 2020-05-18
State machine flowchart generation

nprokopic:
Two things added:
- HTML file with flowchart template
- bash script for generating flowchart

It works for release `11.4.0` and newer.

Usage of script:
```sh
./flowchart.sh azure-operator-master \
    --tenant-cluster $tenant_cluster_id \
    --open-browser \
    --kube-context $kubectl_context \
    --output $flowchart_output_path

# or
./flowchart.sh azure-operator-master -t $tenant_cluster_id -b \
    -c $kubectl_context \
    -o $flowchart_output_path
```

In the example `azure-operator-master` is the name of `App` CR. HTML file with the flowchart will be generated in the current directory and the default browser will be opened.

Notes about arguments/flags:
- `--tenant-cluster` | `-t` is optional, if it is specified the flowchart will include only state transitions for the specified tenant cluster, otherwise it will generate all possible transitions found in all tenant clusters reconciled by specified `azure-operator` `App`
- `--open-browser` | `-b` is optional, if it is specified the default browser will be opened, otherwise the browser is not opened.
- `--kube-context` | `-c` is optional, if it is specified flowchart will be generated for `App` and tenant cluster from that context, otherwise the current kube context will be used.
- `--output` | `-o` is optional, if it is specified the flowchart HTML file will be saved to the specified path, otherwise the output file is named `${azure_operator_app}.${tenant_cluster_id}.flowchart.generated.html` if tenant cluster is specified, or `${azure_operator_app}.flowchart.generated.html` if tenant cluster is not specified and it is saved in the current directory.

Example of generated flowchart:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ integration/test/**/logs/*.txt
 scripts/*.tar.gz
 
 telepresence.log
+*.flowchart.generated.html
--- a/scripts/flowchart.sh
+++ b/scripts/flowchart.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+#
+# Example usage:
+#   flowchart.sh azure-operator-master --tenant-cluster xyz99 --open-browser
+#
+# where 'azure-operator-master' is App CR name and 'xyz99' is tenant cluster ID
+# and '--open-browser' flag indicates that a default browser will open the
+# generated flowchart.
+#
+
+args=()
+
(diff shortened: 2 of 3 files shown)


nprokopic:
> Maybe add the generated reports to gitignore file? or do we want to keep an updated one committed?

Good idea to add it to gitignore just in case it ends up there. I'll also add the option to specify output file.

PR #840, opened 2020-05-19
Allow to use multi tenant SP or single tenant SP

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/11063

If I didn't screw up, these changes shouldn't need any planned release. Everything should work as it currently is. But they allow us to migrate customers to a multi tenant service principal.

# Control plane resources

Currently we pass credentials to the operator. These credentials belongs to a "Single Tenant" Service Principal in customer's Active Directory. We use those credentials when managing resources in the Control Plane cluster.

In this PR a couple of new CLI parameters have been added to the operator so that we can pass the Tenant id and Subscription id for the Control Plane cluster. That way we have a different set of parameters to manage the Control Plane resources.

The code is done so that ff these new parameters are not passed to the operator, we use customer's "single tenant" Service Principal to create the Azure client set to manage the Control Plane resources. **Same as we currently do.**

If the new parameters are passed to the operator, we then have tenants ids for both GiantSwarm and the customer. The code will create an Azure client set using GiantSwarm's "Multi Tenant" Service principal to manage the control plane resources.

# Tenant cluster resources

With the new code we will create an Azure client set to manage tenant cluster resources using:

- Organization/Customer credentials defined in organization secret when clientid/clientsecret are present . **Same as we do now**.
- GiantSwarm's "multi tenant" Service Principal (the one passed as parameter to operator) when no clientid/clientsecret are defined in organization credential secret.

# Migration
When we pass the control plane tenant CLI parameters, the multi-tenant Service Principal will be used for managing the Control Plane resources.

When clientid/clientsecret are removed from the organization credential secret, the multi-tenant Service Principal will be used for managing the Tenant Cluster resources.


service/controller/resource_set.go
```diff
@@ -488,7 +486,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return nil, microerror.Mask(err)
 		}
 
-		azureClients, err := client.NewAzureClientSetWithAuthorizer(authorizer, subscriptionID, partnerID)
+		azureClients, err := client.NewAzureClientSet(organizationAzureClientCredentialsConfig, subscriptionID, partnerID)
```
fiunchinho, line 489: With the new code we will create an Azure client set using:
- Organization/Customer credentials when there is clientid/clientsecret defined in organization secret (same as we do now)
- GiantSwarm multi tenant service principal (the one passed as parameter to operator) when no clientid/clientsecret are defined in organization secret

service/service.go
```diff
@@ -269,3 +271,51 @@ func buildK8sRestConfig(config Config) (*rest.Config, error) {
...
+	if config.Viper.GetString(config.Flag.Service.Azure.SubscriptionID) != "" {
+		settings.Values[auth.SubscriptionID] = config.Viper.GetString(config.Flag.Service.Azure.SubscriptionID)
+	}
+
+	return settings.GetClientCredentials()
+}
+
+// NewCPAzureClientSet return an Azure client set configured for the control plane cluster.
+// If no control plane cluster tenant information has been passed to the operator, auth Azure Tenant will be used as control plane Azure Tenant.
+func NewCPAzureClientSet(config Config, gsClientCredentialsConfig auth.ClientCredentialsConfig) (*client.AzureClientSet, error) {
```
fiunchinho, line 301: A couple of new CLI parameters have been added to the operator so that we can pass the tenant id and subscription id for the control plane cluster.

If these new parameters are not passed to the operator, we use customer's single tenant service principal to create the Azure client set to manage the control plane. Same as we currently do.

If these new parameters are passed to the operator, we have tenants ids for both GiantSwarm and the customer. The code will create an Azure client set using GiantSwarm's Multi Tenant Service principal to manage the control plane.

service/service.go
```diff
@@ -269,3 +271,51 @@ func buildK8sRestConfig(config Config) (*rest.Config, error) {
...
+	cpTenantID := config.Viper.GetString(config.Flag.Service.Azure.HostCluster.Tenant.TenantID)
+	if cpTenantID == "" {
+		cpTenantID = config.Viper.GetString(config.Flag.Service.Azure.TenantID)
+		// We want the code to work both when using Single Tenant Service Principal and Multi Tenant Service Principal,
+		// so we only add the CP Tenant ID as auxiliary id if an explicit CP Tenant ID has been passed.
+		gsClientCredentialsConfig.AuxTenants = append(gsClientCredentialsConfig.AuxTenants, cpTenantID)
```
fiunchinho, line 307: A couple of new CLI parameters have been added to the operator so that we can pass the tenant id and subscription id for the control plane cluster.

If these new parameters are not passed to the operator, we use customer's single tenant service principal to create the Azure client set to manage the control plane. Same as we currently do.

If these new parameters are passed to the operator, we have tenants ids for both GiantSwarm and the customer. The code will create an Azure client set using GiantSwarm's Multi Tenant Service principal to manage the control plane.

client/azure_client_set.go
```diff
@@ -52,35 +52,13 @@ type AzureClientSet struct {
 	VnetPeeringClient *network.VirtualNetworkPeeringsClient
 }
 
-// NewAzureClientSet returns the Azure API clients.
-// Auth is configured taking values from Environment, but parameters have precedence over environment variables.
-func NewAzureClientSet(clientid, clientsecret, tenantid, subscriptionID, partnerID string) (*AzureClientSet, error) {
```
fiunchinho, line 57: Moved to `service.go`. There we have all the inputs: environment and CLI parameters.

pkg/credential/credential.go
```diff
@@ -20,28 +20,61 @@ const (
...
+// so that a Multi-Tenant Service Principal can work on the subscription linked to the organization Tenant ID.
+// It will fallback to a Single Tenant Service Principal when organization's credential secret contains Client ID and Client Secret.
+func GetOrganizationAzureClientCredentialsConfig(k8sClient kubernetes.Interface, cr providerv1alpha1.AzureConfig, gsClientCredentialsConfig auth.ClientCredentialsConfig) (auth.ClientCredentialsConfig, error) {
 	credential, err := k8sClient.CoreV1().Secrets(key.CredentialNamespace(cr)).Get(key.CredentialName(cr), apismetav1.GetOptions{})
 	if err != nil {
 		return auth.ClientCredentialsConfig{}, microerror.Mask(err)
 	}
 
-	clientID, err := valueFromSecret(credential, ClientIDKey)
+	clientID, _ := valueFromSecret(credential, ClientIDKey)
```
fiunchinho, line 32: We no longer fail if credentials are missing from credential secret. They being empty is the indicator that we should use multi tenant Service Principal.

MartaD: will it fail if multi tenant Service Principal is also not set or it will try to use single tenant set up again? (sorry if it's a stupid question!)

fiunchinho: Not stupid at all! The parameters are passed when installing, and the values come from the installations repository.

We are not checking if those parameters contain a value. We could add it.

fiunchinho: Fixed here @MartaD https://github.com/giantswarm/azure-operator/pull/840/commits/24800876858720cb4c3cb51e143f7783d332926f#diff-eeade751807b5e894d1f9a5b21390878R296

fiunchinho: We changed the mechanism to indicate that we should use multi tenant Service Principal: if the organization secret has a label, it's single tenant.

fiunchinho: I've changed it again :joy: . If the tenant id in the organization secret is the same as the tenant id passed as argument to azure-operator for authentication, then it's single tenant.

integration/setup/config.go
```diff
@@ -144,10 +144,11 @@ func NewConfig() (Config, error) {
...
 		c := helmclient.Config{
-			Logger:          logger,
-			K8sClient:       cpK8sClients.K8sClient(),
-			RestConfig:      host.RestConfig(),
-			TillerNamespace: tillerNamespace,
+			Logger:               logger,
+			K8sClient:            cpK8sClients.K8sClient(),
+			RestConfig:           host.RestConfig(),
+			TillerNamespace:      tillerNamespace,
+			TillerUpgradeEnabled: true,
```
nprokopic, line 151: What's the purpose of this change? (sorry if the question is stupid :smile:)

service/service.go
```diff
@@ -178,19 +168,34 @@ func New(config Config) (*Service, error) {
 
 	var clusterController *controller.Cluster
 	{
+		// These credentials will be used when creating AzureClients for Control Plane clusters.
+		gsClientCredentialsConfig, err := credential.NewAzureCredentials(
+			config.Viper.GetString(config.Flag.Service.Azure.ClientID),
+			config.Viper.GetString(config.Flag.Service.Azure.ClientSecret),
+			config.Viper.GetString(config.Flag.Service.Azure.TenantID),
```
nprokopic, line 175: If I understood correctly, in `credential.NewAzureCredentials(...)` the credentials passed in from CLI arguments will have precedence over environment variables, right?
In that case I would maybe make it more explicit by moving these 3 lines that are reading the flags to be inside of the function. Not a blocker :)

integration/setup/config.go
fiunchinho: Oh that's actually not needed. I was testing something, I'll remove it.

service/service.go
fiunchinho: I didn't want to make the credential package to depend on Viper or whatever we use for the CLI parameters. Also, I believe by reading the code as it is now, it looks like you are getting credentials for those parameters values, and that's what we are doing. It's just that if the rare case that they are empty, we will try to use env vars, but you as a client don't need to care because you are passing the parameters. Does that make sense?

flag/service/azure/hostcluster/tenant/tenant.go
```diff
@@ -0,0 +1,7 @@
+package tenant
+
+type Tenant struct {
```
whites11, line 3: Just a silly thing. I don't think "subscription" is a field of "Tenant" in the azure naming.
They kinda have a 1:n relationship.
But guess it doesn't make confusion, we can leave it like it is I guess.

service/service.go
nprokopic: Sounds good!

flag/service/azure/hostcluster/tenant/tenant.go
fiunchinho: Correct. I didn't know which name to use. I needed a term that groups together tenantid, subscriptionid and partnerid. I thought about a GiantSwarm tenant, and I chose that. But I'm not sure.

whites11: Something related to an `organization`? Like `OrganizationAzureSettings`?

fiunchinho: It's used for the control plane, so I really don't like organization.

whites11: ok Tenant is good enough, let's not overcomplicate things

PR #841, opened 2020-05-19
Allow the suffix to be overriden while building the operator binary

fiunchinho:
The binary can then be built like 
```
$ go build -a -v -ldflags "-w -linkmode 'auto' -extldflags '-static' -X 'github.com/giantswarm/azure-operator/v4/pkg/project.wipSuffix=-some-branch' -X 'github.com/giantswarm/azure-operator/v4/pkg/project.gitSHA=1a2b3c4d'"

$ ./azure-operator version
description: The azure-operator manages Kubernetes clusters on Azure.
gitCommit: 1a2b3c4d
name: azure-operator
source: https://github.com/giantswarm/azure-operator
goVersion: go1.14.2
os: linux
arch: amd64
version: 4.1.0-some-branch
```


fiunchinho:
@kopiczko what do you think about this idea?

Our architect-orb sets the `suffix` depending on the scenario:
- When building master, the suffix is `- dev` to keep consistency with what we have so far.
- When building a PR branch, the suffix is the name of the branch.
- When building a tag, the suffix is removed.

This would allow people working on the same operator to deploy their own branches without having to worry about operators reconciling the same versions.

If we take this further, while developing, we could automate the creation of a wip `Release` that would contain the component being built at the same version that architect-orb used. Something like the last step in the Pull Request pipeline is creating a `Release`, or when we run `opsctl deploy azure-operator@branch` then it creates the `Release` with `azure-operator` at version `$projectVersion$projectSuffix`. Something along those lines.

kopiczko:
I see where you're coming from. I think it's similar to https://github.com/giantswarm/giantswarm/issues/9850. It suggests injecting the same version as in chart and creating a Release CR with `opsctl (un)deploy`.

TBH I'm torn apart. I think if we want to inject anything we should inject whole version otherwise the thing gets confusing. And why should we both inject and edit? I personally have a slight preference towards having the version set in the code because then `BINARY version` output is independent of the build command that was executed against the codebase which makes the builds more reproducible.

Also with that you'll have to create a release for each branch. A thing we wanted to avoid initially. Especially that it was really difficult w/o Release CRs we have now. I wouldn't like to have that merged w/o having `opsctl deploy`/`opsctl undeploy` doing this (look at the issue again). But then we'd actually do most of https://github.com/giantswarm/giantswarm/issues/9850.

Also I think we're fighting the symptom here which is messy environments. We should actually work on making them clean because with that the problem will become even worse.

And most importantly we'd need to have a broader agreement here and keep things consistent between repositories.

fiunchinho:
Closing. We decided to have the version on a separate file.

PR #842, opened 2020-05-19
Use helmrelease package from e2esetup

rossf7:
Towards giantswarm/giantswarm#11035

Now helmclient 1.0.0 is released with Helm 3 we need to update the integration tests. I'm also trying to reduce our usage of helmclient as it got out of control.

The release package was moved to e2esetup and this lets us drop helmclient from e2e-harness.  🎉


integration/setup/common.go
```diff
@@ -8,7 +8,7 @@ import (
 	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
 	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
 	"github.com/giantswarm/backoff"
-	"github.com/giantswarm/e2e-harness/pkg/release"
+	"github.com/giantswarm/e2esetup/helmrelease"
```
rossf7, line 11: Interface is the same. Only the package name changes.

fiunchinho:
@rossf7 Can you fix conflicts and re-run e2e tests?

rossf7:
@fiunchinho Thanks I just tried this but it won't work. I need to update e2esetup due to the changes to k8sclient.

I'll come back to this later when I have more time.

whites11:
@rossf7 can we close this PR?

rossf7:
@whites11 Yes totally. This is very stale. We would still like to remove helmclient from the integration tests but not sure when I can work on this.

PR #844, opened 2020-05-20
IPAM resource import from AWS

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/10783

This draft PR is about calculating the subnet to be used by a tenant cluster in the azure operator, by collecting the in-use CIDRs out of a few sources:

- the `AzureConfig` CRs
- the `VNET`'s address space for selected resource groups.


service/controller/resource/ipam/azure_config_persister.go
```diff
@@ -0,0 +1,53 @@
...
+	return p, nil
+}
+
+func (p *AzureConfigPersister) Persist(ctx context.Context, subnet net.IPNet, namespace string, name string) error {
+	cr, err := p.g8sClient.ProviderV1alpha1().AzureConfigs(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	cr.Spec.Azure.VirtualNetwork.CIDR = subnet.String()
```
tuommaki, line 45: In addition to this, I think you need to split the VNET CIDR into POD network and further into {master,worker} subnets.

i.e. Let `subnet` be `10.1.0.0/23`. Split that into half and you get:
- `10.1.0.0/24` for PODs (i.e. `calicoSubnet`)
- `10.1.1.0/24` for nodes.
  - Now split this this into two halves for masters and workers:
    - `10.1.1.0/25` for masters
    - `10.1.1.128/25` for workers

...and store them in CR correspondingly. You can see the current implementation here: https://github.com/giantswarm/kubernetesd/blob/master/service/creator/azure.go#L99-L137

Note that `s.IPAM.CreateSubnet()` performs the allocation and persisting in `crdstorage` - we don't need that.

service/controller/resource/ipam/subnet_collector.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		for _, vnetName := range vnetCandidates {
+			vnet, err := vnetClient.Get(ctx, *group.Name, vnetName, "")
+			if IsNotFound(err) {
+				// VNET with desired name not found, ignore this resource group.
+			} else if err != nil {
+				return nil, microerror.Mask(err)
+			} else {
+				for _, cidr := range *vnet.AddressSpace.AddressPrefixes {
```
tuommaki, line 208: Nice! I think this should now also cover the node pool subnets that are added to tenant cluster vnet, right?

service/controller/resource/deployment/deployment.go
```diff
@@ -2,36 +2,40 @@ package deployment
...
 
-	"github.com/giantswarm/azure-operator/v4/service/controller/controllercontext"
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 	"github.com/giantswarm/azure-operator/v4/service/controller/resource/deployment/template"
+	"github.com/giantswarm/azure-operator/v4/service/network"
 )
 
 func (r Resource) newDeployment(ctx context.Context, customObject providerv1alpha1.AzureConfig, overwrites map[string]interface{}) (azureresource.Deployment, error) {
-	cc, err := controllercontext.FromContext(ctx)
+	// The VPN subnet is not persisted in the AzureConfig so I have to compute it now.
```
fiunchinho, line 17: Should we address this or should we wait until ClusterAPI? @tuommaki

tuommaki: I don't fully get how this is sub-optimal? When we can compute something like this, I think we should. The computation should be deterministic and cheap by default.

whites11:
@tuommaki the VPN subnet calculation is suboptimal because it is different from the other subnets.
The other ones are calculated once and stored in the CR, the VPN one is calculated all the times.
We should add it to the CR, but since it will be unuseful soon that is not a smart thing to do IMHO.

WDYT?

service/controller/resource/ipam/subnet_collector.go
whites11: yes

service/controller/resource/deployment/deployment.go
whites11: @tuommaki the VPN subnet calculation is suboptimal because it is different from the other subnets.
The other ones are calculated once and stored in the CR, the VPN one is calculated all the times.
We should add it to the CR, but since it will be unuseful soon that is not a smart thing to do IMHO.

WDYT?

tuommaki: When it's about to be dropped soon anyway, just leave it like that.

fiunchinho: agreed

service/controller/resource/ipam/subnet_collector.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		// We want to check only once per subscription.
+		if inArray(doneSubscriptions, clientSet.SubscriptionID) {
+			continue
+		}
+
+		nets, err := c.getSubnetsFromSubscription(ctx, clientSet)
+		if err != nil {
+			// We can't use this Azure credentials. Might be wrong in the Secret file.
```
whites11, line 167: This is the real doubt I have. Is this the right way of threating this case?

fiunchinho:
This PR depends on the operator receiving the new parameter `f.Service.Installation.Guest.IPAM.Network.CIDR`. But this parameter is not being passed anywhere? Shouldn't this parameter be added to the configmap created by [the helm chart](https://github.com/giantswarm/azure-operator/blob/73faaf6c25b00958688a52e3d1d317541b9035b2/helm/azure-operator/templates/configmap.yaml#L29)?

fiunchinho:
I've created https://github.com/giantswarm/azure-operator/pull/852 to talk about it.

PR #849, opened 2020-05-21
Merge 4.0.1 changes to master branch

whites11:
Merge 4.0.1 changes to master branch

### DIFF
--- a/helm/azure-operator/Chart.yaml
+++ b/helm/azure-operator/Chart.yaml
@@ -1,6 +1,6 @@
 name: azure-operator
 version: [[ .Version ]]
 description: "Azure operator to create k8s clusters on Azure"
-appVersion: [[ .AppVersion ]]
+appVersion: "[[ .AppVersion ]]"
 home: "https://github.com/giantswarm/azure-operator"
 apiVersion: v1
--- a/helm/azure-operator/Chart.yaml
+++ b/helm/azure-operator/Chart.yaml
@@ -1,6 +1,6 @@
 name: azure-operator
 version: [[ .Version ]]
 description: "Azure operator to create k8s clusters on Azure"
-appVersion: "[[ .AppVersion ]]"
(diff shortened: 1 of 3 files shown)


helm/azure-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: azure-operator
 version: [[ .Version ]]
 description: "Azure operator to create k8s clusters on Azure"
-appVersion: "[[ .AppVersion ]]"
+appVersion: [[ .AppVersion ]]
```
nprokopic, line 4: Do we maybe need quotes here, i.e. `"[[ .AppVersion ]]"`?

nprokopic:
What about [this last commit](https://github.com/giantswarm/azure-operator/commit/89d4c6d50b70a98ed56b15973a7513aa62bbda9f) from [`release-4.0.1` branch](https://github.com/giantswarm/azure-operator/commits/release-4-0-1)?

helm/azure-operator/Chart.yaml
whites11: Dunno, added back just in case

whites11:
> What about [this last commit](https://github.com/giantswarm/azure-operator/commit/89d4c6d50b70a98ed56b15973a7513aa62bbda9f) from [`release-4.0.1` branch](https://github.com/giantswarm/azure-operator/commits/release-4-0-1)?

Very good catch Nikola.
I noticed it was missing as well, then I remembered that I already catched that error a few days ago and fixed it in `master`. https://github.com/giantswarm/azure-operator/blob/master/service/controller/resource/instance/create_wait_for_nodes_to_become_ready.go#L44
I didn't backport it to 4.0.0 because it was only affecting upgrades from flatcar to flatcar, and at the time `master` was going to be the new release.
Since we had a patch release 4.0.1 that fix got lost.

PR #850, opened 2020-05-22
Calculate latest operator release version using architect

fiunchinho:
- Use latest architect-orb version so we don't have to install Helm during e2e tests.
- Use a bigger CircleCI machine.
- Use architect to calculate latest operator release, instead of requesting that info from catalog.

### DIFF
--- a/integration/config/setup.sh
+++ b/integration/config/setup.sh
@@ -1,20 +1,11 @@
 #!/usr/bin/env sh
 
-shipyard_dir=".e2e-harness/workdir/.shipyard"
-mkdir -p "${shipyard_dir}"
-chmod -R 777 ".e2e-harness"
-kind get kubeconfig --name="kind" | sudo tee "${shipyard_dir}/config" > /dev/null
-
-
 shipyard_dir="/workdir/.shipyard"
 sudo mkdir -p "${shipyard_dir}"
 sudo chmod -R 777 "/workdir"
-kind get kubeconfig --name="kind" > "${shipyard_dir}/config"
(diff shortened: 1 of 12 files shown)


integration/config/setup.sh
```diff
@@ -1,20 +1,11 @@
...
 sudo mkdir -p "${shipyard_dir}"
 sudo chmod -R 777 "/workdir"
-kind get kubeconfig --name="kind" > "${shipyard_dir}/config"
+kind get kubeconfig --name="kind" >"${shipyard_dir}/config"
 
-curl -L https://get.helm.sh/helm-v2.16.1-linux-amd64.tar.gz > ./helm.tar.gz
-tar xzvf helm.tar.gz
-chmod u+x linux-amd64/helm
-sudo mv linux-amd64/helm /usr/local/bin/
+GO111MODULE=on go get github.com/giantswarm/architect
```
fiunchinho, line 8: We need `architect` installed now.

integration/env/azure.go
```diff
@@ -26,6 +26,8 @@ const (
 	EnvVarBastionPublicSSHKey       = "BASTION_PUBLIC_SSH_KEY"
 
 	EnvVarCircleBuildNumber = "CIRCLE_BUILD_NUM"
+
+	EnvVarLatestOperatorRelease = "LATEST_OPERATOR_RELEASE"
```
fiunchinho, line 30: Instead of calling our catalog to figure out the operator latest release, we read this info from an environment variable that was set using `architect`.

PR #852, opened 2020-05-23
Pass IPAM network CIDR as parameter to the operator

fiunchinho:
I used the CIDR in the installations repository as default.

### DIFF
--- a/helm/azure-operator/templates/configmap.yaml
+++ b/helm/azure-operator/templates/configmap.yaml
@@ -27,6 +27,10 @@ data:
       installation:
         name: '{{ .Values.Installation.V1.Name }}'
       {{- if .Values.Installation.V1.Guest }}
+        guest:
+          IPAM:
+            Network:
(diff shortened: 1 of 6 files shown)


helm/azure-operator/templates/configmap.yaml
```diff
@@ -27,6 +27,11 @@ data:
       installation:
         name: '{{ .Values.Installation.V1.Name }}'
       {{- if .Values.Installation.V1.Guest }}
```
tuommaki, line 29: Not part of this PR, but what's this if? Naming suggest like this is adaptation for CP and TC app, but it doesn't make sense for `azure-operator`. Something else?

fiunchinho: Not sure. Maybe it was to check if OIDC parameters were being passed?

tuommaki: :man_shrugging:

PR #853, opened 2020-05-25
Refactoring masters and instance (workers) resources

nprokopic:
First pull request towards refactoring and code deduplication between masters and instance.

The general idea is to have a shared `nodes` package with common base for `Config` and `Resource` that has common functionality for all things related to nodes/instances, nothing specific to masters or workers.

The common `Config` and `Resource` then is embedded into `masters` and `instance` `Config` and `Resource`.

Here we have some shared functionality that was almost identical code:
- created common `nodes` resource with codebase that is shared between masters and workers (instance)
- moved `isClusterCreating` function to `nodes` packages
- moved `anyNodesOutOfDate` function to `nodes` packages
- moved `allInstances` function to `nodes` packages
- moved `createDrainerConfig` function to `nodes` packages
- moved `getResourceStatus` function to `nodes` packages
- moved `setResourceStatus` function to `nodes` packages
- removed unused errors

More stuff can be moved with a bit of refactoring, but I figured it would be good idea to merge this as soon as possible, as some functions are moved to a different package (it's also easier to review :smiley:).


service/controller/resource/nodes/create.go
```diff
@@ -0,0 +1,20 @@
+package nodes
+
+import (
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+)
+
+// IsClusterCreating check if the cluster is being created.
+// TODO: move to service
+func IsClusterCreating(cr providerv1alpha1.AzureConfig) bool {
```
fiunchinho, line 9: Maybe this one could live in the `key` package?

nprokopic: It can be moved outside of resource definitely, also `func AnyNodesOutOfDate(ctx context.Context) (bool, error)`.

Now when you say it `key` seems to be a nice option. I was also thinking about maybe some package/service under `service/controller/internal`. @tuommaki any thoughts here?

fiunchinho: We use the `key` package for simple getter functions. Not sure if `InClusterCreating` has too much business logic, I think not, but `AnyNodesOutOfDate` has too much business logic to be in the `key` package, I believe.

service/controller/resource/nodes/resource.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (r *Resource) G8sClient() versioned.Interface {
+	return r.g8sClient
+}
+
+func (r *Resource) InstanceWatchdog() vmsscheck.InstanceWatchdog {
+	return r.instanceWatchdog
+}
+
+func (r *Resource) Logger() micrologger.Logger {
```
tuommaki, line 100: Do you have some other use for these getters than just accessing the base struct fields? I don't think they are needed for that and especially the logging lines earlier in this PR look a bit unusual.

See https://play.golang.org/p/apUb8sVz26T

Could we live without these and just directly use the fields since they are embedded in the "derived" struct?

service/controller/resource/nodes/create.go
nprokopic: Moved `IsClusterCreating()` to `key` package.

service/controller/resource/nodes/resource.go
nprokopic: It's probably more due to my brain's wiring from other programming languages than to some strict idea. :smile: 

The reasoning here was to leave fields in base struct private and expose them via getter functions, but since this is the case of a straightforward embedding, without some business logic around getting those fields, making fields public in the base struct and exposing them is simpler and clearer. Changing it now to exposed `.Logger` field in the base struct.

nprokopic: Updated to use exposed `.Logger` field (did the same for other fields / getters).

service/controller/resource/nodes/clients.go
```diff
@@ -0,0 +1,57 @@
+package nodes
```
fiunchinho, line 1: Sorry, I know I'm a pain in the ass, but I'm not fully convinced into this approach.

It's harder for me to understand the code now with the "inherited" fields. I think it will be harder to test and modify, since changes on the "parent" will affect on both resources now. Specially when it seems master will differ from workers in the near future.

I was thinking about using composition creating objects that allow us to reuse functionality. For example, if we are going to move away of controllercontext in the future, I think it would be a good idea to have a factory that returns azure clients. The logic would live on the factory. Resources just depend on the factory to fetch their clients.
I'd propose something similar for the `GetEncrypterObject` function.

For functions like `AllInstances` or `AnyOutOfDate` we could place them as independent functions in the `vmss` package? (`pkg/helpers/vmss`)

Does it make sense?

This way we could test each of those pieces in isolation and re-use them wherever needed.

nprokopic: > with the "inherited" fields.
> I was thinking about using composition creating objects that allow us to reuse functionality.

Technically speaking, this **is** composition :smiley:  i.e. embedding one type into another, there is no "inheritance" :) It's just golang's syntactic sugar that let's you use fields and functions from the `nodes.Resource` type as if they were in `masters.Resource` / `instance.Resource`.

Another thing that led me to this approach is the fact that `masters.Resource`/`masters.Config` and `instance.Resource`/`instance.Config` are 100% identical structures.

> I think it will be harder to test and modify, since changes on the "parent" will affect on both resources now. Specially when it seems master will differ from workers in the near future.

The idea is that you do not have to think about masters and workers here at all, it's just some set of nodes/machines. Basically separating generic VMs/VMSS/instances/nodes stuff into a single shared package, so that we can reuse it and not copy-paste same code around.

I even had a bit "conservative" approach here, for example:
- `masters/deployment.go` and `instance/deployment.go` are basically identical, it's practically a parametrized VMSS deployment (the only difference is resource naming)
- some reconciliation functions are still the same/identical, for example `DeployementUninitialized`, `DeployementInitialized`, `ProvisioningSuccessful`, `DeploymentCompleted` (and probably few more) transitions are basically copy-paste with only difference around Azure resource naming.

:point_up_2: these could also be deduplicated easily, since they are about reconciliation of a VMSS deployment, nothing masters-/workers-specific there, it's just deployment of compute things.

> For example, if we are going to move away of controllercontext in the future, I think it would be a good idea to have a factory that returns azure clients. The logic would live on the factory. Resources just depend on the factory to fetch their clients.

Some stuff now ended up here just because it was duplicated in `masters` and `instance` packages.

Yeah, clients can definitely be moved outside of controllers/resources into some factory. Will do that and update this PR.

> I'd propose something similar for the GetEncrypterObject function.

`GetEncrypterObject` is not really "static"/global, it's using `Logger` and `k8sClient` fields of the resource, so that would require a bit more refactoring if we were to move it outside of `nodes` package.

> For functions like AllInstances or AnyOutOfDate we could place them as independent functions in the vmss package? (pkg/helpers/vmss)

`nodes` package is supposed to be a kind of "helpers" package, but resource/reconciliation aware. Other option for name I had in mind was `machines`, maybe that would be better since `masters` and `instance` packages are about both Kubernetes nodes and Azure VMSS/instances, either `nodes` or `vmss` is maybe kind of limiting.
I'll take a look into `pkg/helpers/vmss` package.

Generally speaking, with splitting `masters` and `instance` into separate resources, we do have separate deployments now, which is much better, but good part of reconciliation is still largely the same, and once we remove Flatcar-related workarounds, the difference will be even less.

IMHO, when it comes to nodes/VMSS reconciliation, ideally we would have these 3 layers of packages, where the each layer is using the one above/before it:
1. nodes/machines services, factories, utilities, basically a set of functionalities for working with Kubernetes nodes and Azure VMSS (e.g. you should not write multiple times the same code to check if the nodes are creating, out of date, etc.)
2. `Azure VMSS` <-> `Kubernetes nodes` reconciliation: e.g. you should not write multiple times the same code to check if a VMSS deployment has changed, or to create/update a deployment, etc.
3. `masters` and `workers`/`instance` packages, with `masters` and `workers` specific reconciliation.

This would kind of make us think what we want to change. Is it about some Azure client thing, some node state check -> layer 1 above. Reconciling some new flag/property for Azure VMSS -> layer 2. Some etcd related change -> `masters` in layer 3, etc.

On the other hand, if we move away from VMSS to VMs for masters, than that is a different story, but it's questionable if and when will that happen :)

nprokopic: In this 3-layer separation, layer 1 would be about talking to Kubernetes and Azure. While layers 2 and 3 would be about deciding when to talk and who to talk to.

Also if we are not sure if some functionality/behaviour is shared between masters and workers, or only specific to masters or workers only, then I would say we need to refine that task to make that clear.

nprokopic: > While layers 2 and 3 would be about deciding when to talk and who to talk to.

For example, when updating a VMSS deployment, we should not care about some ARM template or deployment specifics, ideally we would have some simple code with some simple logic, focus on reconciliation logic.

Let's take `deploymentInitializedTransition` for example, it comes down to:
```golang
if deployment.IsSucceeded {
    // all done, continue
    return ProvisioningSuccessful
} else if deployment.IsError {
    // error, start from the scratch
    return DeploymentUninitialized
} else {
    // still working
    return currentState
}
```

Switch would maybe make it even more straightforward :)

And it should be basically as simple as that, and not repeating that code every time you want to check a deployment and decide what to do next.

`deploymentUninitializedTransition` is not much different (some other as well), in terms of having a simple reconciliation logic (_layers 2 and 3_), with non-reconciliation logistics (_layer 1_) being out if your hair.

tuommaki: I like the direction where this is going. Few thoughts though:
- Can we limit changes to some manageable size here and settle with state that we can take further in next steps, without fully refactoring everything now?
- What do you think if we actually tried to implement the node pools resources in TDD fashion  in the sense that we try to create some kind of integration and unit tests for the whole implementation while actually doing it? I figure that would help us in the future a lot and might help to solve some concerns here.
- The client factory idea is something we need to drive. That solves the problems I had with metrics collection, but it also makes it easier to implement caching for it as well.
- I wouldn't place too much weight on masters management implementation right now because I think no matter what route we take to HA masters, it will be blown up mostly anyway and won't stay how it is right now.

nprokopic: > Can we limit changes to some manageable size here and settle with state that we can take further in next steps, without fully refactoring everything now?

Yes! This was more about the code deduplication, less about some bigger refactoring.

> What do you think if we actually tried to implement the node pools resources in TDD fashion in the sense that we try to create some kind of integration and unit tests for the whole implementation while actually doing it? I figure that would help us in the future a lot and might help to solve some concerns here.

Interesting idea, let's talk about it.

> The client factory idea is something we need to drive. That solves the problems I had with metrics collection, but it also makes it easier to implement caching for it as well.

Agreed. Now we are getting the clients from `controllercontext`, so I'm looking into that, to see if we can decouple Azure clients from `controllercontext`.

client/azure_client_set.go
```diff
@@ -160,7 +160,7 @@ func prepareClient(client *autorest.Client, authorizer autorest.Authorizer, part
 	return client
 }
 
-func newDeploymentsClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (*resources.DeploymentsClient, error) {
+func newDeploymentsClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (interface{}, error) {
```
fiunchinho, line 163: why not use `toDeploymentsClient` inside this function?

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "4.1.0-dev"
+	version            = "4.1.0-refactoring-masters-instance"
```
fiunchinho, line 8: you probably don't want to merge this to master

client/azure_client_set.go
nprokopic: It's so we can have more generic (yet simple) code for Azure client creation in client factory. It was initially Tuomas' suggestion.

For example this is now the creation/getting of `DeploymentsClient`:
```
func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.DeploymentsClient, error) {
	client, err := f.getClient(cr, "DeploymentsClient", newDeploymentsClient)
	if err != nil {
	}

	return toDeploymentsClient(client), nil
}
```

And this the same for `GroupsClient`:
```
func (f *Factory) GetGroupsClient(cr v1alpha1.AzureConfig) (*resources.GroupsClient, error) {
	client, err := f.getClient(cr, "GroupsClient", newGroupsClient)
	if err != nil {
	}

	return toGroupsClient(client), nil
}
```

So in both cases the client is created with generic function that is returning `interface{}`:
```
type clientCreatorFunc func(autorest.Authorizer, string, string) (interface{}, error)

func (f *Factory) getClient(cr v1alpha1.AzureConfig, clientType string, createClient clientCreatorFunc) (interface{}, error) {
// ...
}
```

Without returning `interface{}` in `newDeploymentsClient` (and other similar functions), we would have to do something like this for creating clients in the factory, having a wrapper function to return an `interface{}:
```
func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.DeploymentsClient, error) {
	createClientFunc := func(authorizer autorest.Authorizer, subscriptionID string, partnerID string) (interface{}, error) {
		return newDeploymentsClient(authorizer, subscriptionID, partnerID)
	}
	client, err := f.getClient(cr, "DeploymentsClient", createClientFunc)
	if err != nil {
	}


	return client.(*resources.DeploymentsClient), nil
}
```

So the whole idea was to make new client factory code simpler and cleaner, and let the old code take a bit of the burden :) [Here is a commit with before and after this change](https://github.com/giantswarm/azure-operator/pull/853/commits/1223182eaae7acf8d5a2e7f6d0078ccfc8601230).

pkg/project/project.go
nprokopic: Yup, I'll remove it before merging :)

client/azure_client_set.go
fiunchinho: I understand now. It's better like this.

PR #854, opened 2020-05-26
Update to operatorkit v1.0.0

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/11195


service/controller/cluster.go
```diff
@@ -109,9 +179,435 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 		}
 	}
 
-	c := &Cluster{
+	return &Cluster{
 		Controller: operatorkitController,
+	}, nil
+}
+
+func newClusterResources(config ClusterConfig, certsSearcher certs.Interface) ([]resource.Interface, error) {
```
fiunchinho, line 187: resources are defined in this function instead of using the `resource_set.go` file

service/controller/cluster.go
```diff
@@ -94,13 +121,56 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
...
-			Name:      config.ProjectName,
-			ResourceSets: []*controller.ResourceSet{
-				resourceSet,
-			},
+			Name:      project.Name(),
+			Resources: resources,
 			NewRuntimeObjectFunc: func() runtime.Object {
 				return new(v1alpha1.AzureConfig)
 			},
+			InitCtx: func(ctx context.Context, obj interface{}) (context.Context, error) {
```
xh3b4sd, line 129: Do you have plans to get rid of the controller context usage? Further, this function should not do more than set the controller context so that it is initialized. All other logic should happen somewhere else. [In the `aws-operator` this is not more than 1 line](https://github.com/giantswarm/aws-operator/blob/762d8a8ce09be1d7ea95815f08a0e6c1fb63e070/service/controller/cluster.go#L112-L114).

service/controller/cluster.go
```diff
@@ -94,13 +121,56 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 		c := controller.Config{
```
xh3b4sd, line 121: Since the handler func is gone it is important that you use the label selector. Otherwise the operator will just watch all CRs instead of the ones having the right version label.

service/controller/cluster.go
fiunchinho: Yes, we will tackle that in the near future.

service/controller/cluster.go
fiunchinho: Added selector.

service/controller/cluster.go
```diff
@@ -1,19 +1,53 @@
...
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/instance"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/ipam"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/masters"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/namespace"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/release"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/resourcegroup"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/service"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/tenantclients"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/vpn"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/vpnconnection"
```
xh3b4sd, line 50: I noticed you version the operator as a package that way itself. Why is that? Do you have any reason or benefits with that? What is your use case?

fiunchinho: This is [how go modules work](https://blog.golang.org/v2-go-modules) AFAIK. We could have (if needed) different versions in different folders at the same time.

xh3b4sd: Sure, we use that in our libraries. But since the operator is not imported anywhere I was wondering why we should do that or care.

tuommaki: My understanding is also that go module version should always follow its major version and since that reflects in imports as well I find it personally more trustworthy to follow the practice than wait for a surprise when the tooling does something unpredictable because we diverged from the general rule.

PR #856, opened 2020-05-28
Resource that creates nodepool vmss

fiunchinho:


service/controller/resource/nodes/instances.go
```diff
@@ -43,3 +44,70 @@ func (r *Resource) AllInstances(ctx context.Context, customObject providerv1alph
 
 	return instances, nil
 }
+
+func (r *Resource) AllWorkerInstances(ctx context.Context, virtualMachineScaleSetVMsClient *compute.VirtualMachineScaleSetVMsClient, resourceGroupName, vmssName string) ([]compute.VirtualMachineScaleSetVM, error) {
```
fiunchinho, line 48: I placed together in this file operations that we do with Azure VMSS API.

service/controller/resource/ipam/azure_machinepool_persister.go
```diff
@@ -0,0 +1,117 @@
...
+
+	p := &AzureMachinePoolPersister{
+		ctrlClient: config.CtrlClient,
+		logger:     config.Logger,
+	}
+
+	return p, nil
+}
+
+func (p *AzureMachinePoolPersister) Persist(ctx context.Context, vnet net.IPNet, obj interface{}) error {
```
fiunchinho, line 45: We save the allocated subnet:
- as annotation in the `AzureMachinPool`.
- in the `AzureCluster.Spec.NetworkSpec.Subnets` . This will make the `AzureCluster` controller to actually create the subnet.

service/controller/resource/ipam/azure_config_persister.go
```diff
@@ -4,42 +4,42 @@ import (
...
+		ctrlClient: config.CtrlClient,
+		logger:     config.Logger,
 	}
 
 	return p, nil
 }
 
-func (p *AzureConfigPersister) Persist(ctx context.Context, vnet net.IPNet, namespace string, name string) error {
-	cr, err := p.g8sClient.ProviderV1alpha1().AzureConfigs(namespace).Get(name, metav1.GetOptions{})
+func (p *AzureConfigPersister) Persist(ctx context.Context, vnet net.IPNet, obj interface{}) error {
```
fiunchinho, line 41: Changed the code so it uses the controller client.

service/controller/resource/instance/status.go
```diff
@@ -1,86 +1,67 @@
...
-
-		// At this point we ensure resource statuses of other resources are
-		// preserved as they are. When we want to remove a status of resource A, but
-		// find a status of resource B, we keep B, because we were not asked to
-		// remove it.
-		if !unversionedNameMatches(r.Name, n) {
-			allResources = append(allResources, r)
-			continue
-		}
+func (r *Resource) saveCurrentState(ctx context.Context, customObject v1alpha3.AzureMachinePool, state string) error {
```
fiunchinho, line 27: The state machine current state is saved in an annotation inside `AzureMachinePool`.

service/controller/resource/instance/deployment.go
```diff
@@ -2,119 +2,254 @@ package instance
...
+		},
+	}
 
-	storageAccountsClient, err := r.ClientFactory.GetStorageAccountsClient(key.CredentialNamespace(obj), key.CredentialName(obj))
-	if err != nil {
-		return azureresource.Deployment{}, microerror.Mask(err)
+	return d, nil
+}
+
+func (r Resource) getSubnetID(azureMachinePool *capzexpv1alpha3.AzureMachinePool, azureCluster *capzv1alpha3.AzureCluster) (string, error) {
```
fiunchinho, line 102: We select the subnet which was allocated by the `ipam` handler.

service/controller/resource/instance/create_scale_workers.go
```diff
@@ -4,126 +4,120 @@ import (
...
+	currentWorkerCount, err := r.GetInstancesCount(ctx, virtualMachineScaleSetsClient, key.ClusterID(&azureMachinePool), key.NodePoolVMSSName(&azureMachinePool))
+	if err != nil {
+		return DeploymentUninitialized, microerror.Mask(err)
+	}
+	r.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("The current number of workers is: %d", currentWorkerCount))
+
 	if desiredWorkerCount > currentWorkerCount {
-		// Raise the worker count by one
-		err = r.scaleVMSS(ctx, cr, key.WorkerVMSSName, currentWorkerCount+1)
+		err = r.ScaleVMSS(ctx, virtualMachineScaleSetsClient, key.ClusterID(&azureMachinePool), key.NodePoolVMSSName(&azureMachinePool), int32(desiredWorkerCount))
```
fiunchinho, line 69: let's try scaling straight to the desired number of workers, instead of going one by one.

service/controller/resource/instance/create.go
```diff
@@ -66,14 +62,19 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
+	err = r.CtrlClient.Status().Update(ctx, &azureMachinePool)
```
fiunchinho, line 65: We are saving the `Status` field of the CR. This is to be able to save the provisioning state.

service/controller/resource/nodepool/deployment.go
```diff
@@ -0,0 +1,251 @@
...
+	keys, err := storageAccountsClient.ListKeys(ctx, resourceGroupName, storageAccountName, "")
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	if len(*(keys.Keys)) == 0 {
+		return "", microerror.Maskf(executionFailedError, "storage account key's list is empty")
+	}
+	primaryKey := *(((*keys.Keys)[0]).Value)
+	containerName := key.BlobContainerName()
```
fiunchinho, line 142: Use new `key` function for the blob name, since we need a different blob for every node pool.

service/controller/resource/ipam/azure_machinepool_checker.go
```diff
@@ -0,0 +1,80 @@
...
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type AzureMachinePoolCheckerConfig struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+}
+
+type AzureMachinePoolChecker struct {
```
nprokopic, line 20: We could call this one `AzureMachinePoolSubnetChecker`, so it's more explicit and obvious what it is about. Also analogously `AzureMachinePoolPersister` -> `AzureMachinePoolSubnetPersister`. Unless there are some other use cases and these will be checking and persisting other things besides subnet in the future.

WDYT @tuommaki?

service/controller/resource/ipam/azure_machinepool_checker.go
```diff
@@ -0,0 +1,80 @@
...
+
+	azureCluster, err := c.getAzureClusterFromCluster(ctx, cluster)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	// In case there is no subnet tracked so far, we want to proceed with the allocation process.
+	for _, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
+		if subnet.Name == azureMachinePool.Name {
+			return false, nil
```
nprokopic, line 61: I would expect `Check` method to return `true` when the subnet is allocated, this seems a bit reversed. WDYT @tuommaki?

nprokopic: Hmm now when I look at `Checker` interface, it says:
> Checker determines whether a subnet has to be allocated.

nprokopic: I would like us to adjust the naming a bit, so that the purpose of interfaces, structs and functions is a bit more clear.

service/controller/resource/ipam/azure_machinepool_checker.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	return true, nil
+}
+
+func (c *AzureMachinePoolChecker) getAzureClusterFromCluster(ctx context.Context, cluster *capiv1alpha3.Cluster) (*capzv1alpha3.AzureCluster, error) {
```
nprokopic, line 68: We have this function repeated in 4 places in the project, I'll place it in some common package so it's reused.

service/controller/resource/ipam/azure_machinepool_checker.go
tuommaki: This is currently following the structure of `aws-operator`. If we change the structure of this, I'd suggest considering the problem from the beginning. For me this feels a bit complicated considering what it actually does. Important thing here (in IPAM) is that necessary operations are executed within acquired lock.

service/controller/resource/ipam/azure_machinepool_checker.go
tuommaki: I think there was something towards the owner reference functionality. Maybe check the spark PR. If not, maybe all of those could be refactored to same?

nprokopic: Thanks, I'll check Spark CR then first.

nprokopic: There is a function in Spark CR for getting a owner `Cluster`, but here we need to get `AzureCluster` for the owner `Cluster`, so the other way around.

nprokopic:
> I didn't review this fully on this go, but just as an information: The node pool subnet should be allocated from the tenant cluster network (it might already do so, but just saying to be explicit).

It's very obvious to me that it's working this way, so I started digging. Still in the rabbit hole...

nprokopic:
I believe that subnet allocation for node pools is not correct.

It's using existing ipam handler which allocates a subnet from the installation range https://github.com/giantswarm/azure-operator/blob/7f7676d0bb5036eccae176ec11b3de5cf4dcfb47/service/controller/resource/ipam/create.go#L78. It should be using VNet range allocated for the tenant cluster.

It's also using existing [subnet collector](https://github.com/giantswarm/azure-operator/blob/master/service/controller/resource/ipam/subnet_collector.go), which is getting all network ranges for all clusters in an installation and counting them as allocated. We should only take into consideration subnets from the tenant cluster VNet where we are adding new node pool.

I'll work on a fix and submit a separate pull request into this one, so it's easier to review.

service/controller/key/key.go
```diff
@@ -346,11 +341,6 @@ func MastersSubnetCIDR(customObject providerv1alpha1.AzureConfig) string {
 	return customObject.Spec.Azure.VirtualNetwork.MasterSubnetCIDR
 }
 
-// WorkerCount returns the desired number of workers.
```
nprokopic, line 349: Not sure why this was removed. It broke `instance` handler, so I'll revert it for now.

service/controller/key/key.go
```diff
@@ -23,7 +23,6 @@ const (
 	installationTagName       = "GiantSwarmInstallation"
 	organizationTagName       = "GiantSwarmOrganization"
 	MastersVmssDeploymentName = "masters-vmss-template"
-	WorkersVmssDeploymentName = "workers-vmss-template"
```
nprokopic, line 26: Reverting this for now, as it broke `instance` handler.

service/controller/key/key.go
tuommaki: These might have been cleanup work towards dropped in-built workers handling?

fiunchinho:
Closing this. Let's revisit it if we need it further down the road.

tuommaki:
wut?

service/controller/resource/nodepool/create.go
```diff
@@ -0,0 +1,94 @@
...
+		reconciliationcanceledcontext.SetCanceled(ctx)
+	} else {
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "no state change")
+	}
+
+	return nil
+}
+
+func isMasterUpgrading(getter key.AnnotationsGetter) bool {
+	masterUpgrading, exists := getter.GetAnnotations()[annotation.IsMasterUpgrading]
```
tuommaki, line 88: I think we need to revisit this in the near future. Having annotation of master[s] state in every `AzureMachinePool` CR is not ideal. I'd expect to find the state of masters in `AzureMachine[s]` of master nodes or alternatively in `AzureCluster` (maybe better since that would be single common location).

Maybe create a separate issue to revisit this, add a comment here and continue with this one?

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -0,0 +1,221 @@
...
+
+	virtualMachineScaleSetVMsClient, err := r.ClientFactory.GetVirtualMachineScaleSetVMsClient(credentialSecret.Namespace, credentialSecret.Name)
+	if err != nil {
+		return currentState, microerror.Mask(err)
+	}
+
+	desiredDeployment, err := r.getDesiredDeployment(ctx, storageAccountsClient, release, azureCluster, machinePool, &azureMachinePool)
+	if IsNotFound(err) {
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "Azure resource not found, canceling resource")
+		return currentState, nil
```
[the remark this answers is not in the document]
tuommaki: Expected errors should be handled gracefully. We should have separate metrics for node pool states so we can create alerts in case node pool gets stuck in creation. If something goes wrong earlier in the chain (by bug or by Azure) then we get paged for that.

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -0,0 +1,221 @@
...
+
+		if nodesNeedToBeRolled {
+			return ScaleUpWorkerVMSS, nil
+		}
+
+		return currentState, nil
+	}
+
+	// Start watcher on the instances to avoid stuck VMs to block the deployment progress forever
+	r.InstanceWatchdog.DeleteFailedVMSS(ctx, virtualMachineScaleSetVMsClient, key.ClusterID(&azureMachinePool), key.NodePoolVMSSName(&azureMachinePool))
```
tuommaki, line 110: We need to monitor this a bit. I'm not fully sure how safe this is with node pools (since there can be hundreds of VMSSs per installation).

service/controller/resource/nodepool/template/main.json
```diff
@@ -0,0 +1,382 @@
...
+      "name": "[variables('vmssName')]",
+      "location": "[resourceGroup().location]",
+      "zones": "[if(greater(length(parameters('zones')),0), parameters('zones'), json('null'))]",
+      "tags": {
+        "provider": "[toUpper(parameters('GiantSwarmTags').provider)]",
+        "cluster-autoscaler-enabled": "true",
+        "cluster-autoscaler-name": "[variables('vmssName')]",
+        "gs-azure-operator.giantswarm.io-version": "[parameters('azureOperatorVersion')]",
+        "capi-azure-machine-pool.giantswarm.io-version": "[parameters('azureMachinePoolVersion')]",
+        "capi-machine-pool.giantswarm.io-version": "[parameters('machinePoolVersion')]"
```
tuommaki, line 267: :heart:

service/controller/resource/subnet/resource.go
```diff
@@ -216,7 +216,10 @@ func (r *Resource) ensureSubnets(ctx context.Context, deploymentsClient *azurere
 // This is required because when removing a node pool, we remove the subnet from `AzureCluster`, so we can remove it here from Azure.
 func (r *Resource) garbageCollectSubnets(ctx context.Context, deploymentsClient *azureresource.DeploymentsClient, subnetsClient *network.SubnetsClient, azureCluster capzv1alpha3.AzureCluster) error {
 	subnetsIterator, err := subnetsClient.ListComplete(ctx, key.ClusterID(&azureCluster), azureCluster.Spec.NetworkSpec.Vnet.Name)
-	if err != nil {
+	if IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Vnet not %#q found, cancelling resource", azureCluster.Spec.NetworkSpec.Vnet.Name))
```
tuommaki, line 220: Formatting directive in wrong position?

tuommaki (commented):
Skimmed through. Mostly looks good. Few things noted. In the future in similar situation let's try to make these PRs smaller by first copying the base (i.e. `instance` -> `nodepool`) first untouched and then renaming on second go and on third phase the actual changes. That way it's easier to identify what changed and the PR size doesn't grow so big. :slightly_smiling_face:

PR #857, opened 2020-05-28
Azure clients factory

nprokopic:
What's in the box:
- Azure clients factory for creating:
  - `DeploymentsClient`
  - `GroupsClient`
  - `VirtualMachineScaleSetsClient`
  - `VirtualMachineScaleSetVMsClient`
  - `StorageAccountsClient`
- `masters` and `instance` resources are now using the factory to create clients instead of using `controllercontext`

Out of scope of this pull request:
- Using Azure clients factory in all resources (as this pull request is towards refactoring of `masters` and `instance` resources)

Still WIP, so TBA:
- Removal of cached clients when the tenant cluster is deleted.

### DIFF
--- a/client/error.go
+++ b/client/error.go
@@ -1,0 +1,9 @@
+package client
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
+var invalidConfigError = &microerror.Error{
+	Kind: "invalidConfigError",
+}
--- a/client/factory.go
+++ b/client/factory.go
@@ -20,6 +20,8 @@ type FactoryConfig struct {
 	GSTenantID string
 }
 
+// Factory is creating Azure clients for specified AzureConfig CRs, so basically for specified
+// tenant clusters. All created clients are cached.
 type Factory struct {
 	mutex      sync.Mutex
 	k8sClient  k8sclient.Interface
@@ -30,7 +32,8 @@ type Factory struct {
 
 type clientCreatorFunc func(autorest.Authorizer, string, string) (interface{}, error)
 
-// call NewFactory in EnsureCreated, so it's used in only a single reconciliation loop
+// NewFactory returns a new Azure client factory that is used throughout entire azure-operator
+// lifetime.
 func NewFactory(config FactoryConfig) (*Factory, error) {
 	if config.K8sClient == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.K8sClient must not be empty", config)
@@ -49,7 +52,8 @@ func NewFactory(config FactoryConfig) (*Factory, error) {
 }
 
 // GetDeploymentsClient returns DeploymentsClient that is used for management of deployments and
-// ARM templates.
+// ARM templates. The client (for specified cluster) is cached after creation, so the same client
+// is returned every time.
 func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.DeploymentsClient, error) {
 	clientSetKey := key.CredentialName(cr)
 
@@ -71,7 +75,8 @@ func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.Depl
 	return f.clients[clientSetKey].DeploymentsClient, nil
 }
 
-// GetGroupsClient returns GroupsClient that is used for management of resource groups.
+// GetGroupsClient returns GroupsClient that is used for management of resource groups. The client
+// (for specified cluster) is cached after creation, so the same client is returned every time.
 func (f *Factory) GetGroupsClient(cr v1alpha1.AzureConfig) (*resources.GroupsClient, error) {
 	clientSetKey := key.CredentialName(cr)
 
@@ -94,7 +99,8 @@ func (f *Factory) GetGroupsClient(cr v1alpha1.AzureConfig) (*resources.GroupsCli
 }
 
 // GetVirtualMachineScaleSetsClient returns VirtualMachineScaleSetsClient that is used for
-// management of virtual machine scale sets.
+// management of virtual machine scale sets. The client (for specified cluster) is cached after
+// creation, so the same client is returned every time.
 func (f *Factory) GetVirtualMachineScaleSetsClient(cr v1alpha1.AzureConfig) (*compute.VirtualMachineScaleSetsClient, error) {
 	clientSetKey := key.CredentialName(cr)
 
@@ -117,7 +123,8 @@ func (f *Factory) GetVirtualMachineScaleSetsClient(cr v1alpha1.AzureConfig) (*co
 }
 
 // GetVirtualMachineScaleSetVMsClient returns GetVirtualMachineScaleSetVMsClient that is used for
-// management of virtual machine scale set instances.
+// management of virtual machine scale set instances. The client (for specified cluster) is cached
+// after creation, so the same client is returned every time.
 func (f *Factory) GetVirtualMachineScaleSetVMsClient(cr v1alpha1.AzureConfig) (*compute.VirtualMachineScaleSetVMsClient, error) {
 	clientSetKey := key.CredentialName(cr)
 
@@ -126,13 +133,13 @@ func (f *Factory) GetVirtualMachineScaleSetVMsClient(cr v1alpha1.AzureConfig) (*
 		f.clients[clientSetKey] = &AzureClientSet{}
 	}
 	if f.clients[clientSetKey].VirtualMachineScaleSetVMsClient == nil {
-		vmssClient, err := f.createClient(cr, func(authorizer autorest.Authorizer, subscriptionID string, partnerID string) (interface{}, error) {
+		vmssVMsClient, err := f.createClient(cr, func(authorizer autorest.Authorizer, subscriptionID string, partnerID string) (interface{}, error) {
 			return newVirtualMachineScaleSetVMsClient(authorizer, subscriptionID, partnerID)
 		})
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
-		f.clients[clientSetKey].VirtualMachineScaleSetVMsClient = vmssClient.(*compute.VirtualMachineScaleSetVMsClient)
+		f.clients[clientSetKey].VirtualMachineScaleSetVMsClient = vmssVMsClient.(*compute.VirtualMachineScaleSetVMsClient)
 	}
 	f.mutex.Unlock()
 
--- a/client/factory.go
+++ b/client/factory.go
@@ -1,0 +1,181 @@
+package client
+
+import (
+	"sync"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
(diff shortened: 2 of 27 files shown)


service/controller/resource/masters/create_check_flatcar_migration_needed.go
```diff
@@ -17,7 +18,7 @@ func (r *Resource) checkFlatcarMigrationNeededTransition(ctx context.Context, ob
 		return Empty, microerror.Mask(err)
 	}
 
-	legacyExists, err := r.vmssExistsAndHasActiveInstance(ctx, key.ResourceGroupName(cr), key.LegacyMasterVMSSName(cr))
+	legacyExists, err := r.vmssExistsAndHasActiveInstance(ctx, cr, key.ResourceGroupName(cr), key.LegacyMasterVMSSName(cr))
```
nprokopic, line 21: Here and in some other similar cases, the last two parameters can now be removed (and calculated inside of the function), just IMO it's not worth it, as these helper functions are mostly added due to migration to Flatcar, so I would not touch those, as that will be cleaned up once all clusters have migrated.

When we clean up Flatcar migration code, this will get cleaned up as well.

client/factory.go
```diff
@@ -0,0 +1,188 @@
...
+		f.clients[clientSetKey] = &AzureClientSet{}
+	}
+	if f.clients[clientSetKey].DeploymentsClient == nil {
+		deploymentClient, err := f.createClient(cr, func(authorizer autorest.Authorizer, subscriptionID string, partnerID string) (interface{}, error) {
+			return newDeploymentsClient(authorizer, subscriptionID, partnerID)
+		})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		f.clients[clientSetKey].DeploymentsClient = deploymentClient.(*resources.DeploymentsClient)
```
nprokopic, line 71: Creating clients in a generic manner by returning `interface{}` from `createClient(...)` and casting back to actual client type, so there is no too much code repetition.

Performance overhead should not be a concern, as this is done only once for every client, over the lifetime of the `azure-operator`.

client/factory.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return f.clients[clientSetKey].DeploymentsClient, nil
+}
+
+// GetGroupsClient returns GroupsClient that is used for management of resource groups. The client
+// (for specified cluster) is cached after creation, so the same client is returned every time.
+func (f *Factory) GetGroupsClient(cr v1alpha1.AzureConfig) (*resources.GroupsClient, error) {
+	clientSetKey := key.CredentialName(cr)
+
+	f.mutex.Lock()
```
nprokopic, line 83: Any opinions around locking? AFAIK we are not reconciling concurrently, but we do spin up go routines in some places in code, so it's not unimaginable that some of those end up using `client.Factory`.

client/factory.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Factory is creating Azure clients for specified AzureConfig CRs, so basically for specified
+// tenant clusters. All created clients are cached.
+type Factory struct {
+	mutex      sync.Mutex
+	k8sClient  k8sclient.Interface
+	gsTenantID string
+
+	clients map[string]*AzureClientSet
```
nprokopic, line 30: Map of all clients in format `org's credential id` => `AzureClientSet`.

client/factory.go
nprokopic: P.S. I miss generics :)

client/factory.go
fiunchinho: We can create a new type in the package like `type credentialID string` so the map is more explicit `clients map[credentialID]*AzureClientSet`

client/factory.go
fiunchinho: I think it doesn't harm to have them :+1:

service/controller/resource/masters/create_deallocate_legacy_instance.go
```diff
@@ -118,8 +119,8 @@ func (r *Resource) getRunningInstances(ctx context.Context, resourceGroup string
 	return instancesRunning, nil
 }
 
-func (r *Resource) getVMSS(ctx context.Context, resourceGroup string, vmssName string) (*compute.VirtualMachineScaleSet, error) {
-	c, err := r.GetScaleSetsClient(ctx)
+func (r *Resource) getVMSS(ctx context.Context, customObject providerv1alpha1.AzureConfig, resourceGroup string, vmssName string) (*compute.VirtualMachineScaleSet, error) {
```
fiunchinho, line 122: Not sure if it's better for these functions to receive the `AzureConfig` or the client so they don't have to create it. What do you think? It feels like passing the client we make explicit that they just need the client, not the `AzureConfig`.

nprokopic: This part of the code is part of Flatcar migration, didn't really want to get into refactoring it, as it will get thrown out anyway.

service/controller/resource/instance/delete.go
```diff
@@ -2,8 +2,17 @@ package instance
...
+
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
 func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.ClientFactory.RemoveAllClients(cr)
```
fiunchinho, line 16: why remove the clients when removing a cluster?

fiunchinho: I mean, maybe two clusters are using the same credential?

tuommaki: Wouldn't this also be wrong point in reconciliation loop? The clients are needed later in the reconciliation flow and would get created again while possibly still allow finalizer get dropped in the last one so this wouldn't get called again.

Could the clients be cached for some time and then get cleaned up when unused for 30min or so?

nprokopic: > I mean, maybe two clusters are using the same credential?

Yup, true. I'm adding cluster tracking, so it's removed only when the last one is removed.

> The clients are needed later in the reconciliation flow and would get created again while possibly still allow finalizer get dropped in the last one so this wouldn't get called again.

Not sure I understand this. You're saying clients are used after `EnsureDeleted`? In that case, yeah those wouldn't get deleted.

> Could the clients be cached for some time and then get cleaned up when unused for 30min or so?

We can do something like this.

tuommaki: > Not sure I understand this. You're saying clients are used after EnsureDeleted? In that case, yeah those wouldn't get deleted.

Yeah. Concrete example:

- These are resources executed over single reconciliation period: https://github.com/giantswarm/azure-operator/blob/master/service/controller/cluster.go#L566-L584
- After `instance` resource there are still:
  - `endpointsResource`
  - `vpnResource`
  - `vpnconnectionResource`
- At least `vpnconnectionResource` needs client, although currently it's not using the one from factory, but maybe it should?

nprokopic: > At least vpnconnectionResource needs client, although currently it's not using the one from factory, but maybe it should?

IMO all resources should get clients from the factory. Maybe just not all of them in this pull request :)

nprokopic: The factory is now using go-cache for storing clients, and the clients are cached for 30 minutes.

nprokopic: More precisely, the cache duration is refreshed on every access. So if some client is not used for 30 minutes, it's removed from cache.

client/error.go
```diff
@@ -0,0 +1,9 @@
+package client
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
+var invalidConfigError = &microerror.Error{
```
tuommaki, line 7: Matcher `IsInvalidConfig(err error)` missing on purpose?

go.mod
```diff
@@ -48,6 +48,7 @@ require (
 	github.com/markbates/pkger v0.15.1
 	github.com/onsi/ginkgo v1.12.0 // indirect
 	github.com/onsi/gomega v1.9.0 // indirect
+	github.com/patrickmn/go-cache v2.1.0+incompatible
```
tuommaki, line 51: Hmm.. No changes to `go.sum`? Did you run `go mod tidy`?

client/factory.go
```diff
@@ -0,0 +1,223 @@
...
+	return factory, nil
+}
+
+// GetDeploymentsClient returns DeploymentsClient that is used for management of deployments and
+// ARM templates. The client (for specified cluster) is cached after creation, so the same client
+// is returned every time.
+func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.DeploymentsClient, error) {
+	createClientFunc := func(authorizer autorest.Authorizer, subscriptionID string, partnerID string) (interface{}, error) {
+		return newDeploymentsClient(authorizer, subscriptionID, partnerID)
+	}
```
tuommaki, line 86: Is this wrapping needed only because the return type differs between those two functions? i.e. `createClientFunc` returns `(interface{}, error)` whereas `newDeploymentsClient()` returns `(*resources.DeploymentsClient, error)`?

If so, would it make sense if you change the signature of `new*Client()` functions to return `interface{}` and introduce casting functions for each type? The casting function could be then also used together with `getClient()` since that also returns `interface{}`.

After that change this function could roughly look like this:
```go
client, err := f.getClient(cr, "DeploymentsClient", newDeploymentsClient)
if err != nil {
}

return toDeploymentsClient(client), err
```

If there is still old code using `newDeploymentsClient` as it used to be, you could wrap those with `toDeploymentsClient()`. It would maybe look a bit funny, but in this case the benefit would be on the new code and old maintained code is the one that suffers slight readability hit. WDYT?

client/error.go
nprokopic: Yup, I was adding stuff as they were needed.

But now when you mention it, it looks like I should add it and use `nodes.IsInvalidConfig(err error)` instread of `masters.IsInvalidConfig(err error)` and `instance.IsInvalidConfig(err error)` in few places.

P.S. I'm not sure I like this repetition of `error.go` and practically same errors/functions in it all over the project, it seems to be error prone.

go.mod
nprokopic: yeah, no changes:

```
nikola@avokado ~/src/giantswarm/azure-operator 2020-06-10 10:34:17
$ go mod tidy
nikola@avokado ~/src/giantswarm/azure-operator 2020-06-10 10:34:18
$ git status
On branch azure-clients-factory
Your branch is up to date with 'origin/azure-clients-factory'.

nothing to commit, working tree clean
```

client/factory.go
nprokopic: Ok doing this now, as I have to fix error handling anyway :)

client/error.go
nprokopic: > But now when you mention it, it looks like I should add it and use nodes.IsInvalidConfig(err error) instead of masters.IsInvalidConfig(err error) and instance.IsInvalidConfig(err error) in few places.

Or not :thinking:, ok I have to check first how is error comparison working in `IsInvalidConfig(err error)`.

tuommaki: Since the specific instance of error type differs from package to another, the matcher must be accompanied with the specific types in every package.

nprokopic: So currently there is no need anywhere for `client.IsInvalidConfig(err)`, the factory is created in `resource_set.go` (it will be changed after rebasing on `master`), and the error from `client.NewFactory(config)` is then just propagated all the way back to `main.go` where it panics.

nprokopic: But this should be corrected in other places in the other pull request for refactoring `masters` and `instance`.

PR #858, opened 2020-06-03
Restrict storage account access from CP and master/workers subnets only

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/9464


service/controller/key/key.go
```diff
@@ -129,7 +129,7 @@ func BlobName(customObject providerv1alpha1.AzureConfig, role string) string {
 }
 
 func CalicoCIDR(customObject providerv1alpha1.AzureConfig) string {
-	return customObject.Spec.Cluster.Calico.Subnet
+	return customObject.Spec.Azure.VirtualNetwork.CalicoSubnetCIDR
```
whites11, line 132: This is a bug from the IPAM feature I guess.

service/controller/resource/containerurl/error.go
```diff
@@ -32,7 +32,8 @@ func IsStorageAccountNotFound(err error) bool {
 	if err == nil {
 		return false
 	}
-	return strings.Contains(microerror.Cause(err).Error(), "ResourceNotFound")
+	return strings.Contains(microerror.Cause(err).Error(), "ResourceNotFound") ||
```
whites11, line 35: Apparently the Azure response for the 404 case changed from a generic one to a specific one recently.

service/controller/resource/deployment/publiips.go
```diff
@@ -0,0 +1,27 @@
+package deployment
```
fiunchinho, line 1: The file name has a typo

service/controller/resource/deployment/publiips.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+)
+
+func (r Resource) getCPPublicIPAddresses(ctx context.Context) ([]string, error) {
```
fiunchinho, line 10: When are these public IP's created? Could they be passed as parameters to the azure-operator instead of fetching them? or do they change? That way I could pass whatever I want when running locally.

service/controller/cluster.go
```diff
@@ -569,10 +571,10 @@ func newClusterResources(config ClusterConfig, certsSearcher certs.Interface) ([
 		namespaceResource,
 		serviceResource,
 		resourceGroupResource,
-		containerURLResource,
 		encryptionkeyResource,
-		blobObjectResource,
 		deploymentResource,
+		containerURLResource,
+		blobObjectResource,
```
tuommaki, line 577: What affects this resource ordering? Changes in `deploymentResource` seemed like contained. Is this needed?

service/controller/resource/deployment/publiips.go
whites11: they're created (today) by terraform at control plane deployment.
I don't think they should be passed. That would mean hardcode them in the `installations` repo.

service/controller/cluster.go
whites11: yes, because if the public IP address of the control plane changes, the containerURLResource and blobObjectResource will fail and stop the reconciliation.
The deploymentResource has to run before them to update the storage account to the newest IP addresses otherwise we get a crash loop.

service/controller/resource/deployment/publiips.go
whites11: thanks, fixed

service/controller/cluster.go
tuommaki: Ah, right, and it's the `instanceResource` then what further uses the blob objects uploaded by `blobObjectResource`?

whites11: yes, together with the `masterResource`

tuommaki: Coolio!

PR #859, opened 2020-06-03
Use single tenant service principal if secret is labeled

fiunchinho:
When migrating to multi tenant service principals, the code will fail to manage tenant cluster resources unless the Service Principal passed as parameter has been invited to customer's Active Directory, and assigned the `Owner` IAM role.

This PR changes the code so that if the credential secret has a label/annotation, all fields (clientid, clientsecret, tenantid...) are used to authenticate, to make it work like before. When service principals migration is over we can remove the label/annotation.

### DIFF
--- a/pkg/credential/credential.go
+++ b/pkg/credential/credential.go
@@ -18,6 +18,7 @@ const (
 	clientSecretKey   = "azure.azureoperator.clientsecret"
 	defaultAzureGUID  = "37f13270-5c7a-56ff-9211-8426baaeaabd"
 	partnerIDKey      = "azure.azureoperator.partnerid"
+	singleTenantLabel = "giantswarm.io/single-tenant-service-principal"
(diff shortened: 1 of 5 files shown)


pkg/credential/credential.go
```diff
@@ -18,16 +18,17 @@ const (
 	clientSecretKey   = "azure.azureoperator.clientsecret"
 	defaultAzureGUID  = "37f13270-5c7a-56ff-9211-8426baaeaabd"
 	partnerIDKey      = "azure.azureoperator.partnerid"
+	singleTenantLabel = "giantswarm.io/single-tenant-service-principal"
```
tuommaki, line 21: Minor nit: Maybe move this to `label` package?

fiunchinho: Done

PR #860, opened 2020-06-11
Add CP subnet to the allowed subnets for storage account in TC

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/11499

This PR adds the CP's subnet to the allowed subnets to acces the storage account in TCs


service/controller/resource/deployment/deployment.go
```diff
@@ -71,6 +70,15 @@ func (r Resource) newDeployment(ctx context.Context, customObject providerv1alph
 	return d, nil
 }
 
+func (r *Resource) getControlPlaneSubscriptionID(ctx context.Context) (string, error) {
```
fiunchinho, line 73: Instead of fetching it from the controllercontext`, I think it'd be better to pass the subscriptionID as parameter. We have it already. Until we migrate to multi-tenant service principals, the logic would be
```
cpSubscriptionID := config.Viper.GetString(config.Flag.Service.Azure.HostCluster.Tenant.SubscriptionID)
if cpSubscriptionID == "" {
	cpSubscriptionID = config.Viper.GetString(config.Flag.Service.Azure.SubscriptionID)
}
```

basically, use the _new_ parameter `config.Flag.Service.Azure.HostCluster.Tenant.SubscriptionID` when it's set, otherwise use the old one `config.Flag.Service.Azure.SubscriptionID` , which is effectively the control plane subscription ID.

whites11: What about taking it from the Azure Client Set?
```
var deploymentResource resource.Interface
	{
		c := deployment.Config{
			Debugger:         newDebugger,
			G8sClient:        config.K8sClient.G8sClient(),
			InstallationName: config.InstallationName,
			Logger:           config.Logger,

			Azure:                      config.Azure,
			ControlPlaneSubscriptionID: config.CPAzureClientSet.SubscriptionID,
		}

		deploymentResource, err = deployment.New(c)
		if err != nil {
		}
	}
```

fiunchinho: I didn't know we had that! Then just use that.

whites11: Done

nprokopic: Azure Client Set might go away soon (not in the current pull request), in favour of Azure client factory. But I can update this then, so not a blocker.

whites11:
Tested on BYOC/multiaccount (different tenant) as well.

service/controller/resource/deployment/deployment.go
nprokopic: One more reason for not using `controllercontext` is that we should stop using it generally everywhere.

PR #862, opened 2020-06-12
Add feature flag to enable MachinePool

fiunchinho:
Current `master` fails if ClusterAPI CRD's are not present. We can revert back the `AzureMachinePool` controller, or disable the code that requires the CRDs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -148,6 +148,7 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Debug.LogsToken, "", "Enable services which help debugging ignition.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Path, "/opt/ignition", "Default path for the ignition base directory.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.SSH.SSOPublicKey, "", "Public key for trusted SSO CA.")
+	daemonCommand.PersistentFlags().String(f.Service.FeatureGates, "", "Feature gates.")
 
 	return newCommand.CobraCommand().Execute()
 }
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -75,11 +75,7 @@ type ClusterConfig struct {
 	VMSSCheckWorkers int
 }
 
-type Cluster struct {
-	*controller.Controller
-}
-
-func NewCluster(config ClusterConfig) (*Cluster, error) {
+func NewCluster(config ClusterConfig) (*controller.Controller, error) {
 	var err error
 
 	var certsSearcher *certs.Searcher
@@ -184,9 +180,7 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 		}
 	}
 
-	return &Cluster{
-		Controller: operatorkitController,
-	}, nil
+	return operatorkitController, nil
 }
 
 func newClusterResources(config ClusterConfig, certsSearcher certs.Interface) ([]resource.Interface, error) {
--- a/service/service.go
+++ b/service/service.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/Azure/go-autorest/autorest/azure/auth"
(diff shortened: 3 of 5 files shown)


service/controller/cluster.go
```diff
@@ -79,7 +79,7 @@ type Cluster struct {
 	*controller.Controller
 }
 
-func NewCluster(config ClusterConfig) (*Cluster, error) {
+func NewCluster(config ClusterConfig) (*controller.Controller, error) {
```
fiunchinho, line 82: What do you think? Any downsides?

service/service.go
```diff
@@ -306,8 +335,7 @@ func New(config Config) (*Service, error) {
 	s := &Service{
 		Version:                 versionService,
 		bootOnce:                sync.Once{},
-		clusterController:       clusterController,
-		machinePoolController:   machinePoolController,
+		controllers:             controllers,
```
fiunchinho, line 338: Having an slice of controllers allow us to add controllers _on demand_ depending on the enabled feature flags.

service/service.go
```diff
@@ -78,6 +80,22 @@ func New(config Config) (*Service, error) {
 
 	var err error
 
+	features := make(map[string]bool)
```
fiunchinho, line 83: Rudimentary mechanism for feature flags. It would allow us to have MachinePools, but disable it if we need to make a new release.

service/service.go
```diff
@@ -78,6 +80,22 @@ func New(config Config) (*Service, error) {
...
+	{
+		// FeatureGates contains a comma separated list like "MachinePool=true,AnotherFeature=false"
```
axbarsan, line 85: I believe using a separate parameter for each feature would be easier to maintain. (so we would access each one by using `config.Flag.Service.FeatureGates.MachinePool`, for example)

What do you think?

service/service.go
nprokopic: Sounds useful!

service/service.go
fiunchinho: No strong feelings about it. It may be simpler that way. I did it like this because it's how k8s does it.

service/controller/cluster.go
nprokopic: Is `Cluster` struct even necessary if we go down this path?

fiunchinho: Yes, I forgot about it!

fiunchinho: Removed.

fiunchinho:
This is related to this @giantswarm/sig-architecture issue https://github.com/giantswarm/giantswarm/issues/9817

main.go
```diff
@@ -148,6 +148,7 @@ func mainError() error {
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Debug.LogsToken, "", "Enable services which help debugging ignition.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Path, "/opt/ignition", "Default path for the ignition base directory.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.SSH.SSOPublicKey, "", "Public key for trusted SSO CA.")
+	daemonCommand.PersistentFlags().String(f.Service.FeatureGates, "", "Feature gates.")
```
kopiczko, line 151:
```suggestion
	daemonCommand.PersistentFlags().StringSlice(f.Service.FeatureGates, []string{}, `Feature gates in format "feature=true/false". Supported features are: "MachinePools"`)
```

service/controller/cluster.go
kopiczko: Some controllers need to store fields like mutex that's why we do it this way.

service/service.go
```diff
@@ -78,6 +80,22 @@ func New(config Config) (*Service, error) {
...
+		if featureGatesParam != "" {
+			featureGates := strings.Split(featureGatesParam, ",")
+			for i := range featureGates {
+				feature := strings.Split(strings.TrimSpace(featureGates[i]), "=")
+				features[feature[0]], err = strconv.ParseBool(feature[1])
+				if err != nil {
+					return nil, microerror.Maskf(invalidConfigError, "feature gate %#q must be either true or false", feature[0])
+				}
+			}
+		}
```
kopiczko, line 96:
```suggestion
		featureGates := config.Viper.GetStringSlice(config.Flag.Service.FeatureGates)
		for i := range featureGates {
			feature := strings.Split(strings.TrimSpace(featureGates[i]), "=")
			features[feature[0]], err = strconv.ParseBool(feature[1])
			if err != nil {
				return nil, microerror.Maskf(invalidConfigError, "feature gate %#q must be either true or false", feature[0])
			}
		}
```

fiunchinho:
Any opinions on using one flag for all gates vs using one flag for each?

puja108:
> Any opinions on using one flag for all gates vs using one flag for each?

k8s itself has a single flag accepting comma separated list for feature gates, but I have no strong opinion on this. I think at GS we want to aim at a low number of featurettes, i.e. we don't keep things in there for long, so it doesn't really matter if one or more. Maybe for it's usage a single one is easier to automate in testing and stuff.

tuommaki:
> Any opinions on using one flag for all gates vs using one flag for each?

With one flag you don't need to wire variables in different levels every time you add/remove a feature, but could have something like `feature.ParseGates(featureGates)` in `main.go` (with package private map of features) and then later where necessary: `if feature.Enabled("machinepool-controller") { /* do stuff */ }`.

fiunchinho:
Closing this. Let's revisit it if we need it further down the road.

PR #863, opened 2020-06-15
Removed MSI extension from masters and workers. It was unused.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/11559

Disable MSI extension on masters and workers VMSSes as it is unused.


whites11:
Still to be tested.

MartaD:
Does it require an update of Azure architecture (either internal or external) documentation?

whites11:
> Does it require an update of Azure architecture (either internal or external) documentation?

Nothing at all.
With the next release we will simply have some less stuff running in our VMs but no change in practice (as the stuff we're removing wasn't used anyway).
So no change needed in the documentation.

whites11:
Tested, it works.

fiunchinho:
Looks good! One more thing. I read [these docs](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template#new-service-principal) saying that this may fail if we don't add `"principalType": "ServicePrincipal"` to the properties. I'm using it in my PR and it works fine.

whites11:
> Looks good! One more thing. I read [these docs](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template#new-service-principal) saying that this may fail if we don't add `"principalType": "ServicePrincipal"` to the properties. I'm using it in my PR and it works fine.

Sounds good, added

PR #866, opened 2020-06-19
Update architect orb

fiunchinho:
Use latest version that includes retries when pushing to catalog.


fiunchinho:
Going with a ping.

PR #870, opened 2020-06-22
Make client factory not depend on AzureConfig

fiunchinho:
The factory that we use to create Azure clients is coupled to our `AzureConfig` CR. If we change to CAPI types, we can't use the factory. This is blocking me for the node pools story.

I thought that one potential solution we could make it CR independant by creating a credentials provider. In our case, the provider is kubernetes (we use k8s secrets). And the factory doesn't know how to read the credentials: it asks the credentials provider.

What do you think?

### DIFF
--- a/client/factory.go
+++ b/client/factory.go
@@ -11,20 +11,16 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
 	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage"
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
-	"github.com/giantswarm/k8sclient/v3/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	gocache "github.com/patrickmn/go-cache"
 
 	"github.com/giantswarm/azure-operator/v4/pkg/credential"
-	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
 const (
 	cacheHitLogKey       = "cacheHit"
 	clientTypeLogKey     = "clientType"
-	clusterIDLogKey      = "clusterID"
 	credentialNameLogKey = "credentialName"
 	logLevelLogKey       = "level"
 	logLevelDebug        = "debug"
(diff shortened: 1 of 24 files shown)


client/factory.go
```diff
@@ -32,19 +30,19 @@ const (
...
 }
 
 // Factory is creating Azure clients for specified AzureConfig CRs, so basically for specified
 // tenant clusters. All created clients are cached.
 type Factory struct {
-	gsTenantID string
-	k8sClient  k8sclient.Interface
-	logger     micrologger.Logger
-	mutex      sync.Mutex
+	credentialProvider credential.Provider
```
fiunchinho, line 42: Instead of directly fetching the credentials, the factory uses the provider. This provider is configured with the GiantSwarm tenant ID.

client/factory.go
```diff
@@ -85,8 +75,8 @@ func NewFactory(config FactoryConfig) (*Factory, error) {
 // GetDeploymentsClient returns DeploymentsClient that is used for management of deployments and
 // ARM templates. The client (for specified cluster) is cached after creation, so the same client
 // is returned every time.
-func (f *Factory) GetDeploymentsClient(cr v1alpha1.AzureConfig) (*resources.DeploymentsClient, error) {
-	client, err := f.getClient(cr, "DeploymentsClient", newDeploymentsClient)
+func (f *Factory) GetDeploymentsClient(credentialNamespace, credentialName string) (*resources.DeploymentsClient, error) {
```
fiunchinho, line 78: Instead of passing a CR, we pass the namespace and name for the credentials that we want to use. This makes the factory CR independent.

pkg/credential/credential.go
```diff
@@ -23,37 +21,42 @@ const (
...
+	gsTenantID string
+}
+
 // GetOrganizationAzureCredentials returns the organization's credentials.
 // This means a configured `ClientCredentialsConfig` together with the subscription ID and the partner ID.
 // The Service Principals in the organizations' secrets will always belong the the GiantSwarm Tenant ID in `gsTenantID`.
-func GetOrganizationAzureCredentials(ctx context.Context, k8sClient k8sclient.Interface, cr providerv1alpha1.AzureConfig, gsTenantID string) (auth.ClientCredentialsConfig, string, string, error) {
-	credential := &v1.Secret{}
-	err := k8sClient.CtrlClient().Get(ctx, client.ObjectKey{Namespace: key.CredentialNamespace(cr), Name: key.CredentialName(cr)}, credential)
+func (k K8SCredential) GetOrganizationAzureCredentials(ctx context.Context, credentialNamespace, credentialName string) (auth.ClientCredentialsConfig, string, string, error) {
```
fiunchinho, line 32: This is our current credentials provider implementation.

client/factory.go
nprokopic: Great idea! I was also thinking about making it more generic, just didn't expect that we will have a need so soon :smile:

client/factory.go
nprokopic: > This makes the factory CR independent. Do you think it makes sense?

It does.

> is it awkward?

No. :) 

WDYT about having a single parameter, maybe something like `credentialKey`, `credentialConfig`, `credentialID`, so it's more neutral? (then you can slice it and dice it in the provider code as you wish).

I'm thinking that getting credential by namespace and name is maybe too specific for it to be neutral. On the other hand if some other hypothetical credential provider needs only one string, then one of the two parameters would simply be ignored.

fiunchinho: Talked over slack. Creating such parameter would make the clients of the provider having to know how to _encode_ the namespace and name of the credential. Furthermore we only have the k8s provider anway, so let's not overengineer it.

PR #877, opened 2020-06-30
Sentry

whites11:
This PR adds Sentry support to the azure operator by bumping the `operatorkit` version to the latest one.
The Sentry integration is disabled by default but can be enabled by editing the azure operator configuration in the `installations` repository.
So IMHO it's worth adding the support before we release 11.5.0


whites11:
> We'd need to update the Helm chart as well, so that the parameter is passed to the operator during the install process.

Can we do it if the value is not in the Installations repo yet?

whites11:
> > We'd need to update the Helm chart as well, so that the parameter is passed to the operator during the install process.
> 
> Can we do it if the value is not in the Installations repo yet?

Yes we can, tested locally and it works. Pushed.

PR #878, opened 2020-06-30
Added changes since 4.1.0 in the changelog

whites11:
This PR adds entries to the changelog regarding changes happened since the 4.1.0 release.


whites11:
> Not 100% sure if this includes everything (didn't cross-check git log) but wording for these seems good to me.

All the major stuff should be there. Will wait some more time to let others have a look at their features and then merge. Thx

PR #879, opened 2020-06-30
Make vmss check to use the azure client factory

fiunchinho:
I'm using the factory in my nodepools branch, and I need this part to be different. Instead of adding the changes to my PR, I think it's better to fix it on master.


fiunchinho:
Done

PR #881, opened 2020-07-01
Create cloud-config handler for CAPI types

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/11926

We have an operatorkit handler that saves an Azure blob object with the `cloud-config` files required for our nodes. In order to render these files we need some information about the cluster being created, and until now, this operatorkit resource was executed receiving the old `AzureConfig` CR which contained all the required information.

But now we want to create the `cloud-config` files when we create different node pools. So I copied the whole `blobobject` resource and call it `cloudconfig`. They are the same, but
- `cloudconfig` will be executed on `AzureMachinePool` events, using info from `AzureMachinePool`, `AzureCluster`, `MachinePool` and `Cluster`.
- `blobobject` will be executed on `AzureConfig` events, using info from `AzureConfig`.

### DIFF
--- a/service/controller/cloudconfig/types.go
+++ b/service/controller/cloudconfig/types.go
@@ -5,6 +5,10 @@ import (
 	"github.com/giantswarm/certs"
 	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v6/pkg/template"
 	"github.com/giantswarm/randomkeys"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	expcapzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
+	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	expcapiv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 )
 
 type templateData struct {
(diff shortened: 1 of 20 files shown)


service/controller/cloudconfig/types.go
```diff
@@ -49,8 +53,12 @@ type ingressLBFileParams struct {
 }
 
 type IgnitionTemplateData struct {
-	CustomObject providerv1alpha1.AzureConfig
-	ClusterCerts certs.Cluster
-	ClusterKeys  randomkeys.Cluster
-	Images       k8scloudconfig.Images
+	AzureMachinePool *expcapzv1alpha3.AzureMachinePool
```
fiunchinho, line 56: We need to add more fields so we can fetch all the information that we need.

service/controller/resource/cloudconfig/desired.go
```diff
@@ -0,0 +1,164 @@
...
+			return nil, microerror.Mask(err)
+		}
+
+		defaultVersions := key.DefaultVersions()
+		versions.KubernetesAPIHealthz = defaultVersions.KubernetesAPIHealthz
+		versions.KubernetesNetworkSetupDocker = defaultVersions.KubernetesNetworkSetupDocker
+		images := k8scloudconfig.BuildImages(r.registryDomain, versions)
+
+		ignitionTemplateData = cloudconfig.IgnitionTemplateData{
+			AzureMachinePool: &cr,
```
fiunchinho, line 84: This should be the only difference between the old `blobobject` resource and this one.

PR #883, opened 2020-07-02
Subnet handler for AzureCluster controller

fiunchinho:

This PR adds a new operatorkit handler that creates a new subnet for every subnet defined in `AzureCluster`.


service/controller/resource/deployment/template/main.json
```diff
@@ -810,56 +810,7 @@
...
-                    }
-                  },
-                  {
-                    "name":"[parameters('vnetGatewaySubnetName')]",
-                    "properties":{
-                      "addressPrefix":"[parameters('vpnSubnetCidr')]"
-                    }
-                  }
-                ]
+                }
```
fiunchinho, line 813: We no longer create the subnets as part of the virtual network definition, but as separate deployments.

service/controller/resource/deployment/resource.go
```diff
@@ -263,7 +263,7 @@ func (r *Resource) enrichControllerContext(ctx context.Context, customObject pro
 	}
 
 	{
-		v, err := r.getDeploymentOutputValue(ctx, deploymentsClient, resourceGroupName, "virtual_network_setup", "masterSubnetID")
+		v, err := r.getDeploymentOutputValue(ctx, deploymentsClient, resourceGroupName, "legacy_masters_subnet", "masterSubnetID")
```
fiunchinho, line 266: now every subnet has its own deployment

service/controller/resource/deployment/deployment.go
```diff
@@ -77,3 +84,39 @@ func getVPNSubnet(customObject providerv1alpha1.AzureConfig) (*net.IPNet, error)
 
 	return &subnets.VPN, nil
 }
+
+func (r *Resource) initialProvisioning(ctx context.Context, customObject providerv1alpha1.AzureConfig) (string, error) {
```
fiunchinho, line 88: forgive me father for i have sinned

service/controller/resource/deployment/template/main.json
```diff
@@ -732,6 +732,7 @@
     {
       "apiVersion":"2016-09-01",
       "name":"virtual_network_setup",
+      "condition":"[equals(parameters('initialProvisioning'), 'Yes')]",
```
fiunchinho, line 735: We don't apply the virtual_network_setup deployment unless the three expected subnets are already in place. This means that if we want to change the name or cidr of the virtual network, we can't do it using the ARM deployment.

fiunchinho:
I've tested this PR and it's working fine. I used the [4.2.0 milestone](https://github.com/giantswarm/azure-operator/milestone/2). @whites11 I believe you are working on the next release? Do you think it would be better to wait for the 4.2.0 release before merging this PR? or can I merge?

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	armTemplate, err := subnet.GetARMTemplate()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, allocatedSubnet := range cr.Spec.NetworkSpec.Subnets {
```
fiunchinho, line 96: For every subnet declared in the `AzureCluster.Spec.NetworkSpec.Subnets` field, we submit a deployment to Azure to create the subnet.

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,273 @@
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+			reconciliationcanceledcontext.SetCanceled(ctx)
+
+			return nil
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "template and parameters unchanged")
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment is in state '%s'", *currentDeployment.Properties.ProvisioningState))
+
+		if key.IsFailedProvisioningState(*currentDeployment.Properties.ProvisioningState) {
```
fiunchinho, line 143: We remove the deployment if it's in `Failed` state.

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,273 @@
...
+			}
+
+			return nil
+		}
+	}
+
+	return nil
+}
+
+func (r *Resource) isDeploymentOutOfDate(ctx context.Context, cr capzv1alpha3.AzureCluster, currentDeployment azureresource.DeploymentExtended) (bool, error) {
```
fiunchinho, line 158: This functions decides whether or not the ARM deployment is out of date. For that, we use [the `Generation` field](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#objectmeta-v1-meta) from the AzureCluster CR. This `Generation` field should change when there is a change in the CR. This means we will submit the ARM template when `AzureCluster` changes, even though there are no changes in the subnets themselves, but I think that's fine.

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,273 @@
...
+	return crVersion != deploymentVersion, nil
+}
+
+func (r *Resource) getDeploymentParameters(ctx context.Context, clusterID, azureClusterVersion, virtualNetworkName string, allocatedSubnet *capzv1alpha3.SubnetSpec) (map[string]interface{}, error) {
+	return map[string]interface{}{
+		"azureClusterVersion": azureClusterVersion,
+		"nateGatewayName":     "workers-nat-gw",
+		"nodepoolName":        allocatedSubnet.Name,
+		"routeTableName":      fmt.Sprintf("%s-%s", clusterID, "RouteTable"),
+		"securityGroupName":   fmt.Sprintf("%s-%s", clusterID, "WorkerSecurityGroup"),
```
fiunchinho, line 181: I'm assuming here that the route table and security group will have this name. Do you think it will be better to fetch them from the ARM Deployment outputs? It will add Azure API calls, but this code will be independent from the naming.

service/controller/azure_cluster_controller.go
```diff
@@ -165,10 +170,52 @@ func newAzureClusterResources(config AzureClusterConfig, certsSearcher certs.Int
...
+		newDebugger, err = debugger.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var subnetResource resource.Interface
+	{
+		c := subnet.Config{
+			AzureClientsFactory: clientFactory,
```
nprokopic, line 202: Not really important here, but I was thinking that we could start injecting only those clients that some resource actually needs.
With the current way we are basically doing service-locator pattern.

service/controller/resource/subnet/error.go
```diff
@@ -0,0 +1,59 @@
...
+var missingOutputValueError = &microerror.Error{
+	Kind: "missingOutputValueError",
+}
+
+// IsMissingOutputValue asserts missingOutputValueError.
+func IsMissingOutputValue(err error) bool {
+	return microerror.Cause(err) == missingOutputValueError
+}
+
+var tooManyCredentialsError = &microerror.Error{
```
nprokopic, line 49: I've been told that we always need those error matcher functions (like `IsMissingOutputValue` here) for every type of error we add (even though it is not used :shrug:).

service/controller/resource/subnet/resource.go
nprokopic: > This handler makes no assumptions, it just creates the subnet using all the details from AzureCluster.

Love this! :rocket:

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,277 @@
...
+	armTemplate, err := subnet.GetARMTemplate()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, allocatedSubnet := range cr.Spec.NetworkSpec.Subnets {
+		deploymentName := fmt.Sprintf("%s-%s", mainDeploymentName, allocatedSubnet.Name)
+		currentDeployment, err := deploymentsClient.Get(ctx, key.ClusterID(&cr), deploymentName)
+		if IsNotFound(err) {
+			// fallthrough
```
nprokopic, line 102: Can we do this, since `currentDeployment` is used later in code? For example there is this few blocks later:

```
shouldSubmitDeployment := currentDeployment.IsHTTPStatus(404)
if !shouldSubmitDeployment {
	shouldSubmitDeployment, err = r.isDeploymentOutOfDate(ctx, cr, currentDeployment)
	if err != nil {
		return microerror.Mask(err)
	}
}
```

service/controller/resource/subnet/resource.go
nprokopic: What about having functions in `key` package for getting NAT gateway, route table and security group name?

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,277 @@
...
+
+// EnsureDeleted is a noop since the deletion of deployments is redirected to
+// the deletion of resource groups because they garbage collect them.
+func (r *Resource) createDeployment(ctx context.Context, deploymentsClient *azureresource.DeploymentsClient, resourceGroup, deploymentName string, desiredDeployment azureresource.Deployment) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring subnets deployments")
+
+	res, err := deploymentsClient.CreateOrUpdate(ctx, resourceGroup, deploymentName, desiredDeployment)
+	if err != nil {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment failed; deployment: %#v", desiredDeployment), "stack", microerror.JSON(microerror.Mask(err)))
+		return microerror.Mask(err)
```
nprokopic, line 199: maybe reuse error variable here:
```
reterr := microerror.Mask(err)
r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment failed; deployment: %#v", desiredDeployment), "stack", microerror.JSON(reterr))
return reterr
```

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	deploymentExtended, err := deploymentsClient.CreateOrUpdateResponder(res.Response())
+	if err != nil {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deployment failed; deployment: %#v", deploymentExtended), "stack", microerror.JSON(microerror.Mask(err)))
```
nprokopic, line 203: Same for reusing error variable in logging and return statement (not fully sure what's better).

service/controller/azure_cluster_controller.go
nprokopic: Just not sure if it is possible, as we need credential name to create the client.

In that case injecting a factory method would probably be cleaner, something like:
```
func (namespace string, name string) DeploymentsClient`
```

fiunchinho: Yeah, that would be great. The main problem right now is that we need the runtime object (CR) in order to create the clients. We can't create the object graph upfront, you know? It's tricky.

service/controller/resource/subnet/resource.go
fiunchinho: We could do that. I would need to change the `deployment` handler as well, because the naming is hardcoded in the ARM template.

service/controller/resource/subnet/resource.go
fiunchinho: if it doesn't exist `shouldSubmitDeployment` will be true. That code path always returns, so other occurrences of `currentDeployment` won't use a `nil` variable. Is that what you meant? Should I set `shouldSubmitDeployment` to true on line 102 instead?

service/controller/azure_cluster_controller.go
nprokopic: > The main problem right now is that we need the runtime object (CR) in order to create the clients. We can't create the object graph upfront, you know? It's tricky.

Yeah, we definitely cannot pass in an already created client. We could use creator functions though, and then we could simply pass in factory functions like `factory.GetDeploymentsClient`.

I remember that this pattern is typical in (good) .NET DI libraries, as an alternative to directly injecting an already created dependency object, you can inject a factory method (in case like we have now, only with DI container doing some reflection magic for you :smile:).

But this is out of scope of this PR :)

service/controller/resource/subnet/resource.go
nprokopic: Sorry, I wasn't clear enough. I meant that if condition `IsNotFound(err)` is met, that would mean that `err != nil`, which would mean that returned `currentDeployment` was `nil`, so this line `currentDeployment.IsHTTPStatus(404)` would panic.

nprokopic: That's if I got that right.

service/controller/resource/subnet/resource.go
nprokopic: It would be ideal to have the names in one place. Not sure how we did it before for `masters` and `instance` resource. Not a blocker this one :)

service/controller/resource/subnet/resource.go
fiunchinho: We always get a `Response` back, even when the deployment is not found. So `currentDeployment` can't be nil.

service/controller/resource/subnet/resource.go
```diff
@@ -0,0 +1,290 @@
...
+
+		if currentDeployment.IsHTTPStatus(404) || isDeploymentOutOfDate(cr, currentDeployment) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "template or parameters changed")
+			err = r.createDeployment(ctx, deploymentsClient, key.ClusterID(&cr), deploymentName, desiredDeployment)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+			reconciliationcanceledcontext.SetCanceled(ctx)
```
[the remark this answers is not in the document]
fiunchinho: IIUC, I'm only canceling the reconciliation after submitting the ARM deployment. So it's not actually blocking the reconciliation of the subnets. I don't think that deadlock scenario could be reached.

PR #886, opened 2020-07-07
Allow zero worker control plane

tuommaki:
Don't require AzureMachine CRs for CP creation. This should be first step to
prevent creation of baked-in workers in node pool clusters.


tuommaki:
Tested also with "regular" way of cluster creation and it succeeded fine once I used correct version of etcd (due to recent `k8scloudconfig` upgrade). Going with this.

PR #890, opened 2020-07-07
Only submit vnet ARM deployment the first time we deploy

fiunchinho:

Until now we were creating the vpn, masters and workers subnets as part of the Virtual Network definition. This means that when the ARM deployment gets re-applied, subnets added to the virtual network would be removed, since they are not part of the definition.

~~This PR changes how we create the subnets, by creating a separate ARM deployment for every subnet, decoupling the virtual network definition from its subnets.~~

~~I've changed the code to conditionally deploy the vnet ARM deployment. It will only be submitted if the default subnets are missing (vpn, masters, workers). That way subnets created by other processes won't be removed.~~

I've changed the code again. Now it will conditionally deploy the vnet ARM deployment depending on whether or not it's the first time we deploy.


service/controller/resource/deployment/deployment.go
```diff
@@ -77,3 +88,39 @@ func getVPNSubnet(customObject providerv1alpha1.AzureConfig) (*net.IPNet, error)
 
 	return &subnets.VPN, nil
 }
+
+func (r *Resource) initialProvisioning(ctx context.Context, customObject providerv1alpha1.AzureConfig) (string, error) {
```
fiunchinho, line 92: forgive me father for i have sinned

service/controller/resource/deployment/resource.go
```diff
@@ -263,7 +263,7 @@ func (r *Resource) enrichControllerContext(ctx context.Context, customObject pro
 	}
 
 	{
-		v, err := r.getDeploymentOutputValue(ctx, deploymentsClient, resourceGroupName, "virtual_network_setup", "masterSubnetID")
+		v, err := r.getDeploymentOutputValue(ctx, deploymentsClient, resourceGroupName, "legacy_masters_subnet", "masterSubnetID")
```
fiunchinho, line 266: now every subnet has its own deployment

service/controller/resource/deployment/deployment.go
nprokopic: `bool` wasn't good enough for you sinner here? :smile:

service/controller/resource/deployment/template/main.json
```diff
@@ -732,6 +732,7 @@
     {
       "apiVersion":"2016-09-01",
       "name":"virtual_network_setup",
+      "condition":"[equals(parameters('initialProvisioning'), 'Yes')]",
```
fiunchinho, line 735: This is the most relevant change. With this change, we only submit the Virtual Network deployment if we didn't find the three expected subnets (vpn, masters, workers). If we find them, this deployment is not submitted. Meaning we wouldn't be able to update the CIDR of an existing virtual network using the ARM deployment.

service/controller/resource/deployment/deployment.go
```diff
@@ -77,3 +88,39 @@ func getVPNSubnet(customObject providerv1alpha1.AzureConfig) (*net.IPNet, error)
...
+		}
+
+		err = result.NextWithContext(ctx)
+		if err != nil {
+			return IsInitialProvisioning, microerror.Mask(err)
+		}
+	}
+
+	initialProvisioning := "No"
+	if expectedSubnets != 3 {
```
nprokopic, line 121: What if it is 4 or more because there are 4 or more node pools?

fiunchinho: I'm only increasing if it matches the expected names. If we create more subnets for nodepools for example, they won't use the same name, hence they won't increase `expectedSubnets`.

service/controller/resource/deployment/deployment.go
fiunchinho: We were already using that parameter, so I didn't want to change it :dancers:

service/controller/resource/deployment/deployment.go
nprokopic: Oh yes, missed that.

service/controller/resource/deployment/template/main.json
nprokopic: So if we ever need to modify VNet somehow, we cannot use this ARM template anymore and we will have to do it with SDK?

fiunchinho: Yes. The deployment would not be submitted since it already exists, and changes done using the SDK would be kept.

fiunchinho:
I removed the code to move subnets to their own ARM deployment. I think we only need the conditional approach for the whole thing.

fiunchinho:
The parameter `initialProvisioning` was already being passed when the cluster-main-template deployment was not found, so I removed the logic that tries to find subnets.

service/controller/resource/deployment/template/main.json
fiunchinho: I changed the code: we will only submit the virtual network deployment if the `cluster-main-template` deployment is not found.

fiunchinho:
I've tested this PR and it's working fine. I used the 4.2.0 milestone. @whites11 I believe you are working on the next release? Do you think it would be better to wait for the 4.2.0 release before merging this PR? or can I merge?

whites11:
> I've tested this PR and it's working fine. I used the 4.2.0 milestone. @whites11 I believe you are working on the next release? Do you think it would be better to wait for the 4.2.0 release before merging this PR? or can I merge?

At some point we need to stop merging stuff to master because it gets complicated to perform tests.
As of now we're not running tests yet, so I guess you can merge it (if it works)

fiunchinho:
Yes, it's working fine. If we are not currently testing a new release, I'll go ahead and merge it if that's ok.

service/controller/resource/deployment/template/main.json
nprokopic: This is then a temporary change IIUC, as it is probably just a matter of time when we will need to change something around VNet again, so in the future will be doing that from code.

tuommaki: Could this be changed to only consider vpn & masters. I just had a cluster without workers and technically the CP is okayish (I know there aren't resources for some deployments but that's different story).

tuommaki: Stupid me. Wrote & submitted before reading till the end. All good.

tuommaki: ...but where's the related code?

fiunchinho: What related code? [This](https://github.com/giantswarm/azure-operator/blob/ae43c102dae983790a716a2110d1690482e0c267/service/controller/resource/deployment/resource.go#L115-L118)?

PR #891, opened 2020-07-08
Avoid dependency with controllercontext

fiunchinho:

[The `cloudconfig` handler was copied from the `blobobject` handler](https://github.com/giantswarm/azure-operator/pull/881). The latter was depending on another handler setting values in the controller context. With these changes, the `cloudconfig` handler is independent and doesn't need other handlers.

### DIFF
--- a/pkg/credential/spec.go
+++ b/pkg/credential/spec.go
@@ -9,3 +9,9 @@ import (
 type Provider interface {
 	GetOrganizationAzureCredentials(ctx context.Context, credentialNamespace, credentialName string) (auth.ClientCredentialsConfig, string, string, error)
 }
+type EmptyProvider struct {
+}
+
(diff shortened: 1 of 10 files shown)


pkg/credential/spec.go
```diff
@@ -9,3 +9,9 @@ import (
 type Provider interface {
 	GetOrganizationAzureCredentials(ctx context.Context, credentialNamespace, credentialName string) (auth.ClientCredentialsConfig, string, string, error)
 }
+type EmptyProvider struct {
```
fiunchinho, line 12: I need a dummy implementation for unit tests.

service/controller/resource/cloudconfig/desired.go
```diff
@@ -41,6 +41,11 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	release, err := r.getReleaseFromMetadata(ctx, azureCluster.ObjectMeta)
```
fiunchinho, line 44: Fetch release using the release label, instead of the `controllercontext`.

PR #892, opened 2020-07-09
Don't return error if TC clients not available on nodes upgrade check

tuommaki:
It could happen that on cluster creation a broken VM gets created and failing
on `masters` resource prevents vpn connection to be created and therefore
blocks reconciliation of cluster.


tuommaki:
Uups, I accidentally merged this commit in my other PR. No diff here anymore; anyways, this got tested and worked ;)

PR #895, opened 2020-07-09
Avoid dependency with controllercontext

fiunchinho:
Following up on the work started [here](https://github.com/giantswarm/azure-operator/pull/891), I'm removing the last `controllercontext` dependency (I hope) from the `cloudconfig` handler.

### DIFF
--- a/service/controller/resource/cloudconfig/create.go
+++ b/service/controller/resource/cloudconfig/create.go
@@ -7,12 +7,31 @@ import (
 	"github.com/giantswarm/microerror"
 
 	"github.com/giantswarm/azure-operator/v4/service/controller/blobclient"
-	"github.com/giantswarm/azure-operator/v4/service/controller/controllercontext"
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	cc, err := controllercontext.FromContext(ctx)
+	cr, err := key.ToAzureMachinePool(obj)
 	if err != nil {
+		return microerror.Mask(err)
+	}
+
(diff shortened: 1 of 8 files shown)


service/controller/resource/cloudconfig/desired.go
```diff
@@ -95,9 +94,33 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		cloudConfig, err = cloudconfig.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
 	output := []ContainerObjectState{}
 	{
-		b, err := cc.CloudConfig.NewMasterTemplate(ctx, ignitionTemplateData, encrypter)
+		b, err := cloudConfig.NewMasterTemplate(ctx, ignitionTemplateData, encrypter)
```
fiunchinho, line 123: This is the relevant change. We create the cloudconfig object instead of fetching it from the `controllercontext`.

service/controller/resource/cloudconfig/create.go
```diff
@@ -24,7 +43,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 	for _, containerObject := range containerObjectToCreate {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating container object %#q", containerObject.Key))
 
-		_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, cc.ContainerURL)
+		_, err := blobclient.PutBlockBlob(ctx, containerObject.Key, containerObject.Body, containerURL)
```
fiunchinho, line 46: Using the same approach than in `current.go`.

fiunchinho:
the `cloudconfig` handler will be replaced by `spark` and `cloudconfigblob`

PR #898, opened 2020-07-13
State diagram: Upgrade to node pools cluster

tuommaki:
Introduced a node pools document and placed CRD structure + upgrade process SVGs into it.

Towards https://github.com/giantswarm/giantswarm/issues/10130


tuommaki:
I feel like this needs improving, but I lack the ideas of what. Please provide critique so we can make this better. The PlantUML formatting is super cool and it's nice to work with in VS.Code :rocket:

nprokopic:
After deleting the old VMSS, we will need to stop submitting VMSS template for workers, so we would need some check there before submitting it, or maybe even entirely remove submitting VMSS (as there will be an upgrade path after which the current workers VMSS template is not required).

tuommaki:
> or maybe even entirely remove submitting VMSS 

My original intention was that there shouldn't be reason to reconcile worker deployment anymore once the upgrade is in progress. Can someone think of anything why this would be needed?

fiunchinho:
> After deleting the old VMSS, we will need to stop submitting VMSS template for workers, so we would need some check there before submitting it, or maybe even entirely remove submitting VMSS (as there will be an upgrade path after which the current workers VMSS template is not required).

After the upgrade, the `AzureConfig` CR will stop being reconciled by older azure-operator versions. The one that will reconcile it  will only submit VMSS templates for `AzureMachinePool` CR's. So I believe by design it will stop submitting the VMSS template for workers.

@tuommaki "ensure first node pool" would mean ensuring that `AzureMachinePool` exists, so that the operator creates it?

nprokopic:
> My original intention was that there shouldn't be reason to reconcile worker deployment anymore once the upgrade is in progress.

> After the upgrade, the AzureConfig CR will stop being reconciled by older azure-operator versions. The one that will reconcile it will only submit VMSS templates for AzureMachinePool CR's. So I believe by design it will stop submitting the VMSS template for workers.

Sounds good!

tuommaki:
> @tuommaki "ensure first node pool" would mean ensuring that AzureMachinePool exists, so that the operator creates it?

Yeah, my thinking has been that when the new operator starts reconciling, there's the migration handler that checks if
1. Old built-in VMSS exists?
2. If node pool with same specs exists?
3. If not, then create one.
4. Migrate the workload
5. Cleanup

tuommaki:
Ok. Mermaid flowchart (the plain `graph`) doesn't seem to support multiple lines of text in boxes. As this won't be needed too long, I propose to stop here. Approval?

tuommaki:
Going with a ping

PR #903, opened 2020-07-15
Name persister accordingly, use controllerclient

fiunchinho:
Rename the ipam `ClusterChecker` to `AzureConfigChecker` since it's the one used for `AzureConfig` CRs. For `AzureMachinePool` I'll create the `AzureMachinePoolChecker`.

I also changed it so it uses the generic controller client.


service/controller/resource/ipam/azure_config_checker.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
+)
+
+type AzureConfigCheckerConfig struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+}
+
+type AzureConfigChecker struct {
```
fiunchinho, line 19: This replaces `ClusterChecker`.

PR #904, opened 2020-07-15
Handle node pool deletion

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12213

Notice that the target branch is not `master` but the branch for the node pools PR.

When a delete event is watched, we:
- delete the subnet from the `AzureCluster` CR (this will trigger an update event and the `AzureCluster` controller will remove the subnet)
- delete the ARM deployment for the VMSS node pool
- delete the VMSS

### DIFF
--- a/service/controller/resource/instance/delete.go
+++ b/service/controller/resource/instance/delete.go
@@ -94,15 +94,15 @@ func (r *Resource) deleteARMDeployment(ctx context.Context, credentialSecret *v1
 	r.Logger.LogCtx(ctx, "message", "Deleting machine pool ARM deployment")
 
 	deploymentsClient, err := r.ClientFactory.GetDeploymentsClient(credentialSecret.Namespace, credentialSecret.Name)
-	if IsDeploymentNotFound(err) {
-		r.Logger.LogCtx(ctx, "message", "Machine pool ARM deployment was already deleted")
-		return nil
-	} else if err != nil {
+	if err != nil {
 		return microerror.Mask(err)
 	}
 
 	_, err = deploymentsClient.Delete(ctx, resourceGroupName, deploymentName)
-	if err != nil {
+	if IsDeploymentNotFound(err) {
+		r.Logger.LogCtx(ctx, "message", "Machine pool ARM deployment was already deleted")
+		return nil
+	} else if err != nil {
 		return microerror.Mask(err)
 	}
 
--- a/service/controller/resource/instance/delete.go
+++ b/service/controller/resource/instance/delete.go
@@ -35,12 +35,12 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	err = r.removeSubnetFromAzureCluster(ctx, azureCluster, azureMachinePool.Name)
+	err = r.removeNodePool(ctx, azureMachinePool, credentialSecret)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	err = r.removeNodePool(ctx, azureMachinePool, credentialSecret)
+	err = r.removeSubnetFromAzureCluster(ctx, azureCluster, azureMachinePool.Name)
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/service/controller/resource/instance/delete.go
+++ b/service/controller/resource/instance/delete.go
@@ -89,6 +89,7 @@ func (r *Resource) removeNodePool(ctx context.Context, azureMachinePool capzexpv
 	return nil
 }
 
+// deleteARMDeployment deletes the ARM deployment from Azure.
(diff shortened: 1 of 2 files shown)


service/controller/resource/instance/delete.go
```diff
@@ -2,10 +2,133 @@ package instance
 
 import (
 	"context"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	capzexpv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
```
tuommaki, line 9: Not sure if it matters or should be aligned, but I have been using `expcapzv1alpha3` shorthand for this.

service/controller/resource/instance/delete.go
```diff
@@ -2,10 +2,133 @@ package instance
...
+
+	err = r.removeSubnetFromAzureCluster(ctx, azureCluster, azureMachinePool.Name)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = r.removeNodePool(ctx, azureMachinePool, credentialSecret)
+	if err != nil {
+		return microerror.Mask(err)
+	}
```
tuommaki, line 46: Should these be ordered the other way round? Subnet can't be deleted if there are still VMs running on it. It might be that the "delay" here is still not enough and it takes two reconciliation rounds to complete, but from logical point of view I'd may switch. WDYT?

service/controller/resource/instance/delete.go
```diff
@@ -2,10 +2,133 @@ package instance
...
+	subnetPosition := -1
+	for i, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
+		if subnet.Name == subnetName {
+			subnetPosition = i
+		}
+	}
+
+	subnetIsFound := subnetPosition >= 0
+	if subnetIsFound {
+		azureCluster.Spec.NetworkSpec.Subnets = append(azureCluster.Spec.NetworkSpec.Subnets[:subnetPosition], azureCluster.Spec.NetworkSpec.Subnets[subnetPosition+1:]...)
```
tuommaki, line 61: :see_no_evil: Maybe one day there will be generics and one could have `func (s *Slice) Delete(idx int)`...

service/controller/resource/instance/delete.go
```diff
@@ -2,10 +2,133 @@ package instance
...
+
+	return nil
+}
+
+// deleteARMDeployment deletes the ARM deployment from Azure.
+func (r *Resource) deleteARMDeployment(ctx context.Context, credentialSecret *v1alpha1.CredentialSecret, resourceGroupName, deploymentName string) error {
+	r.Logger.LogCtx(ctx, "message", "Deleting machine pool ARM deployment")
+
+	deploymentsClient, err := r.ClientFactory.GetDeploymentsClient(credentialSecret.Namespace, credentialSecret.Name)
+	if IsDeploymentNotFound(err) {
```
tuommaki, line 97: Is this check in wrong place? The actual `Delete()` invocation is down below.

service/controller/resource/instance/delete.go
```diff
@@ -2,10 +2,133 @@ package instance
...
+	err = r.deleteARMDeployment(ctx, credentialSecret, key.ClusterID(&azureMachinePool), key.NodePoolDeploymentName(&azureMachinePool))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = r.deleteVMSS(ctx, credentialSecret, key.ClusterID(&azureMachinePool), key.NodePoolVMSSName(&azureMachinePool))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
```
tuommaki, line 88: What about security group? Didn't each node pool have their own or is it part of subnet management?

service/controller/resource/instance/delete.go
fiunchinho: Yeah, let's align. We can update them later.

service/controller/resource/instance/delete.go
fiunchinho: Yeah, it makes more sense that way, I'll change it.

service/controller/resource/instance/delete.go
fiunchinho: I've been using always the same security group: the one used in the workers subnet. Do you think it makes sense to create one for every node pool? I believe they would be identical?

service/controller/resource/instance/delete.go
fiunchinho: lol, my bad.

service/controller/resource/instance/delete.go
tuommaki: For now I guess they can be all the same. No need to change. Somewhere in the future we might want to define per-nodepool security restrictions, but not now. All good.

fiunchinho:
Updated, please take a look.

PR #905, opened 2020-07-15
Garbage collect subnets that shouldn't exist after deleting node pools

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12213

Please, notice that the target branch is not `master`, but the `subnet` handler PR.

This will remove subnets that exist in the virtual network but are not defined in the `AzureCluster` object and are not the legacy subnets (masters, workers, vpn subnets).

### DIFF
--- a/service/controller/resource/subnet/resource.go
+++ b/service/controller/resource/subnet/resource.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-11-01/network"
 	azureresource "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
 	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage"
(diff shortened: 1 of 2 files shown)


service/controller/resource/subnet/resource.go
```diff
@@ -187,6 +212,87 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for subnetsIterator.NotDone() {
+		subnetInAzure := subnetsIterator.Value()
+
+		if !isSubnetInAzureClusterSpec(ctx, azureCluster, *subnetInAzure.Name) && !isProtectedSubnet(*subnetInAzure.Name) {
+			err = r.deleteSubnet(ctx, subnetsClient, key.ClusterID(&azureCluster), azureCluster.Spec.NetworkSpec.Vnet.Name, *subnetInAzure.Name)
+			if IsSubnetInUse(err) {
```
fiunchinho, line 228: due to the asynchronous nature of the operators, it's possible that we try to remove the subnet even though the VMSS is still there. I don't want to show an error since this is expected. Instead, I create an error matcher and when that happens we just wait for the next reconciliation loop.

PR #906, opened 2020-07-17
Add Spark handler to create ignition data secret

tuommaki:
Spark handler is checking if corresponding Spark CRD exists and renders
k8scloudconfig for a node pool based on existing CAPI & CAPZ CRs.

`blobobject` handler shall be then refactored to pull the ignition file from
corresponding secret pointed by `Spark.Status.DataSecretName`.

This handler sort of simulates `ignition-operator` and shall be removed once
`ignition-operator` can fully render ignition templates for Azure nodes.


service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,308 @@
...
+	cr, err := key.ToAzureMachinePool(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	var sparkCR corev1alpha1.Spark
+	{
+		err = r.ctrlClient.Get(ctx, client.ObjectKey{Namespace: cr.Namespace, Name: cr.Name}, &sparkCR)
+		if errors.IsNotFound(err) {
+			// TODO: Cancel reconciliation until Spark CR is there.
```
tuommaki, line 44: Probably just missing `return nil` here.

service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+	}
+
+	var cloudConfig cloudconfig.Interface
+	{
+		// TODO: Construct cloudconfig instance.
+		//
+		// It probably requires that some configuration settings are wired from
+		// configmap to resource. Some can be hopefully hardcoded here. Some
+		// should be pulled from the outside configmap later (such as OIDC).
```
tuommaki, line 176: This is the tricky part. Theoretically it's nothing more than just wiring up big pile of configuration and pulling other from available CRs. OIDC config doesn't exist right now, but we need to find a place to put it. Then it can be also configured here.

service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,308 @@
...
+	return base + "-machine-pool-ignition"
+}
+
+// toAzureConfig performs mapping from CAPI & CAPZ types to AzureConfig which
+// fulfills k8scloudconfig requirements for rendering node pool worker node
+// ignition templates.
+func toAzureConfig(cluster *capiv1alpha3.Cluster, azureCluster *capzv1alpha3.AzureCluster, machinePool *expcapiv1alpha3.MachinePool, azureMachinePool *expcapzv1alpha3.AzureMachinePool) providerv1alpha1.AzureConfig {
+	ac := providerv1alpha1.AzureConfig{}
+
+	// TODO: Implement <3
```
tuommaki, line 305: This is pretty much just annoying struct construction by copying relevant fields from available CRs.

tuommaki: I'm not fully sure if my grep finds everything, but there might not be _that_ many fields to set. Note that these are from [CommonClusterType](https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/provider/v1alpha1/azure_types.go#L35):
```
k8scloudconfig $ grep -R '\.Cluster\.' | sed -s 's|.*\(\.Cluster[.a-zA-Z]*\).*|\1|g'|sort |uniq
.Cluster.Calico.CIDR
.Cluster.Calico.MTU
.Cluster.Docker.Daemon.CIDR
.Cluster.Etcd.Domain
.Cluster.Etcd.Prefix
.ClusterIPRange
.Cluster.Kubernetes.API.Domain
.Cluster.Kubernetes.API.SecurePort
.Cluster.Kubernetes.CloudProvider
.Cluster.Kubernetes.DNS.IP
.Cluster.Kubernetes.Domain
.Cluster.Kubernetes.Kubelet.Domain
.Cluster.Kubernetes.Kubelet.Labels
.Cluster.Kubernetes.NetworkSetup.KubeProxy.ConntrackMaxPerCore
.Cluster.Kubernetes.SSH.UserList
```

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return "", microerror.Mask(err)
+		}
+	}
+
+	return string(cloudconfigSecret.Data[key.CloudConfigSecretKey]), nil
+}
+
+// getBootstrapSecretName will try to find Ignition CRs instead of Spark CRs when Ignition Operator is deployed.
+// It tries to find a Bootstrap CR which is named after the MachinePool. We may want to change it so we use `MachinePool.Spec.Template.Spec.Bootstrap`.
```
fiunchinho, line 110: [There are some fields in `MachinePool.Spec.Template.Spec`](https://github.com/kubernetes-sigs/cluster-api/blob/e3a4ed7977b6789e03cb9e4d6e5ca201e43a726e/api/v1alpha3/machine_types.go#L188-L209) that may be useful. Should we use them? I'm currently looking for a Spark CR that has the same name as the node pool, maybe I should use `ConfigRef` instead?

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -0,0 +1,178 @@
...
+
+		if !sparkCR.Status.Ready || sparkCR.Status.DataSecretName == "" {
+			return "", microerror.Mask(bootstrapCRNotReady)
+		}
+	}
+
+	return sparkCR.Status.DataSecretName, nil
+}
+
+func (r *Resource) getContainerURL(ctx context.Context, credentialSecret *v1alpha1.CredentialSecret, resourceGroupName, storageAccountName string) (azblob.ContainerURL, error) {
```
fiunchinho, line 128: This was taken from `controllercontext` in other resources/handlers. I think it's fine retrieving it from the API. WDYT?

service/controller/resource/cloudconfigblob/delete.go
```diff
@@ -0,0 +1,9 @@
+package cloudconfigblob
+
+import (
+	"context"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
fiunchinho, line 7: Should we remove the blob when node pool is removed?

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	machinePool, err := r.getOwnerMachinePool(ctx, azureMachinePool.ObjectMeta)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if machinePool == nil {
```
fiunchinho, line 35: If we don't have an owner lets fail and wait for next reconciliation loop. The [provider agnostic controllers](https://github.com/giantswarm/azure-operator/pull/928) should take care of this.

service/controller/resource/spark/delete.go
```diff
@@ -0,0 +1,45 @@
...
+
+	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/microerror"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
+)
+
+// EnsureDeleted will delete the `Spark` CR that was created for this specific node pool, and the `Secret` referenced by it.
```
fiunchinho, line 14: The secret will be removed by k8s because its `owner` is deleted.

service/controller/resource/spark/delete.go
```diff
@@ -0,0 +1,45 @@
...
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	azureMachinePool, err := key.ToAzureMachinePool(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	var sparkCR corev1alpha1.Spark
```
fiunchinho, line 21: Maybe we can just remove it, instead of fetching it and then removing it?

service/controller/resource/cloudconfigblob/create.go
tuommaki: Yeah, at least for now. We can iterate here later if we need to improve it somehow.

service/controller/resource/cloudconfigblob/delete.go
tuommaki: Yep, I think that would make sense.

fiunchinho: ok, added the deletion logic. I chose the option to include all the snapshots in the deletion, according to https://docs.microsoft.com/en-us/rest/api/storageservices/delete-blob#request-headers

service/controller/key/key.go
```diff
@@ -152,6 +153,10 @@ func BlobName(customObject LabelsGetter, role string) string {
 	return fmt.Sprintf("%s-%s-%s", OperatorVersion(customObject), cloudConfigVersion, role)
 }
 
+func BootstrapBlobName(customObject expcapzv1alpha3.AzureMachinePool) string {
```
fiunchinho, line 156: @tuommaki I added a new function to generate the blob name. I realized I was using the old `BlobName` function that would return the same name for all node pools. In this one I'm including cluster name, node pool name and operator version.

service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,639 @@
...
+			err = r.ctrlClient.Create(ctx, dataSecret)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Secret %#q created with ignition cloud config", secretName(sparkCR.Name)))
+		}
+	}
+
+	{
+		if !bytes.Equal(ignitionBlob, dataSecret.Data[key.CloudConfigSecretKey]) {
```
tuommaki, line 133: Remember to drop this update to take place after spark CR update.

service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,639 @@
...
+			XXX: Let's see if we can live without this. -tuommaki
+		{
+			clusterStatus, err := r.newClusterStatus(ctx, request, azureConfig.Labels[label.OperatorVersion])
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+
+			azureConfig.Status.Cluster = clusterStatus
+		}
+	*/
```
tuommaki, line 428:
```suggestion
```

Just Zap it :P

service/controller/resource/spark/delete.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// EnsureDeleted will delete the `Spark` CR that was created for this specific node pool, and the `Secret` referenced by it.
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	azureMachinePool, err := key.ToAzureMachinePool(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = r.ctrlClient.Delete(ctx, &corev1alpha1.Spark{
```
tuommaki, line 22:
```suggestion
        // With use of OwnerReference this deletion will also cascade to data secret object.
	err = r.ctrlClient.Delete(ctx, &corev1alpha1.Spark{
```

integration/test/crmapping/crmapping_test.go
```diff
@@ -354,21 +350,19 @@ func verifyCR(t *testing.T, client client.Client, testName string, obj runtime.O
...
+		t.Fatalf("err = %#q, want %#v", microerror.JSON(err), nil)
+	}
+
+	goldenMap := make(map[string]interface{})
+	err = yaml.Unmarshal(goldenFile, goldenMap)
+	if err != nil {
+		t.Fatalf("err = %#q, want %#v", microerror.JSON(err), nil)
+	}
+
+	if !cmp.Equal(objMap, goldenMap) {
```
tuommaki, line 365: Had to change the comparison because map ordering is not guaranteed in Go so labels were breaking in tests.

service/controller/resource/spark/create.go
```diff
@@ -0,0 +1,683 @@
...
+				return microerror.Mask(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *Resource) createIgnitionBlob(ctx context.Context, azureMachinePool *expcapzv1alpha3.AzureMachinePool) ([]byte, error) {
+	machinePool, err := r.getOwnerMachinePool(ctx, azureMachinePool.ObjectMeta)
```
fiunchinho, line 180: This should work with our current owner references. Also, I'm now using the utility function to find the cluster by name.

service/controller/resource/spark/delete.go
fiunchinho: This seems not to be true, so we explicitly remove the secret.

fiunchinho:
what's left to merge this PR?

tuommaki:
> what's left to merge this PR?

Testing, although we can do it when smashing all together (IPAM is also ~ready).

fiunchinho:
@tuommaki @nprokopic

tuommaki:
I created this PR so can't approve, but...
LGTM

PR #909, opened 2020-07-21
Create azureconfig from CAPI/CAPZ crs so we can render

fiunchinho:
We need an `AzureConfig` to be able to use `k8scloudconfig`, so we need to map from the CAPI/CAPZ types to our `AzureConfig`.

Some new parameters are needed for creating the `AzureConfig`. We didn't need those before because `kubernetesd` was setting those values when creating the `AzureConfig`, like `SSHUserList`, `EtcdPrefix`, etc.

I also removed the `cloudconfig` handler since it's not needed anymore.


service/controller/resource/spark/create.go
```diff
@@ -22,26 +26,37 @@ import (
...
 
+	"github.com/giantswarm/azure-operator/v4/pkg/label"
 	"github.com/giantswarm/azure-operator/v4/service/controller/cloudconfig"
 	"github.com/giantswarm/azure-operator/v4/service/controller/encrypter"
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
 const (
-	ignitionBlobKey = "ignitionBlob"
+	dockerVolumeSizeGB  = 50
```
whites11, line 36: Can't we un-hardcode these values? It would be nice to be able to set them on a per-cluster basis

fiunchinho: I don't think these values are even used.

We just need a valid `AzureConfig` object that renders the required cloudconfig ignition files. This part of the code has nothing to do with Azure disks.

service/controller/resource/spark/create.go
```diff
@@ -125,7 +140,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	{
-		if bytes.Compare(ignitionBlob, dataSecret.Data[ignitionBlobKey]) != 0 {
+		if !bytes.Equal(ignitionBlob, dataSecret.Data[ignitionBlobKey]) {
```
fiunchinho, line 143: The linter asked for this change.

service/controller/resource/spark/create.go
```diff
@@ -296,13 +342,310 @@ func secretName(base string) string {
...
+		if err != nil {
+			return availabilityZones, microerror.Mask(err)
+		}
+		availabilityZones = append(availabilityZones, intFd)
+	}
+
+	return availabilityZones, nil
+}
+
+func (r *Resource) buildAzureConfig(ctx context.Context, cluster *capiv1alpha3.Cluster, azureCluster *capzv1alpha3.AzureCluster, machinePool *expcapiv1alpha3.MachinePool, azureMachinePool *expcapzv1alpha3.AzureMachinePool) (providerv1alpha1.AzureConfig, error) {
```
fiunchinho, line 358: I copied the code from the `azureconfig` handler that @tuommaki added to create `AzureConfig` crs from CAPI crs.

service/controller/resource/spark/create.go
```diff
@@ -296,13 +342,310 @@ func secretName(base string) string {
...
+	azureConfig.Spec.Azure.DNSZones.Etcd.Name = hostDNSZone
+	azureConfig.Spec.Azure.DNSZones.Etcd.ResourceGroup = hostResourceGroup
+	azureConfig.Spec.Azure.DNSZones.Ingress.Name = hostDNSZone
+	azureConfig.Spec.Azure.DNSZones.Ingress.ResourceGroup = hostResourceGroup
+
+	{
+
+		masterNodes := []providerv1alpha1.AzureConfigSpecAzureNode{
+			{
+				VMSize:              "Standard_D4_v2",
```
fiunchinho, line 431: I don't think these values are even used when rendering cloudconfig, so anything would do.

main.go
```diff
@@ -131,8 +131,8 @@ func mainError() error {
 
 	daemonCommand.PersistentFlags().String(f.Service.Cluster.BaseDomain, "ghost.westeurope.azure.gigantic.io", "Cluster base domain without k8s/g8s prefixes.")
 
-	daemonCommand.PersistentFlags().Int(f.Service.Cluster.Calico.CIDR, 0, "Calico cidr of guest clusters.")
-	daemonCommand.PersistentFlags().Int(f.Service.Cluster.Calico.MTU, 0, "Calico MTU of guest clusters.")
+	daemonCommand.PersistentFlags().Int(f.Service.Cluster.Calico.CIDR, 16, "Calico cidr of guest clusters.")
```
fiunchinho, line 134: I believe these are sane defaults?

main.go
```diff
@@ -153,7 +153,6 @@ func mainError() error {
 	daemonCommand.PersistentFlags().Int(f.Service.Cluster.Kubernetes.IngressController.SecurePort, 0, "Secure port of guest cluster Ingress Controller.")
 	daemonCommand.PersistentFlags().String(f.Service.Cluster.Kubernetes.Kubectl.Docker.Image, "", "Full docker image of kubectl.")
 	daemonCommand.PersistentFlags().String(f.Service.Cluster.Kubernetes.Kubelet.AltNames, "", "Alternative names for guest cluster kubelet certificates.")
-	daemonCommand.PersistentFlags().String(f.Service.Cluster.Kubernetes.Kubelet.Labels, "", "Labels for guest cluster kubelets.")
```
fiunchinho, line 156: Instead of requiring a parameter to be passed for the kubelet labels, I added a function to the `key` package. This function mimics what `kubernetesd` is doing.

service/controller/machine_pool.go
```diff
@@ -14,44 +14,47 @@ import (
...
+	SSHUserList               string
+	SSOPublicKey              string
 }
 
 type MachinePool struct {
 	*controller.Controller
 }
 
 func NewMachinePool(config MachinePoolConfig) (*MachinePool, error) {
-	if config.Logger == nil {
```
fiunchinho, line 47: These are checked later.

service/controller/resource/spark/create.go
fiunchinho: Maybe I missed it @whites11 but I couldn't find anywhere in `k8scloudconfig`where these values are used.

service/controller/key/key.go
```diff
@@ -293,8 +293,14 @@ func IsSucceededProvisioningState(s string) bool {
...
 }
 
-func MachinePoolOperatorVersion(cr expcapzv1alpha3.AzureMachinePool) string {
-	return cr.GetLabels()[LabelOperatorVersion]
+// These are the same labels that kubernetesd adds when creating/updating an AzureConfig.
+func KubeletLabelsNodePool(getter LabelsGetter) string {
+	var labels string
+
+	labels = ensureLabel(labels, label.Provider, "azure")
+	labels = ensureLabel(labels, label.OperatorVersion, OperatorVersion(getter))
```
nprokopic, line 301: We should slowly start moving these labels to `apiextensions/pkg/label`.

nprokopic: Not a blocker.

service/controller/resource/spark/resource.go
```diff
@@ -14,33 +23,121 @@ import (
...
+		return nil, microerror.Maskf(invalidConfigError, "%T.APIServerSecurePort must not be empty", config)
+	}
+
+	err := config.Azure.Validate()
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if config.Calico.MTU == 0 {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Calico must not be empty", config)
```
nprokopic, line 78:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, "%T.Calico.MTU must not be empty", config)
```

service/controller/resource/spark/resource.go
```diff
@@ -14,33 +23,121 @@ import (
...
+	}
+
+	if config.CertsSearcher == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CertsSearcher must not be empty", config)
+	}
+
+	if config.ClusterIPRange == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.clusterIPRange must not be empty", config)
```
nprokopic, line 86:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, "%T.ClusterIPRange must not be empty", config)
```

service/controller/resource/spark/resource.go
```diff
@@ -14,33 +23,121 @@ import (
...
+	if config.CredentialProvider == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CredentialProvider must not be empty", config)
+	}
+
 	if config.CtrlClient == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.CtrlClient must not be empty", config)
 	}
+
+	if config.EtcdPrefix == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.etcdPrefix must not be empty", config)
```
nprokopic, line 98:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, "%T.EtcdPrefix must not be empty", config)
```

service/controller/resource/spark/resource.go
```diff
@@ -14,33 +23,121 @@ import (
...
+	}
+
+	if config.Ignition.Path == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Ignition must not be empty", config)
```
nprokopic, line 102:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, "%T.Ignition.Path must not be empty", config)
```

service/controller/resource/spark/resource.go
```diff
@@ -81,3 +178,58 @@ func (r *Resource) toEncrypterObject(ctx context.Context, secretName string) (en
 
 	return enc, nil
 }
+
+func (r *Resource) getCredentialSecret(ctx context.Context, azureMachinePool key.LabelsGetter) (*v1alpha1.CredentialSecret, error) {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding credential secret")
+
+	organization, exists := azureMachinePool.GetLabels()[label.Organization]
```
nprokopic, line 185: You could use `label.Organization` from `apiextensions` here.

service/controller/key/key.go
fiunchinho: Yeah, let's do it in a different PR though.

service/controller/resource/spark/resource.go
fiunchinho: I'll do a bulk change when we update apiextensions.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/Azure/go-autorest/autorest/to v0.3.0
 	github.com/coreos/go-semver v0.3.0
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32 // indirect
-	github.com/giantswarm/apiextensions v0.4.16-0.20200717100705-0940ca9b44ea
+	github.com/giantswarm/apiextensions v0.4.16-0.20200723091752-e5ce54ab1302
```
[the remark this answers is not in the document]
nprokopic: It's released.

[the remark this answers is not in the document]
fiunchinho: I need the contents of a branch though. [I need this](https://github.com/giantswarm/apiextensions/pull/527), which is not merged yet.

PR #910, opened 2020-07-22
Retry failed deployments

whites11:
This PR changes the `deployment` resource behaviour in 2 ways:

1) if the main deployment is in a failed state, the whole deployments get applied (even if there is no change in the checksums)
2) it the main deployment is in a failed state, the `initialProvisioning` parameter is set to Yes. This will make the `virtual_network_setup` template to be applied as well.

This is needed to support tenant cluster upgrades from 11.4.0 to 12.0.0.

@fiunchinho I kindly ask you to think of the implications of this change for node pools and, in particular, with the handling of subnets for node pools


whites11:
> If the Deployment is in failed state, it would remove the subnets created for the node pools. What if we remove the Deployment when it's in Failed state? That's the approach I took in the `subnet` handler https://github.com/giantswarm/azure-operator/pull/883/files#r453585162

I don't get the difference between the two.

fiunchinho:
> > If the Deployment is in failed state, it would remove the subnets created for the node pools. What if we remove the Deployment when it's in Failed state? That's the approach I took in the `subnet` handler https://github.com/giantswarm/azure-operator/pull/883/files#r453585162
> 
> I don't get the difference between the two.

Sheit, you are correct. It would do the same. I don't know an easy way out then.

PR #920, opened 2020-07-22
wait for nat gateway before creating the virtual network

whites11:
It can happen that the nat gateway is not created in time and the subnets gets created without a nat gateway.

This PR adds a dependency to avoid this problem


fiunchinho:
Why was this closed btw?

whites11:
It was a false alarm. It's working as is, no need for the change

PR #922, opened 2020-07-23
Fix setting azureClientsFactory in cloudconfig resource

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/12063


kopiczko:
How it wasn't detected for so long?

fiunchinho:
This handler is not being used yet. It was added for node pools, but we will most likely remove it due to taking a different direction.

PR #923, opened 2020-07-23
Persist security rules

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/8562
This PR aims at preserving the security rules between cluster upgrades.
What happened before this PR is that the azure operator always restored the security groups to their default values (the one in the ARM template) and thus deleting all custom rules (for example the ones created by kubernetes when we deploy a LoadBalancer).

KNOWN ISSUE: there is no way currently to delete a rule from the list of default ones (it will be kept forever unless deleted manually).
I guess we can live with this limitation and possibly handle it in the future with the azure vacuum cleaner.


fiunchinho:
image


:loudspeaker: Please add _"Towards..."_ and select the milestone (probably 4.2.0?) :loudspeaker:

whites11:
Done brosky

client/azure_client_set.go
```diff
@@ -209,7 +209,7 @@ func newPublicIPAddressesClient(authorizer autorest.Authorizer, subscriptionID,
 	return &client, nil
 }
 
-func newSecurityGroupsClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (*network.SecurityRulesClient, error) {
+func newSecurityRulesClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (interface{}, error) {
```
fiunchinho, line 212: Could we keep both `SecurityGroupsClient` and `SecurityRulesClient`?

whites11: It was a typo. The method was called `getSecurityGroupsClient` but was returning a `SecurityRulesClient`

service/controller/resource/deployment/securityrules.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	for iterator.NotDone() {
+		rule := iterator.Value()
+
+		// Check if rule is a default one by comparing the name.
+		isDefault := false
+		{
+			for _, candidate := range defaultRules {
+				if *rule.Name == *candidate.Name {
```
fiunchinho, line 44: What if _something_ modifies a default security rule? With this solution the modification would be lost, the default rule would be used instead. Not sure if this could ever happen.

fiunchinho:
Do we know which security rules will be added on top of the default ones (like the load balancer ones)? I was thinking if it would make sense to only keep those that we know we need, instead of keeping all existing security rules, you know what I mean? Defining in azure-operator those that we know we need to keep.

service/controller/resource/deployment/securityrules.go
whites11: That's true. The assumption here is that security rules are immutable. 
If you need to override a default security rule you need a new one with lower priority.
I guess we can't do much but accept this limitation

whites11:
> Do we know which security rules will be added on top of the default ones (like the load balancer ones)? I was thinking if it would make sense to only keep those that we know we need, instead of keeping all existing security rules, you know what I mean? Defining in azure-operator those that we know we need to keep.

Apart from the fact that it would be hard to detect them (the LB ones have randomly generated names) this would not let customers create custom rules. And this is an interesting feature to have IMHO (especially now that customers can vnet peer their networks)

fiunchinho:
> > Do we know which security rules will be added on top of the default ones (like the load balancer ones)? I was thinking if it would make sense to only keep those that we know we need, instead of keeping all existing security rules, you know what I mean? Defining in azure-operator those that we know we need to keep.
> 
> Apart from the fact that it would be hard to detect them (the LB ones have randomly generated names) this would not let customers create custom rules. And this is an interesting feature to have IMHO (especially now that customers can vnet peer their networks)

ok, I see. Cool!

whites11:
Tested, it works.

PR #925, opened 2020-07-27
Delete Spark CR when deleting a node pool

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12271

When we use the real Ignition Operator, we wouldn't have to worry about the `Secret` deletion from here. 

But should we care about the `Spark` CR deletion? or should the piece that created it in the first place take care of deleting it @giantswarm/team-ludacris? I mean in the future when we have the Ignition operator in place.

### DIFF
--- a/service/controller/resource/spark/create.go
+++ b/service/controller/resource/spark/create.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/Azure/go-autorest/autorest/to"
 	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
 	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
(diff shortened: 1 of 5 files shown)


service/controller/resource/spark/create.go
```diff
@@ -100,6 +100,14 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      secretName(sparkCR.Name),
 					Namespace: azureMachinePool.Namespace,
+					OwnerReferences: []metav1.OwnerReference{
```
fiunchinho, line 103: I set the OwnerReference field in the Secret created to hold the cloudconfig string, that way when the Spark CR is deleted, the Secret will be deleted by Kubernetes because it's an orphan.

Is this the right approach or should I explicitly delete the secret from here?

tfussell: In my view, these are effectively the same. We can delete dependent object in our own finalizer step or by leveraging built-in kubernetes garbage collection via `OwnerReferences`. It means less code for us so it seems preferable to use `OwnerReferences` when possible.

service/controller/resource/spark/create.go
```diff
@@ -100,6 +100,14 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+						{
+							APIVersion: sparkCR.APIVersion,
+							Kind:       sparkCR.Kind,
+							Name:       sparkCR.Name,
+							UID:        sparkCR.UID,
+						},
```
tfussell, line 109: You might want to use foreground deletion by setting `BlockOwnerDeletion` to `true` here. This way the `Spark` CR won't be deleted until the corresponding `Secret` is deleted if I understand correctly.

fiunchinho: Good idea. Done.

PR #926, opened 2020-07-27
Store bootstrap secret cloudconfig in Storage Account blob

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12271

After the spark handler creates the required cloudconfig and writes it to a `Secret`, we need to store that data in the Storage Account so VMSS virtual machines can fetch it.


service/controller/resource/cloudconfigblob/create.go
```diff
@@ -0,0 +1,143 @@
...
+		err := r.ctrlClient.Get(ctx, client.ObjectKey{Namespace: azureMachinePool.Namespace, Name: sparkCR.Status.DataSecretName}, &cloudconfigSecret)
+		if err != nil {
+			return "", microerror.Mask(err)
+		}
+	}
+
+	return cloudconfigSecret.String(), nil
+}
+
+func (r *Resource) getContainerURL(ctx context.Context, credentialSecret *v1alpha1.CredentialSecret, resourceGroupName, storageAccountName string) (azblob.ContainerURL, error) {
```
fiunchinho, line 97: This function is copy paste code from the `containerurl` handler that was saving the `containerURL` in the controller context.

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	blobName := key.BlobName(&azureMachinePool, key.PrefixWorker())
+
+	var payload string
+	{
+		payload, err = r.getCloudConfigFromBootstrapSecret(ctx, azureMachinePool)
+		if errors.IsNotFound(err) {
```
fiunchinho, line 38: If required objects are not created yet, or bootstrap CR is not ready, we cancel the current resource/handler, hoping another handler will fix that.

service/controller/machine_pool.go
```diff
@@ -159,8 +175,23 @@ func NewMachinePoolResourceSet(config MachinePoolConfig) ([]resource.Interface,
...
+
+		cloudconfigblobResource, err = cloudconfigblob.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
 	resources := []resource.Interface{
 		sparkResource,
+		cloudconfigblobResource,
```
fiunchinho, line 194: This new handler is executed right after the `sparkResource` tries to create the `Secret` with the right `cloudconfig`.

service/controller/resource/cloudconfigblob/create.go
whites11: Do you need to actually cancel the reconciliation of this resource or is returning just enough?

fiunchinho: We can:
- Return, which cancels this resource reconciliation, allowing the controller to continue to the next resource
- Set reconciliation canceled, which makes the controller to stop until next reconciliation loop, avoiding the controller to continue to the next resource

I believe in this case we want the first one. At least while we don't have any more resources in this controller.

fiunchinho: Nah, I think you are right. I'm going to cancel the whole reconciliation loop. We can't do anything if cloudconfig is not rendered.

service/controller/resource/spark/create.go
```diff
@@ -141,8 +140,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	{
-		if !bytes.Equal(ignitionBlob, dataSecret.Data[ignitionBlobKey]) {
-			dataSecret.Data[ignitionBlobKey] = ignitionBlob
+		if !bytes.Equal(ignitionBlob, dataSecret.Data[key.CloudConfigSecretKey]) {
+			dataSecret.Data[key.CloudConfigSecretKey] = ignitionBlob
```
fiunchinho, line 144: I moved this to the `key` package because we need it when writing but also when reading the value from the secret.

PR #927, opened 2020-07-27
Service Enpoint manual update

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/8562

This PR reverts the following modifications recently made:

- https://github.com/giantswarm/azure-operator/pull/923 (security rules)
- https://github.com/giantswarm/azure-operator/pull/911 (initial provisioning flag)

Because they broke the upgrade path from 11.4.0 to 12.0.0 and were causing a long downtime during cluster upgrades.

There is also a new check, needed just for cluster upgrades, that ensures that the `Master` subnet has the `Microsoft.Storage` service endpoint enabled before the upgrade starts.

I'm not happy **at all** with the outcome, but I really tried all options and no one was able to provide:

- a working upgrade path
- almost-zero downtime for customer workload

at the same time.


service/controller/resource/deployment/serviceendpoints.go
```diff
@@ -0,0 +1,57 @@
...
+	})
+
+	_, err = subnetsClient.CreateOrUpdate(ctx, key.ResourceGroupName(cr), key.VnetName(cr), key.MasterSubnetName(cr), subnet)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "info", "message", fmt.Sprintf("service '%s' was added to subnet %s", storageServiceEndpoint, *subnet.Name))
+
+	reconciliationcanceledcontext.SetCanceled(ctx)
```
fiunchinho, line 54: why do we have to cancel the reconciliation? Now that we ensured the endpoints, can't we just continue?

client/azure_client_set.go
```diff
@@ -209,7 +209,7 @@ func newPublicIPAddressesClient(authorizer autorest.Authorizer, subscriptionID,
 	return &client, nil
 }
 
-func newSecurityRulesClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (interface{}, error) {
+func newSecurityGroupsClient(authorizer autorest.Authorizer, subscriptionID, partnerID string) (*network.SecurityRulesClient, error) {
```
fiunchinho, line 212: This was a typo right? Let's keep the fix.

service/controller/resource/deployment/serviceendpoints.go
whites11: when the endpoint gets created, the virtual network is in an "updating" state for a while.
If we don't cancel the reconciliation, the subsequent deployment applications (it's the very next instruction so it's a quick thing) fails because azure is a liquid stinky shit.
So we have no alternative but wait one reconciliation loop if we change the service endpoints.

client/azure_client_set.go
whites11: done

PR #928, opened 2020-07-28
Added provider agnostic controllers

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12428

This PR adds two provider agnostic controllers: machine pool controller and cluster controller. These controllers try to mimic the upstream CAPI provider agnostic controllers.

By setting the `ownerReferences` in our CR's we can leverage Kubernetes garbage collection. This means that if a `Cluster` is deleted, owned resources like `AzureCluster` and `MachinePool` will be deleted (and also `AzureMachinePool` because it's owned by `MachinePool`).

At the moment these controllers only do this small task, so I _inlined_ the handlers inside the controllers. Not sure about this approach though. We could create two new handlers "the normal way" and just reference them from the controllers.

## machine pool controller

The machine pool controller makes sure that:
- `MachinePool` is owned by `Cluster`.
- `MachinePool` is labeled with `Cluster` id.
- `AzureMachinePool` is owned by `MachinePool`.

## cluster controller

The cluster controller makes sure that:
- `AzureCluster` is owned by `Cluster`.
- `AzureCluster` is labeled with `Cluster` id.

### DIFF
--- a/service/controller/azure_config.go
+++ b/service/controller/azure_config.go
@@ -1,0 +1,642 @@
+package controller
+
+import (
+	"context"
+	"net"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/certs"
+	"github.com/giantswarm/k8sclient/v3/pkg/k8sclient"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/controller"
+	"github.com/giantswarm/operatorkit/resource"
+	"github.com/giantswarm/operatorkit/resource/crud"
+	"github.com/giantswarm/operatorkit/resource/wrapper/metricsresource"
+	"github.com/giantswarm/operatorkit/resource/wrapper/retryresource"
+	"github.com/giantswarm/randomkeys"
+	"github.com/giantswarm/statusresource"
+	"github.com/giantswarm/tenantcluster"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/giantswarm/azure-operator/v4/client"
+	"github.com/giantswarm/azure-operator/v4/pkg/credential"
+	"github.com/giantswarm/azure-operator/v4/pkg/label"
+	"github.com/giantswarm/azure-operator/v4/pkg/locker"
+	"github.com/giantswarm/azure-operator/v4/pkg/project"
+	"github.com/giantswarm/azure-operator/v4/service/controller/cloudconfig"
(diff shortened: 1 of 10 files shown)


service/controller/azure_config.go
```diff
@@ -0,0 +1,642 @@
+package controller
```
fiunchinho, line 1: This isn't new. This is the controller that we already had to watch `AzureConfig` CRs, but renamed from `cluster.go` to `azure_config.go`.

service/controller/machine_pool.go
```diff
@@ -0,0 +1,226 @@
...
+	return resources, nil
+}
+
+type MachinePoolOwnerReferencesConfig struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+	Scheme     *runtime.Scheme
+}
+
+type MachinePoolOwnerReferencesResource struct {
```
fiunchinho, line 142: Inlined handler.

service/controller/cluster.go
```diff
@@ -2,81 +2,30 @@ package controller
...
 	"github.com/giantswarm/operatorkit/resource"
-	"github.com/giantswarm/operatorkit/resource/crud"
 	"github.com/giantswarm/operatorkit/resource/wrapper/metricsresource"
 	"github.com/giantswarm/operatorkit/resource/wrapper/retryresource"
-	"github.com/giantswarm/randomkeys"
-	"github.com/giantswarm/statusresource"
-	"github.com/giantswarm/tenantcluster"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
```
fiunchinho, line 15: This is a new controller, even though the diff shows it as modified. It's the provider agnostic cluster controller.

service/controller/cluster.go
```diff
@@ -191,452 +76,128 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
...
-			Logger:             config.Logger,
+type ClusterOwnerReferencesConfig struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+	Scheme     *runtime.Scheme
+}
 
-			NetworkRange: config.IPAMNetworkRange,
-		}
+type ClusterOwnerReferencesResource struct {
```
fiunchinho, line 138: Inlined handler.

service/unittest/default_k8sclient.go
```diff
@@ -30,6 +35,22 @@ func FakeK8sClient() k8sclient.Interface {
 		if err != nil {
 			panic(err)
 		}
+		err = expcapiv1alpha3.AddToScheme(scheme)
```
fiunchinho, line 38: Added CAPI/CAPZ types here so we can use it for testing.

fiunchinho:
Superseeded by https://github.com/giantswarm/azure-operator/pull/933

PR #941, opened 2020-08-06
use new certs searcher approach

xh3b4sd:
## Context

Coming from https://gigantic.slack.com/archives/C017CV2A274/p1596621081001800. Trying to help Team Batman with this.


service/controller/resource/blobobject/desired.go
```diff
@@ -25,11 +27,78 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	var clusterCerts certs.Cluster
+	var masterCertFiles []certs.File
+	var workerCertFiles []certs.File
```
xh3b4sd, line 31: We do this twice in different handlers. That could be refactored but I was lazy now. 😬

service/controller/resource/blobobject/desired.go
```diff
@@ -25,11 +27,78 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+			return nil
+		})
+
+		g.Go(func() error {
+			tls, err := r.certsSearcher.SearchTLS(key.ClusterID(&cr), certs.WorkerCert)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+			m.Lock()
+			workerCertFiles = append(workerCertFiles, certs.NewFilesWorker(tls)...)
```
xh3b4sd, line 91: Would be nice if someone could double check if the certs are all there.

service/controller/resource/cloudconfig/desired.go
```diff
@@ -51,11 +53,84 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	var clusterCerts certs.Cluster
+	// TODO
+	//
+	//     * no dependabot
+	//     * ambiguous deps
+	//     *
+	//
```
xh3b4sd, line 61: Just my WIP comments on what I noticed. I can drop them but we may want to do something about it. Let's get dependabot setup here. `devctl` can help with that. And then we should take a look at all the different versions we use. At best we use one version per dependency and not three.

rossf7: Agree IMO dependabot would help but I've removed this comment for now.

@tuommaki @nprokopic PTAL and track in Celestial if there is stuff to do here.

service/controller/resource/blobobject/desired.go
tuommaki: Since this is `blobobject`, this will most probably go away at some point. All good :+1:

service/controller/resource/blobobject/desired.go
tuommaki: We'll get to it in few days in general. We are at the point of stitching together different components right now in order to jumpstart actual node pool cluster anytime soon.

rossf7:
Tested on `godsmack`.

PR #942, opened 2020-08-10
Rename integration to e2e

tuommaki:
Current E2E tests are in directory called `integration` even though they are
not really integration tests but more like e2e tests (as also called in many
context).

In order to not confuse e2e tests with actual integration tests, separate these
two into corresponding directories.


tuommaki:
I just renamed `integration` dir to `e2e`. E2e tests green after the change. Going with a ping.

PR #943, opened 2020-08-10
Add boilerplate structure & wiring for capzcrs handler

tuommaki:
The added handler skeleton will be mapping AzureConfig to CAPI & CAPZ types.
This is only introducing the handler structure so it's just a NOP handler.


tuommaki:
This is just an empty handler. The meat follows up later. CI green, I'm going ->

PR #944, opened 2020-08-10
Add initial structure for CR mapping integration test

tuommaki:
This change provides initial structure to introduce CR mapping integration test
for AzureConfig to CAPI CRs functionality.

### DIFF
--- a/integration/test/crmapping/crmapping_test.go
+++ b/integration/test/crmapping/crmapping_test.go
@@ -1,0 +1,132 @@
+package crmapping
+
+import (
+	"context"
+	"io/ioutil"
+	"path"
+	"strconv"
+	"testing"
+
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/giantswarm/operatorkit/resource"
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/capzcrs"
+)
+
+func TestAzureConfigCRMapping(t *testing.T) {
+	testCases := []struct {
+		name         string
+		inputFiles   []string
+		errorMatcher func(error) bool
+	}{
+		{
+			name: "case 0: Simple AzureConfig",
+			inputFiles: []string{
+				"simple_azureconfig.yaml",
+			},
+			errorMatcher: nil,
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			t.Log(tc.name)
+
+			client := newFakeClient()
+			ensureCRsExist(t, client, tc.inputFiles)
+
+			r := constructResource(t, client)
+
+			err := r.EnsureCreated(context.Background(), nil)
+
+			// TODO: Read expected CRs w/ client and compare to "golden" versions.
+
+			switch {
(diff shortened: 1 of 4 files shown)


integration/test/crmapping/crmapping_test.go
```diff
@@ -0,0 +1,132 @@
...
+			t.Log(tc.name)
+
+			client := newFakeClient()
+			ensureCRsExist(t, client, tc.inputFiles)
+
+			r := constructResource(t, client)
+
+			err := r.EnsureCreated(context.Background(), nil)
+
+			// TODO: Read expected CRs w/ client and compare to "golden" versions.
```
tuommaki, line 48: In next PR I start the implementation of `capzcrs` functionality. Basic logic is required there in order to being able to then implement the gold file structure here. I'll provide both in the same PR then.

My idea here is to have set of input files that are "loaded" into fake client (i.e. the baseline before executing the handler) and then run the mapping once and then compare the resulting state with what is stored in "golden files".

This way it's easy to make thorough test coverage for CR fields while keeping maintenance work low when changing the mapping. One should then also see nicely the concrete diff of changes in end result CRs when changing the mapping logic.

integration/test/crmapping/crmapping_test.go
```diff
@@ -0,0 +1,132 @@
...
+	m := make(map[string]interface{})
+	err = yaml.Unmarshal(bs, m)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	switch m["kind"] {
+	case "AzureConfig":
+		obj = &providerv1alpha1.AzureConfig{}
+	default:
```
tuommaki, line 115: I'll extend this switch later with relevant CAPI & CAPZ types.

integration/test/crmapping/testdata/simple_azureconfig.yaml
```diff
@@ -0,0 +1,201 @@
...
+        port: 10250
+      networkSetup:
+        docker:
+          image: quay.io/giantswarm/k8s-setup-network-environment:1f4ffc52095ac368847ce3428ea99b257003d9b9
+        kubeProxy:
+          conntrackMaxPerCore: 0
+      ssh:
+        userList:
+        - name: foobar
+          publicKey: ssh-rsa foobarbaz== foobar
```
tuommaki, line 93: This CR is from real live cluster but I redacted SSH keys out (even though they are public keys).

integration/test/crmapping/crmapping_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"io/ioutil"
+	"path"
+	"strconv"
+	"testing"
+
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/giantswarm/operatorkit/resource"
+	"gopkg.in/yaml.v2"
```
axbarsan, line 14: I've noticed that in most upstream projects they use `sigs.k8s.io/yaml` instead of `gopkg.in/yaml.v2`. Can we run into some edge cases where the behaviour would be different?

tuommaki: I'm not aware of any. So far we have been always using `gopkg.in/yaml.v2` or `ghodss/yaml`. My quick grep over ~50 common repos found from my machine yields only few hits that are using k8s one and all rest using either one of aforementioned alternatives.

For consistency purposes I figure it would be a good thing to settle on one of these and change rest of them, but until anything breaks, it doesn't probably matter that much which one is used. While YAML itself is pretty complicated format with anchors & all, the used subset in k8s context is most of the time very simple so I don't expect to hit corner cases anytime soon. :slightly_smiling_face:

tuommaki: Ok. I ended up having some struggles with YAML parsing in the end so among other changes I switched to k8s parser. Thanks for note! :+1:

PR #945, opened 2020-08-11
Implement CR mapping from AzureConfig to CAPI & CAPZ

tuommaki:
This change implements CR mapping from AzureConfig to CAPI & CAPZ types.

Towards https://github.com/giantswarm/giantswarm/issues/12694


tuommaki:
**NOTE:** Proper test cases are still missing and I haven't given this a smoke run on real cluster yet, but here is a draft to introduce the structure of this code. The design here can be reviewed already.

integration/test/crmapping/crmapping_test.go
```diff
@@ -22,15 +29,23 @@ import (
...
 )
 
-func TestAzureConfigCRMapping(t *testing.T) {
+var update = flag.Bool("update", false, "update .golden reference files")
+
+// Test_AzureConfigCRMapping uses golden files.
+//
+//  go test ./integration/test/crmapping -run Test_AzureConfigCRMapping -update
+//
+func Test_AzureConfigCRMapping(t *testing.T) {
```
tuommaki, line 38: These are actually unit tests at this point, but when getting this complete, my plan is to merge functionality of `azureconfig` and `capzcrs` resources to test cases such that they map each others after initial state and then the final state is verified.

tuommaki: Bidirectional tests are now here. There's a lot of common functionality between these unit tests and bidirectional integration tests. What's your opinion @nprokopic & @axbarsan - should I move the unit tests out from here and move to `capzcrs` package or are they ok here?

service/controller/resource/capzcrs/create.go
```diff
@@ -2,8 +2,294 @@ package capzcrs
...
+
+		// Acquire accessors for ObjectMeta and TypeMeta fields of CR.
+		desiredMeta, err := meta.Accessor(desired)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		desiredType, err := meta.TypeAccessor(desired)
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
axbarsan, line 213: Nice 👍

integration/test/crmapping/crmapping_test.go
nprokopic: No hard opinion for me, however you find it appropriate.

Once we remove all `AzureConfig` stuff, we will I guess remove also mapping tests?

tuommaki: Yep, the mapping is gone then.

tuommaki: done

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "4.2.1-dev"
+	version            = "5.0.0-dev"
```
tuommaki, line 8: I think we can now bump the major version since this PR will start migrating the `AzureConfig` into CAPI & CAPZ types so it would then activate the CAPI & CAPZ controllers as well and therefore start working towards :mage:

PR #946, opened 2020-08-12
IPAM implementation fixes for node pools

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/9

### What's in the box?

- `AzureMachinePoolSubnetCollector`: collects all allocated node pool networks
- `AzureMachinePoolNetworkRangeGetter`: gets tenant cluster VNet from which the subnet will be allocated, and gets node pool subnet mask
- `AzureConfigNetworkRangeGetter`: gets installation network range from which the tenant cluster VNet will be allocated, and gets tenant cluster VNet mask
- Helper functions for getting CRs
- Small fixes and refactorings here and there

### Why new `Collector` for node pools?

Existing subnet collector (it should really be VNet collector) was collecting all allocated VNets across the installation, it is used when new VNet is allocated for the new tenant cluster. In this case we need to know which available range (out of entire installation range) we can use for the new VNet, so we check which VNets are already allocated, and then we can get a free VNet range.

New node pool needs a new subnet that should be allocated from the tenant cluster VNet range. Therefore, we need to know which subnets are already allocated (`AzureMachinePoolSubnetCollector`), so we can get a free subnet range from the tenant cluster VNet with the required node pool subnet mask (`AzureMachinePoolNetworkRangeGetter`).

### What is this new `NetworkRangeGetter` interface?

```
// NetworkRangeGetter implementation returns a network range from which a free
// IP range can be allocated.
type NetworkRangeGetter interface {
	// GetNetworkRange gets the network range from which the VNet/subnet range
	// will be allocated. It receives the CR that is being reconciled.
	GetNetworkRange(ctx context.Context, obj interface{}) (net.IPNet, error)

	// GetRequiredIPMask returns an IP mask that is required by the network range
	// that will be allocated.
	GetRequiredIPMask() net.IPMask
}
```

This way we keep `ipam` implementation generic, and every specific `ipam` implementation gets to decide:
- Total network range from which it can allocate a subrange for new VNet/subnet. These are:
    - when reconciling `AzureConfig`: entire installation network range, used when new VNet is allocated for the tenant cluster. Parameter in `azure-operator`.
    - when reconciling `AzureMachinePool`: tenant cluster network range obtained from `AzureCluster` CR (`Cluster.Spec.ClusterNetwork.Services.CIDRBlocks` or `AzureCluster.Spec.NetworkSpec.Vnet.CidrBlock`, checking this), used when new subnet is allocated for the node pool.
- What is required mask for the newly required VNet/subnet network range
    - when reconciling `AzureConfig`: defaults to `/16`, can be specified as `azure-operator` parameter
    - when reconciling `AzureMachinePool`: defaults to `/24`, can be specified in `AzureCluster.Spec.NetworkSpec` (not sure if we accept custom values there?)

P.S. Maybe `NetworkRangeGetter` is not the best name, shout if you have some suggestions how to name it better.


service/controller/resource/ipam/spec.go
```diff
@@ -15,7 +15,19 @@ type Checker interface {
 // Collector implementation must return all networks that are allocated on any
 // given moment. Failing to do that will result in overlapping allocations.
 type Collector interface {
-	Collect(ctx context.Context) ([]net.IPNet, error)
+	Collect(ctx context.Context, obj interface{}) ([]net.IPNet, error)
```
nprokopic, line 18: Added `obj interface{}`, since we need to pass reconciled CR for node pools implementation. Existing `AzureConfig` implementation will just ignore this additional parameter.

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
```diff
@@ -0,0 +1,81 @@
...
+		err = microerror.Maskf(invalidObjectError, "%T.Services must not be empty", cluster.Spec.ClusterNetwork)
+		return net.IPNet{}, microerror.Mask(err)
+	}
+
+	if len(cluster.Spec.ClusterNetwork.Services.CIDRBlocks) == 0 {
+		err = microerror.Maskf(invalidObjectError, "%T.CIDRBlocks must not be empty", cluster.Spec.ClusterNetwork.Services)
+		return net.IPNet{}, err
+	}
+
+	cidrBlock := cluster.Spec.ClusterNetwork.Services.CIDRBlocks[0] // or should we use AzureCluster.Spec.NetworkSpec.Vnet.CidrBlock here?
```
nprokopic, line 68: Not sure why I used this `Cluster.Spec.ClusterNetwork.Services.CIDRBlocks` here (checking it), more probably `AzureCluster.Spec.NetworkSpec.Vnet.CidrBlock` is the correct place to get the VNet CIDR from.

pkg/helpers/api.go
```diff
@@ -0,0 +1,101 @@
...
+	// If one credential secret is found, we use that.
+	if len(secretList.Items) == 1 {
+		secret := secretList.Items[0]
+
+		credentialSecret := &v1alpha1.CredentialSecret{
+			Namespace: secret.Namespace,
+			Name:      secret.Name,
+		}
+
+		// r.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found credential secret %s/%s", credentialSecret.Namespace, credentialSecret.Name))
```
fiunchinho, line 87: logging is commented out

service/controller/azure_config.go
```diff
@@ -502,17 +502,28 @@ func newAzureConfigResources(config AzureConfigConfig, certsSearcher certs.Inter
...
-			Checker:   azureConfigChecker,
-			Collector: subnetCollector,
-			Locker:    config.Locker,
-			Logger:    config.Logger,
-			Persister: azureConfigPersister,
-
-			AllocatedSubnetMaskBits: config.GuestSubnetMaskBits,
-			NetworkRange:            config.IPAMNetworkRange,
+			Checker:            azureConfigChecker,
+			Collector:          subnetCollector,
```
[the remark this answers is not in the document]
fiunchinho: It's not correct, so I'd change it.

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
```diff
@@ -0,0 +1,64 @@
...
+	"net"
+
+	"github.com/giantswarm/microerror"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-operator/v4/pkg/helpers"
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
+)
+
+var nodePoolIPMask = net.CIDRMask(24, 32)
```
fiunchinho, line 14: any reason to have this variable here instead of just returning the value from the `GetRequiredIPMask` function?

pkg/helpers/api.go
nprokopic: Oh yes,  I forgot to pass in logger here. Will fix.

service/controller/azure_config.go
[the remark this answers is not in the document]
nprokopic: I agree.

@tuommaki did you mention last week something about consistency with `aws-operator` or some other IPAM implementation, or something like that?

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
nprokopic: Not any strong reason, just for the sake of not creating it every time as it's always the same.

Would it be better to do `return net.CIDRMask(24, 32)` inside the function? More idiomatic maybe?

service/controller/azure_config.go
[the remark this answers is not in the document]
fiunchinho: Originally, our `ipam` handler [was copied from aws-operator](https://github.com/giantswarm/aws-operator/tree/master/service/controller/resource/ipam). If we now change it, both `ipam` handlers will diverge. Not sure if it's a problem. When aws-operator moves to CAPI they can copy our implementation like we did?

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
fiunchinho: I'm fine either way, I was just curious :smile:

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
```diff
@@ -0,0 +1,64 @@
...
+		return net.IPNet{}, microerror.Mask(err)
+	}
+
+	// Get AzureCluster CR where the NetworkSpec is stored.
+	azureCluster, err := helpers.GetAzureClusterFromMetadata(ctx, g.client, azureMachinePool.ObjectMeta)
+	if err != nil {
+		return net.IPNet{}, microerror.Mask(err)
+	}
+
+	_, ipNet, err := net.ParseCIDR(azureCluster.Spec.NetworkSpec.Vnet.CidrBlock)
```
tuommaki, line 52: Should we have condition to gracefully handle situation where `azureCluster.Spec.NetworkSpec.Vnet.CidrBlock` is not yet set? This would happen e.g. when creating cluster with default node pool (default behavior in `gsctl`).

service/controller/resource/ipam/azure_machinepool_subnet_collector.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	// Get AzureCluster CR where the subnets are stored.
+	azureCluster, err := helpers.GetAzureClusterFromMetadata(ctx, c.client, azureMachinePool.ObjectMeta)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
```
tuommaki, line 74: Bail out if `azureCluster.Spec.NetworkSpec.Vnet.CidrBlock` is still empty.

service/controller/resource/ipam/azure_machinepool_subnet_collector.go
```diff
@@ -0,0 +1,190 @@
...
+	// cluster ID is tenant cluster resource group name
+	resourceGroupName := azureCluster.Name
+
+	// Not assuming VNet name here, keeping it flexible. In order to keep it correct and stable, we
+	// should have a webhook for enforcing a VNet name convention. This check can be removed once
+	// we have implemented a webhook.
```
tuommaki, line 159:
```suggestion
	// should have a webhook for enforcing a VNet name convention.
```
I don't think it hurts to have this here even if webhook validates it as well?

service/controller/azure_config.go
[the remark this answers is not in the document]
tuommaki: > @tuommaki did you mention last week something about consistency with aws-operator or some other IPAM implementation, or something like that?

Yeah, but given how this is laid out here, I'm happy with the current design. All good :+1:

> Originally, our ipam handler was copied from aws-operator. If we now change it, both ipam handlers will diverge. Not sure if it's a problem. When aws-operator moves to CAPI they can copy our implementation like we did?

Yeah, but there's also the thing that AWS networking is quite different from the Azure one and atm. I don't think it makes sense to try to make them look 1:1 similar.

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
fiunchinho: I'm never fully sure on how to handle those async cases. For example, 5 lines above, the code will fail if the cluster label is not set, but we have controllers that ensure the label is set. Should handle that case more gracefully?

pkg/helpers/api.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err := c.Get(ctx, key, azureCluster); err != nil {
+		return nil, err
+	}
+
+	return azureCluster, nil
+}
+
+func GetCredentialSecretFromMetadata(ctx context.Context, c client.Client, obj metav1.ObjectMeta) (*v1alpha1.CredentialSecret, error) {
+	// r.Logger.LogCtx(ctx, "level", "debug", "message", "finding credential secret")
```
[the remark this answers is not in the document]
nprokopic: Passed in logger so we can actually log :)

pkg/helpers/api.go
nprokopic: Added logging.

pkg/helpers/api.go
```diff
@@ -0,0 +1,101 @@
...
+		return credentialSecret, nil
+	}
+
+	// If no credential secrets are found, we use the default.
+	credentialSecret := &v1alpha1.CredentialSecret{
+		Namespace: credentialNamespace,
+		Name:      credentialDefaultName,
+	}
+
+	// r.Logger.LogCtx(ctx, "level", "debug", "message", "did not find credential secret, using default secret")
```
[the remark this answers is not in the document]
nprokopic: Added logger :)

service/controller/azure_config.go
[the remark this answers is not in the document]
nprokopic: Thanks for the feedback guys :) I'll rename this then.

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
nprokopic: Gracefully handling missing (still not set) `Cluster` fields/labels/annotations would be done by returning `nil` from `ipam`'s `EnsureCreated`, right?

service/controller/resource/ipam/azure_machinepool_subnet_collector.go
nprokopic: It doesn't hurt, but I believe that if the webhook is validating the name and setting a default value, the name here will be always valid.

On the other side, if we do the defaulting in operator code, we need the check here as well :)

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
tuommaki: > I'm never fully sure about how to best handle those async scenarios. For example, 5 lines above, the code will fail if the cluster label is not set, but we have controllers that ensure the label is set. Should handle that case more gracefully?

On principal we should never error on these kind of issues where it will become eventually consistent (after 1-3 reconciliation rounds of different controllers), because otherwise it's counted in monitoring for alerts and that's not what we want. We know that it's ok that initially there are blank fields that might not be ready for full blown cluster creation and we should just defer the execution where possible.

> Gracefully handling missing (still not set) Cluster fields/labels/annotations would be done by returning nil from ipam's EnsureCreated, right?

Yeah :+1:

nprokopic: Bailing out done.

service/controller/resource/ipam/azure_machinepool_subnet_collector.go
nprokopic: Done.

service/controller/resource/ipam/virtualnetwork_collector.go
```diff
@@ -198,7 +198,7 @@ func (c *SubnetCollector) getSubnetsFromSubscription(ctx context.Context, client
 
 		// Search a VNET with any of the expected names.
 		vnetCandidates := []string{
-			fmt.Sprintf("%s-VirtualNetwork", *group.Name),
+			fmt.Sprintf("%s-VirtualNetwork", *group.Name), // TODO: get VNet name from AzureCluster CR
```
tuommaki, line 201: **TODO?**

tuommaki (approved):
Besides one TODO comment, this LGTM now. :+1:

service/controller/resource/ipam/create.go
```diff
@@ -66,7 +66,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "finding allocated subnets")
 
 		allocatedSubnets, err = r.collector.Collect(ctx, obj)
-		if err != nil {
+		if IsNetworkRangeStillNotKnown(err) {
```
fiunchinho, line 69: I don't want to block this PR, but wouldn't make sense to move the comment to the `IsNetworkRangeStillNotKnown()` function so that we don't have to duplicate it?

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
```diff
@@ -38,6 +46,11 @@ func NewAzureMachinePoolNetworkRangeGetter(config AzureMachinePoolNetworkRangeGe
 // node pool subnet is getting its IP address range from all available address
 // ranges in the tenant cluster virtual network.
 func (g *AzureMachinePoolNetworkRangeGetter) GetNetworkRange(ctx context.Context, obj interface{}) (net.IPNet, error) {
+	g.logger.LogCtx(
+		ctx,
+		"level", "debug",
+		"message", "getting tenant cluster's VNet range from which the node pool subnet will be allocated")
+
```
fiunchinho, line 53: If we are adding a new error to gracefully handle not having a `CidrBlock` just yet, I think we should do the same for the case when the cluster label is missing and `GetAzureClusterFromMetadata()` returns an error.
We can do it in another PR if you want.

nprokopic: :thinking: not sure if those two cases are the same
- `CidrBlock` is missing because it's still not allocated, when a new cluster is being created that happens by design
- cluster label, when it's missing, means that it is not set, while it has to be set according to the spec - IMHO this should be solved by proper defaulting, so we don't have to check everywhere in code if something that should be set is set or not, since a missing label is an error in the CR

I think should have a chat about validation and defaulting :)

WDYT?

service/controller/resource/ipam/virtualnetwork_collector.go
nprokopic: Replaced with a more clear note :)

service/controller/resource/ipam/create.go
nprokopic: Updated comments a bit, so that:
- comments in `EnsureCreated` are about how the errors affects the reconciliation
- comments in `error.go` are about the error itself

P.S. The doubling was on purpose, so you see what's happening more quickly/easily, but now it's clear as well I believe :)

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
fiunchinho: We have a controller that adds the label as well, so eventually it will be there. If we want to make it mandatory to have the label, then errorring out is fine I guess.

fiunchinho (approved):

service/controller/resource/ipam/azure_machinepool_network_range_getter.go
nprokopic: I believe we decided that it's mandatory to have the label per spec.

PR #947, opened 2020-08-14
Extract OwnerReference handlers to own packages

tuommaki:
Cluster & MachinePool controllers need to be configured with more handlers so
it makes sense to extract embedded handlers to their own packages as others.

There are no logic changes, the implementation is just moved to separate packages.


tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/12727

PR #948, opened 2020-08-14
Add CAPI handlers to wait for dependent CRs on deletion

tuommaki:
When deleting cluster, Cluster controller must wait for all cluster related
child CRs before the finalizer is removed. Similar to this, MachinePool
controller must wait for all node pool related CRs to be deleted before
allowing the finalizer to be removed.

This prevents orphaned dependents when CRs are deleted in unspecified order.

### DIFF
--- a/service/controller/resource/clusterdependents/resource.go
+++ b/service/controller/resource/clusterdependents/resource.go
@@ -12,7 +12,6 @@ import (
 	expcapiv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/giantswarm/azure-operator/v4/pkg/label"
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
@@ -107,7 +106,7 @@ func (r *Resource) infrastructureCRExists(ctx context.Context, cr capiv1alpha3.C
 
 func (r *Resource) machinePoolCRsExist(ctx context.Context, cr capiv1alpha3.Cluster) (bool, error) {
 	o := client.MatchingLabels{
-		label.Cluster: key.ClusterID(&cr),
+		capiv1alpha3.ClusterLabelName: key.ClusterID(&cr),
 	}
 	mpList := new(expcapiv1alpha3.MachinePoolList)
 	err := r.ctrlClient.List(ctx, mpList, o)
--- a/service/controller/resource/clusterdependents/resource.go
+++ b/service/controller/resource/clusterdependents/resource.go
@@ -90,8 +90,12 @@ func (r *Resource) Name() string {
 }
 
 func (r *Resource) infrastructureCRExists(ctx context.Context, cr capiv1alpha3.Cluster) (bool, error) {
+	objKey := client.ObjectKey{
+		Namespace: cr.Namespace,
+		Name: cr.Spec.InfrastructureRef.Name,
+	}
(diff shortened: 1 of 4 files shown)


tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/12727

Wiring of these handlers will follow later in separate PR.

service/controller/resource/clusterdependents/resource.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Name returns the resource name.
+func (r *Resource) Name() string {
+	return Name
+}
+
+func (r *Resource) infrastructureCRExists(ctx context.Context, cr capiv1alpha3.Cluster) (bool, error) {
+	azureCluster := new(v1alpha3.AzureCluster)
+	err := r.ctrlClient.Get(ctx, client.ObjectKey{Namespace: cr.Namespace, Name: cr.Spec.InfrastructureRef.Name}, azureCluster)
```
axbarsan, line 94: To keep it consistent with the `machinePoolCRsExist` function below

```suggestion
	objKey := client.ObjectKey{
		Namespace: cr.Namespace,
		Name: cr.Spec.InfrastructureRef.Name,
	}
	azureCluster := new(v1alpha3.AzureCluster)
	err := r.ctrlClient.Get(ctx, objKey, azureCluster)
```

service/controller/resource/clusterdependents/resource.go
```diff
@@ -0,0 +1,115 @@
...
+
+	exists, err = r.machinePoolCRsExist(ctx, cr)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if exists {
+		finalizerskeptcontext.SetKept(ctx)
+		return nil
+	}
```
axbarsan, line 82: So I could understand better:
If someone tries to delete the `Cluster` resource before deleting the `AzureCluster` and `MachinePool` resources, the deletion will not be finished because we're forcefully keeping the azure-operator finalizer in place. Once the `AzureCluster` and all `MachinePool` resources that belong to the cluster have been deleted, the finalizer will also be removed, so deleting the `Cluster` resource will now finish and the world will be great again.

Is this correct?

service/controller/resource/clusterdependents/resource.go
```diff
@@ -0,0 +1,115 @@
...
+	} else if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
+func (r *Resource) machinePoolCRsExist(ctx context.Context, cr capiv1alpha3.Cluster) (bool, error) {
+	o := client.MatchingLabels{
+		label.Cluster: key.ClusterID(&cr),
```
nprokopic, line 106: Should we use GiantSwarm `giantswarm.io/cluster` or Cluster API `cluster.x-k8s.io/cluster-name` label for these queries?
I believe we are currently mixing them here and there, so it would be good to decide and use the same one across the codebase.

tuommaki: I was just independently looking at the same problem. As discussed, I feel that it would be good to rely on `cluster.x-k8s.io` one and that's documented in the Node Pools Sticky Notes issue as well. I'll fix it in this PR now and create a separate issue to fix `api` & other places.

service/controller/resource/clusterdependents/resource.go
tuommaki: Yes. Then to dive deeper, similar handler in `MachinePool` controller ensures that `AzureMachinePool` CRs are gone before it can be deleted.

PR #949, opened 2020-08-17
Wire CAPI dependents deletion handlers

tuommaki:
CAPI dependents deletion handlers wait until all dependent CRs are gone before
removing finalizer from Cluster and MachinePool.


tuommaki:
I somehow forgot this. Here we go. Towards https://github.com/giantswarm/giantswarm/issues/12727

PR #950, opened 2020-08-17
Ensure Cluster OwnerReference on MachinePool & AzureMachinePool

tuommaki:
~I'll sort out the unit test once we figure if we want all these set.~ Done.

### DIFF
--- a/service/controller/resource/clusterownerreference/resource.go
+++ b/service/controller/resource/clusterownerreference/resource.go
@@ -8,7 +8,9 @@ import (
 	"github.com/giantswarm/micrologger"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	expcapzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
 	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	expcapiv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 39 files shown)


service/controller/resource/clusterownerreference/resource.go
```diff
@@ -86,16 +118,79 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	for _, azureMachinePool := range mpList.Items {
+		r.logger.LogCtx(ctx, "message", fmt.Sprintf("Ensuring %s label and 'ownerReference' fields on AzureMachinePool '%s/%s'", capiv1alpha3.ClusterLabelName, azureMachinePool.Namespace, azureMachinePool.Name))
+
+		if azureMachinePool.Labels == nil {
+			azureMachinePool.Labels = make(map[string]string)
+		}
+		azureMachinePool.Labels[capiv1alpha3.ClusterLabelName] = cluster.Name
+
+		// Set Cluster as owner of AzureMachinePool
+		err = controllerutil.SetControllerReference(&cluster, &azureMachinePool, r.scheme)
```
fiunchinho, line 182: From the comments of the `SetControllerReference` function

> SetControllerReference sets owner as a Controller OwnerReference on controlled.
This is used for garbage collection of the controlled object and for
reconciling the owner object on changes to controlled (with a Watch + EnqueueRequestForOwner).
Since only one OwnerReference can be a controller, it returns an error if
there is another OwnerReference with Controller flag set.

IIUC this will fail because the `machinepoolownerreference` handler already sets the `MachinePool` as owner of `AzureMachinePool`, and `AzureMachinePool` can't contain two `OwnerReference` that are controllers.

Instead we should use the other method to add owner references without being controllers, like we do to set `Cluster` as owner of `MachinePool`:

```
	machinePool.OwnerReferences = capiutil.EnsureOwnerRef(machinePool.OwnerReferences, metav1.OwnerReference{
		APIVersion: capiv1alpha3.GroupVersion.String(),
		Kind:       "Cluster",
		Name:       cluster.Name,
		UID:        cluster.UID,
	})
```

fiunchinho:
I asked [this question](https://kubernetes.slack.com/archives/C8TSNPY4T/p1595864697323600) a couple of weeks ago

image

They [linked to this docs](https://cluster-api.sigs.k8s.io/developer/crd-relationships.html)

image

IIUC, `Cluster` doesn't own `AzureMachinePool` in upstream (or `AzureMachine`, or `AzureMachineTemplate`, etc).

That would mean that we'd need to replace this

```
	cluster, err := r.getOwnerCluster(ctx, azureMachinePool.ObjectMeta)
	if err != nil {
	}
```

with a different mechanism to find the related `Cluster` CR, probably using the cluster label. Looking at [the kubeadm bootstrap controller](https://github.com/kubernetes-sigs/cluster-api/blob/43dd623ce667a1ff33e0d2d649b9a3dea2977f02/bootstrap/kubeadm/controllers/kubeadmconfig_controller.go#L149-L165), it seems that they find the `MachinePool` using `ownerReference`, and then use the `clusterName` field from it to fetch the `Cluster` object from the k8s API.

I don't see anything wrong having `Cluster` owning `AzureMachinePool`, but if we want to be as close to upstream as possible, then we'd have to follow that path.

tuommaki:
Fishing for LGTM

PR #952, opened 2020-08-19
Handle missing storage account & blob gracefully on create & delete

tuommaki:
When cluster gets deleted, the top level cluster reconciliation deletes whole
resource group. If machinepool cleanup is still in progress, it might happen
that storage account gets deleted before the machine pool is fully gone so
handle the not found error gracefully.

Equally when it's created, the storage account might not be there yet so
cancel resource while waiting for it.

### DIFF
--- a/service/controller/resource/cloudconfigblob/create.go
+++ b/service/controller/resource/cloudconfigblob/create.go
@@ -64,7 +64,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	var containerURL azblob.ContainerURL
 	{
 		containerURL, err = r.getContainerURL(ctx, credentialSecret, key.ClusterID(&azureMachinePool), key.StorageAccountName(&azureMachinePool))
-		if err != nil {
+		if IsStorageAccountNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "did not find storage account")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		} else if err != nil {
 			return microerror.Mask(err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -153,11 +157,7 @@ func (r *Resource) getPrimaryKey(ctx context.Context, storageAccountsClient *sto
 	r.logger.LogCtx(ctx, "level", "debug", "message", "Finding PrimaryKey for encryption in Storage Account")
 
 	_, err := storageAccountsClient.GetProperties(ctx, resourceGroupName, storageAccountName, storage.AccountExpandGeoReplicationStats)
-	if IsStorageAccountNotFound(err) {
```
tuommaki, line 156: I think this was accidentally in wrong place.

service/controller/resource/cloudconfigblob/delete.go
```diff
@@ -41,7 +41,10 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 	var containerURL azblob.ContainerURL
 	{
 		containerURL, err = r.getContainerURL(ctx, credentialSecret, key.ClusterID(&azureMachinePool), key.StorageAccountName(&azureMachinePool))
-		if err != nil {
+		if IsStorageAccountNotFound(err) {
```
tuommaki, line 44: This must be handled on creation as well because this can easily happen when creating cluster with defaults (so nodepool gets created together with cluster).

service/controller/resource/cloudconfigblob/delete.go
```diff
@@ -41,14 +41,17 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
+			return nil
+		} else if err != nil {
 			return microerror.Mask(err)
 		}
 	}
 
 	blob := containerURL.NewBlockBlobURL(blobName)
 	_, err = blob.Delete(ctx, azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
-	if blobclient.IsNotFound(err) {
+	if blobclient.IsNotFound(err) || blobclient.IsBlobNotFound(err) {
```
fiunchinho, line 54: what's the difference between the two? :smile:

tuommaki: https://github.com/giantswarm/azure-operator/blob/master/service/controller/blobclient/error.go#L23-L53 dont' ask why the return status code doesn't match with `404`. Azure? :man_shrugging:

fiunchinho (approved):

tuommaki:
Logs seem clean

PR #953, opened 2020-08-20
Handle missing infrastructure errors gracefully in machinepool reconciliation

tuommaki:
When creating or deleting cluster, machinepool reconciliation might be either                                                                                                                                                                                                                                                 
too early or too late when creating / deleting corresponding resources. In the                                                                                                                                                                                                                                                
beginning the support infrastructure resources, such as storage account, may                                                                                                                                                                                                                                                  
not exist yet or later on cluster deletion they might be already gone with the                                                                                                                                                                                                                                                
resource group deletion. These errors must be gracefully handled so that they                                                                                                                                                                                                                                                 
don't emit false alerts to monitoring infrastructure.


whites11 (approved):
LGTM when tested

PR #954, opened 2020-08-20
Mapping AzureConfig to first node pool CRs

nprokopic:


tuommaki:
Don't do this in `capzcrs` but in separate handler that's built for migration (or modified `instance` if that's for migration).

If default node pool is created in `capzcrs`, then the end user can't get rid of it until we move away from `AzureConfig`. It should be in the migration handler that only executes through once (having trigger with old VMSS; when it doesn't exist anymore, the handler doesn't execute anymore).

nprokopic:
> If default node pool is created in `capzcrs`, then the end user can't get rid of it until we move away from `AzureConfig`.

We could set `AzureConfig` `Workers` to `null` after mapping, but using actual old VMSS as a trigger sounds much more reliable.

PR #956, opened 2020-08-24
Cancel reconciliation if nat gateway is not ready yet

fiunchinho:
Subnets are linked to a nat gateway so that workers can access the internet. When creating the subnets, this nat gateway needs to be already provisioned and in a `Succeeded` state.

With these changes, we cancel the subnet reconciliation if the nat gateway is not in the required state.


service/controller/resource/subnet/resource.go
```diff
@@ -121,12 +130,21 @@ func getSubnetARMDeploymentName(subnetName string) string {
...
 }
 
-func (r *Resource) ensureSubnets(ctx context.Context, deploymentsClient *azureresource.DeploymentsClient, storageAccountsClient *storage.AccountsClient, azureCluster capzv1alpha3.AzureCluster) error {
+func (r *Resource) ensureSubnets(ctx context.Context, deploymentsClient *azureresource.DeploymentsClient, storageAccountsClient *storage.AccountsClient, natGatewaysClient *network.NatGatewaysClient, azureCluster capzv1alpha3.AzureCluster) error {
 	armTemplate, err := subnet.GetARMTemplate()
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
+	natGw, err := natGatewaysClient.Get(ctx, key.ClusterID(&azureCluster), "workers-nat-gw", "")
```
nprokopic, line 139: Maybe we should have this stored in a CR as an annotation? :)

nprokopic (approved):
LGTM

PR #960, opened 2020-08-25
Gracefully handle k8s API returning conflicts when saving objects concurrently

fiunchinho:


service/controller/resource/nodepool/create.go
```diff
@@ -62,16 +63,15 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
-	err = r.CtrlClient.Status().Update(ctx, &azureMachinePool)
```
fiunchinho, line 65: This was unnecessary. The variable `azureMachinePool` won't contain local changes made in the transition functions anyway.

tuommaki (approved):
Makes sense. Eventually consistent.

PR #961, opened 2020-08-26
Improve capzcrs CR merging

tuommaki:
Generic implementation doesn't preserve original CR spec fields. This is needed
to maintain dynamically managed subnets in AzureCluster.

### DIFF
--- a/service/controller/resource/capzcrs/create.go
+++ b/service/controller/resource/capzcrs/create.go
@@ -24,6 +24,13 @@ import (
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
 
+type crmapping struct {
+	obj runtime.Object
+
+	needUpdateFunc func(orig, desired runtime.Object) (bool, error)
+	mergeFunc      func(orig, desired runtime.Object) (runtime.Object, error)
+}
+
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	cr, err := key.ToCustomResource(obj)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


tuommaki:
When testing on `ghost` the subnets are now preserved.

fiunchinho (approved):
LGTM

service/controller/resource/capzcrs/create_test.go
```diff
@@ -114,6 +125,76 @@ func Test_AzureConfigCRMapping(t *testing.T) {
...
+						Subnets: []*capzv1alpha3.SubnetSpec{
+							&capzv1alpha3.SubnetSpec{
+								ID:   "nodepool-np201",
+								Role: capzv1alpha3.SubnetNode,
+							},
+						},
+					},
+				},
+			},
+			desired: &capzv1alpha3.AzureCluster{},
```
nprokopic, line 157: Would it be useful to have a test case where there are, let's say, 2 desired and 3 original? An example where there is a diff, but both are non-null?

service/controller/resource/capzcrs/create.go
```diff
@@ -290,6 +294,109 @@ func (r *Resource) updateCRs(ctx context.Context, desiredCRs []runtime.Object) e
...
+	}
+
+	if !cmp.Equal(d, o, cmpopts.IgnoreTypes(&metav1.ObjectMeta{}), cmpopts.IgnoreTypes(&metav1.TypeMeta{}, cmpopts.IgnoreTypes(&capzv1alpha3.Subnets{}))) {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func mergeAzureCluster(orig, desired runtime.Object) (runtime.Object, error) {
```
nprokopic, line 316: The func will always take subnets from original and set them to desired, right? There isn't an use case where subnets from both original and desired will be taken?

service/controller/resource/capzcrs/create_test.go
tuommaki: Mapping from `AzureConfig` -> `AzureCluster` doesn't recognize subnets at all. Therefore desired would always have nil for subnets. So: No, it wouldn't be useful :slightly_smiling_face:

service/controller/resource/capzcrs/create.go
tuommaki: Nope. More details above :slightly_smiling_face:

service/controller/resource/capzcrs/create_test.go
nprokopic: Thanks!

nprokopic (approved):

PR #963, opened 2020-08-26
azure v5: use CAPI cluster name label instead of GS one

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12772


axbarsan:
Closing because the branch name makes this undeployable

PR #964, opened 2020-08-26
azure v5: use CAPI cluster name label instead of GS one

axbarsan:
Towards giantswarm/giantswarm#12772

Adding the CAPI cluster.x-k8s.io/cluster-name label for storing cluster IDs wherever the giantswarm.io/cluster label is used. All queries by cluster ID are now done using the CAPI label.
I still kept the giantswarm.io/cluster label everywhere, for compatibility with AWS.

Note: I have no idea what I'm doing, so hopefully I haven't missed anything


fiunchinho (approved):
LGTM, remember to clean the `project.go` "version" field

tuommaki (approved):
Looks good! :rocket:

PR #965, opened 2020-08-26
Pass node pool ID to test so it can assert later

fiunchinho:
`MultiAZ` test needs the node pool id, so that it can assert it has been deployed in the right AZs.


tuommaki (approved):
:+1:

PR #967, opened 2020-08-26
Customize e2e test cluster ID

nprokopic:
New env vars:

- `CLUSTER_ID_PREFIX`: by default it's `ci` (current naming scheme)
- `RANDOMIZE_CLUSTER_ID`: by default it's `false`
  - `false`: cluster ID = `cluster ID prefix` + `Tested version` (by default `wip`) + `commit SHA` + `Test name hash`
  - `true`: cluster ID = `cluster ID prefix` + `random 5 chars` + `Tested version` (by default `wip`)

### DIFF
--- a/e2e/entityid/generator.go
+++ b/e2e/entityid/generator.go
@@ -1,0 +1,48 @@
+package entityid
+
+import (
+	"math/rand"
+	"regexp"
+	"strconv"
+	"time"
+)
+
+const (
+	// idChars represents the character set used to generate cluster IDs.
+	// (does not contain 1 and l, to avoid confusion)
+	idChars = "023456789abcdefghijkmnopqrstuvwxyz"
+
+	// idLength represents the number of characters used to create a cluster ID.
+	idLength = 5
+)
+
+var (
+	// Use local instance of RNG. Can be overwritten with fixed seed in tests
+	// if needed.
+	localRng = rand.New(rand.NewSource(time.Now().UnixNano()))
+)
+
+func New() string {
+	pattern := regexp.MustCompile("^[a-z]+$")
+	for {
+		letterRunes := []rune(idChars)
+		b := make([]rune, idLength)
+		for i := range b {
+			b[i] = letterRunes[localRng.Intn(len(letterRunes))]
+		}
+
+		id := string(b)
+
+		if _, err := strconv.Atoi(id); err == nil {
+			// string is numbers only, which we want to avoid
+			continue
+		}
+
+		if pattern.MatchString(id) {
+			// strings is letters only, which we also avoid
+			continue
+		}
+
+		return id
+	}
+}
--- a/e2e/entityid/generator.go
+++ b/e2e/entityid/generator.go
@@ -22,7 +22,7 @@ var (
 	localRng = rand.New(rand.NewSource(time.Now().UnixNano()))
 )
 
-func New() string {
+func Generate() string {
 	pattern := regexp.MustCompile("^[a-z]+$")
 	for {
 		letterRunes := []rune(idChars)
--- a/e2e/env/common.go
+++ b/e2e/env/common.go
@@ -91,7 +91,10 @@ func init() {
 	}
 
 	if randomizeClusterID && (len(clusterIDPrefix)+len(testedVersion) > MaxClusterIDPrefixAndTestedVersionLength) {
-		panic(fmt.Sprintf("Max length for cluster ID prefix (%s) + tested version (%s) version, when combined, is %d due to Azure resource name limitations (e.g. storage account name)", clusterIDPrefix, testedVersion, MaxClusterIDPrefixAndTestedVersionLength))
+		panic(fmt.Sprintf(
+			"Max length for cluster ID prefix (%s) + tested version (%s) version, when combined, "+
+				"is %d due to Azure resource name limitations (e.g. storage account name). Try passing a shorter prefix in '%s'.",
+			clusterIDPrefix, testedVersion, MaxClusterIDPrefixAndTestedVersionLength, EnvVarClusterIDPrefix))
 	}
 
 	testDir = os.Getenv(EnvVarTestDir)
--- a/e2e/env/common.go
+++ b/e2e/env/common.go
@@ -3,59 +3,75 @@ package env
 import (
 	"crypto/sha256"
 	"fmt"
-	"math/rand"
 	"os"
-	"regexp"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/giantswarm/versionbundle"
 
+	"github.com/giantswarm/azure-operator/v4/e2e/entityid"
 	"github.com/giantswarm/azure-operator/v4/pkg/project"
 )
 
 const (
-	DefaultTestedVersion = "wip"
+	DefaultClusterIDPrefix    = "ci"
+	DefaultRandomizeClusterID = false
+	DefaultTestedVersion      = "wip"
 
 	EnvVarCircleSHA               = "CIRCLE_SHA1" // #nosec
+	EnvVarClusterIDPrefix         = "CLUSTER_ID_PREFIX"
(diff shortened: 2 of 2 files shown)


e2e/entityid/generator.go
```diff
@@ -0,0 +1,48 @@
...
+	idLength = 5
+)
+
+var (
+	// Use local instance of RNG. Can be overwritten with fixed seed in tests
+	// if needed.
+	localRng = rand.New(rand.NewSource(time.Now().UnixNano()))
+)
+
+func New() string {
```
fiunchinho, line 25: New sounds to me like I'm building an object. I'd prefer `NewID`.

e2e/env/common.go
```diff
@@ -53,6 +77,22 @@ func init() {
...
+		clusterIDPrefix = DefaultClusterIDPrefix
+		fmt.Printf("No value found in '%s': using default value %s\n", EnvVarClusterIDPrefix, DefaultClusterIDPrefix)
+	}
+
+	if randomizeClusterID {
+		randomPrefixPart := entityid.New()
+		clusterIDPrefix = fmt.Sprintf("%s-%s", clusterIDPrefix, randomPrefixPart)
+	}
+
+	if randomizeClusterID && (len(clusterIDPrefix)+len(testedVersion) > MaxClusterIDPrefixAndTestedVersionLength) {
```
fiunchinho, line 92: It would be good to tell what the user can do to fix this, something like 
```
panic(fmt.Sprintf("Max length for cluster ID prefix (%s) + tested version (%s) version, when combined, is %d due to Azure resource name limitations (e.g. storage account name). Try passing a shorter prefix in '%s'", clusterIDPrefix, testedVersion, MaxClusterIDPrefixAndTestedVersionLength), EnvVarClusterIDPrefix)
```

e2e/env/common.go
```diff
@@ -53,6 +77,22 @@ func init() {
 		fmt.Printf("No value found in '%s': using default value %s\n", EnvVarTestedVersion, DefaultTestedVersion)
 	}
 
+	clusterIDPrefix = os.Getenv(EnvVarClusterIDPrefix)
```
fiunchinho, line 80: I was thinking if it would be better if we:
- added new `clusterid` variable
- brought the logic inside the `ClusterID()` function here instead
- made the `ClusterID()` function to just return the `clusterid` variable

I'm suggesting it because it feels weird to me to have different places to manage basically the same thing.

fiunchinho (commented):
I'm not completely sold on the random it. What's the use case for it? I think there is value in having a deterministic name instead of random one, so that two runs of the same job yield the same results and outcomes.

tuommaki:
> cluster ID = cluster ID prefix + Tested version (by default wip) + commit SHA + Test name hash

I'd imagine this hits several different identifier length limits. Starting from CR name max length on k8s, but possibly also several things in the cloud such as VMSS name.

e2e/env/common.go
nprokopic: I was thinking the same at some moment, but then I just went with the flow :) I'll do this.

e2e/entityid/generator.go
nprokopic: In that case you would invoke it with `entityid.NewID()`, which AFAIU is against golang philosophy of not repeating parts of the name :)

Does `entity.NewID()` sound better maybe?

nprokopic:
> I'm not completely sold on the random it. What's the use case for it?

Even with this change, the default e2e test behaviour stays the same (deterministic).

In the context of running e2e tests locally: if you have same cluster ID for every test job, then multiple test jobs will try to reuse the same Azure resources, so only the first test job is "clean", while all test jobs after the first one are "compromised", as they will try to use same Azure resources (we never use existing Azure resource when creating a new cluster).

In that case you would have to delete Azure resources after every test run, and waiting for Azure to delete a tenant cluster resource group with everything we deploy can be very slow, so running the e2e tests locally could be even slower than CircleCI now.

> I think there is value in having a deterministic name instead of random one, so that two runs of the same job yield the same results and outcomes.

IMHO the test behaviour is still deterministic, the tests still yield the same results, just another instance of those results, where Azure resources have different names, but the resources that get deployed should be identical every time.
If we somehow get different results/resources for multiple test jobs, then our tests (or some parts of `azure-operator` logic) are not deterministic and we have an issue somewhere.

The idea here is that you can locally set `RANDOMIZE_CLUSTER_ID` to `true`, and then do dev + test iterations, where for test you just run the same command every time, and you can be sure that your test job got an environment with clean slate.

If you prefer deterministic cluster ID, you leave it to `false` (which is a default behaviour, so you don't have to set anything).

> > cluster ID = cluster ID prefix + Tested version (by default wip) + commit SHA + Test name hash
> 
> I'd imagine this hits several different identifier length limits. Starting from CR name max length on k8s, but possibly also several things in the cloud such as VMSS name.

This is the current behaviour, which I just kept here as a default :smile: e.g. `ci` + `wip` + `h9c82` + `98wdf` (with `-` as a glue).

If you want random names, you would now get this: `ci` (or whatever you set for `CLUSTER_ID_PREFIX`) + `random 5 chars` + `wip` (test version), with `-` as a glue, so you get a bit more space for custom `CLUSTER_ID_PREFIX`.

We could though make the generation of test cluster ID somehow simpler. Any suggestions?

nprokopic:
Related to determinism and isolation in tests, a quote from [Martin Flower's article _Eradicating Non-Determinism in Tests_](https://martinfowler.com/articles/nonDeterminism.html#LackOfIsolation):

> In order to get tests to run reliably, you must have clear control over the environment in which they run, so you have a well-known state at the beginning of the test. **If one test creates some data in the database and leaves it lying around, it can corrupt the run of another test which may rely on a different database state.**
>
> **Therefore I find it's really important to focus on keeping tests isolated.** Properly isolated tests can be run in any sequence. As you get to larger operational scope of functional tests, it gets progressively harder to keep tests isolated. When you are tracking down a non-determinism, lack of isolation is a common and frustrating cause.
>
> **There are a couple of ways to get isolation - either always rebuild your starting state from scratch, or ensure that each test cleans up properly after itself.** In general I prefer the former, as it's often easier - and in particular easier to find the source of a problem. If a test fails because it didn't build up the initial state properly, then it's easy to see which test contains the bug. With clean-up, however, one test will contain the bug, but another test will fail - so it's hard to find the real problem.
>
> Starting from a blank state is usually easy with unit tests, but can be much harder with functional tests - particularly if you have a lot of data in a database that needs to be there. Rebuilding the database each time can add a lot of time to test runs, so that argues for switching to a clean-up strategy.

e2e/entityid/generator.go
fiunchinho: `entityid.Generate()`?

fiunchinho:
okok, fine by me :smile_cat:

tuommaki (approved):
> This is the current behaviour, which I just kept here as a default smile e.g. ci + wip + h9c82 + 98wdf (with - as a glue).

Ah, right. I thought the hashes where the full length ones. Fine then.

I'm good with this. No strong opinions to any direction. Sort the conversation with Jose and proceed :+1:

e2e/env/common.go
nprokopic: Done.

e2e/env/common.go
```diff
@@ -91,10 +98,49 @@ func init() {
 			versionBundleVersion = vbs[len(vbs)-2].Version
 		}
 	}
+
 	err := os.Setenv(EnvVarVersionBundleVersion, VersionBundleVersion())
 	if err != nil {
 		panic(err)
 	}
+
+	clusterID := os.Getenv("CLUSTER_NAME")
```
nprokopic, line 107: This was here from the beginning :see_no_evil:  but we just recalculated the ID every time in the ClusterID func. Now it's also possible to set entire cluster ID with `CLUSTER_NAME` env var.

PR #968, opened 2020-08-26
Cancel nodepool handler if Subnet is not ready yet

fiunchinho:
We sometimes hit an error when creating the nodepool ARM deployment because the subnet chosen for the VMSS is still being created. It eventually succeeds but in order to avoid errors as much as possible, I'm canceling the nodepool handler if we don't have the `AzureCluster.Spec.NetworkSpec.subnet.ID` field set for the nodepool subnet. This `subnet.ID` field will be set by the subnet handler when it finds the subnet ARM deployment in a successful state.

### DIFF
--- a/service/controller/resource/nodepool/create_deployment_uninitialized.go
+++ b/service/controller/resource/nodepool/create_deployment_uninitialized.go
@@ -73,6 +73,10 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	if IsNotFound(err) {
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "Azure resource not found, canceling resource")
 		return currentState, nil
+	} else if IsSubnetNotReadyError(err) {
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "subnet is not Ready, it's probably still being created")
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return currentState, nil
 	} else if err != nil {
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 		return currentState, microerror.Mask(err)
--- a/service/controller/resource/nodepool/deployment.go
+++ b/service/controller/resource/nodepool/deployment.go
@@ -104,6 +104,10 @@ func (r Resource) newDeployment(ctx context.Context, storageAccountsClient *stor
 func (r Resource) getSubnetName(azureMachinePool *capzexpv1alpha3.AzureMachinePool, azureCluster *capzv1alpha3.AzureCluster) (string, string, error) {
 	for _, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
 		if azureMachinePool.Name == subnet.Name {
+			if subnet.ID == "" {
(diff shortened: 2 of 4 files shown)


service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -73,6 +73,10 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	if IsNotFound(err) {
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "Azure resource not found, canceling resource")
 		return currentState, nil
+	} else if IsSubnetNotReadyError(err) {
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "subnet is not Ready, it's probably still being created")
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return currentState, nil
```
fiunchinho, line 79: We don't return error since it's an expected case, but we cancel the reconciliation.

service/controller/resource/subnet/resource.go
```diff
@@ -233,6 +236,16 @@ func (r *Resource) ensureSubnets(ctx context.Context, deploymentsClient *azurere
 		}
 	}
 
+	// Update AzureCluster so that subnet.ID is saved.
```
fiunchinho, line 239: We save the `AzureCluster` object back to the k8s API after finish iterating all the subnets. In case we hit a conflict we log the error but don't return it. On the next iteration the handler will try to save it again.

service/controller/resource/nodepool/deployment.go
```diff
@@ -104,6 +104,10 @@ func (r Resource) newDeployment(ctx context.Context, storageAccountsClient *stor
 func (r Resource) getSubnetName(azureMachinePool *capzexpv1alpha3.AzureMachinePool, azureCluster *capzv1alpha3.AzureCluster) (string, string, error) {
 	for _, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
 		if azureMachinePool.Name == subnet.Name {
+			if subnet.ID == "" {
+				return azureCluster.Spec.NetworkSpec.Vnet.Name, subnet.Name, microerror.Maskf(subnetNotReadyError, fmt.Sprintf("Subnet %#q ID field is empty, which means the Subnet is not Ready", subnet.Name))
```
tuommaki, line 108: When returning an error we should return empty values for other fields. Is there something in the logic where we need the vnet and subnet names with the error?

service/controller/resource/subnet/resource.go
```diff
@@ -215,6 +216,8 @@ func (r *Resource) ensureSubnets(ctx context.Context, deploymentsClient *azurere
 				return microerror.Mask(err)
 			}
 
+			allocatedSubnet.ID = subnetID
```
tuommaki, line 219: Did you test if this works? I'd imagine `allocatedSubnet` to be copied instance of the element from `azureCluster` subnets. If that's the case, then the ID update here won't get effective when updating the `azureCluster` object. I'd maybe change the for loop to `for i := 0; i < len(azureCluster.Spec.NetworkSpec.Subnets); i++` form and then you could change the corresponding line here to:
`azureCluster.Spec.NetworkSpec.Subnets[i].ID = subnetID`

service/controller/resource/nodepool/deployment.go
fiunchinho: Nope. I'll change it.

fiunchinho:
Please, take a look again.

tuommaki (approved):
Awesome! :rocket:

PR #969, opened 2020-08-27
Http health check

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/12949

This PR changes the protocol used for k8s API load balancer health checks from TCP to HTTP.


calvix (approved):
nice

PR #970, opened 2020-08-27
Delete AzureMachine CRs when deleting master instances

tuommaki:
For now, when master instances are managed with AzureConfig reconciliation,
deletion of AzureMachines is most natural when deleting the instances as well.

This will change once the tenant cluster control plane is also reconciled by
CAPI&CAPZ types.


whites11 (approved):

tuommaki:
Tested. It works.

PR #973, opened 2020-08-27
Add monitoring label

omegas27:


helm/azure-operator/templates/service.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: {{ include "resource.default.namespace"  . }}
   labels:
     {{- include "labels.common" . | nindent 4 }}
+    giantswarm.io/monitoring: "true"
```
fiunchinho, line 8: What do you think about adding it as [a _common_ label here](https://github.com/giantswarm/azure-operator/blob/bb6a7330bbe7bf28bd799218b658252d632468f5/helm/azure-operator/templates/_helpers.tpl#L19)?

omegas27: If we do, the label will be everywhere when we only need it on the service so we can scrape it using the new prometheus setup

fiunchinho: Oh you are right. Never mind me then :smiley_cat:

omegas27: Well, now you know :dog: 
If we actually add it everywhere, we might scrape the endpoints multiple times

mig4 (approved):

PR #976, opened 2020-08-27
Make scaling e2e test provider CAPI compatible

fiunchinho:
The e2e tests that were shared between different providers were made in an abstract way, allowing the different providers to provide a set of functions to fulfill a `provider` interface so that the same test could be ran in different providers.

For the scaling test, we had our azure provider that worked with `AzureConfig` by adding and removing workers. We need to make it work with CAPI/CAPZ types now.

### DIFF
--- a/e2e/setup/config.go
+++ b/e2e/setup/config.go
@@ -2,6 +2,8 @@ package setup
 
 import (
 	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
 	"github.com/giantswarm/apprclient"
 	"github.com/giantswarm/e2e-harness/pkg/framework"
 	"github.com/giantswarm/e2e-harness/pkg/release"
@@ -12,7 +14,9 @@ import (
 	"github.com/giantswarm/k8sclient/v3/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	expcapzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
+	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	expcapiv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 
 	"github.com/giantswarm/azure-operator/v4/e2e/env"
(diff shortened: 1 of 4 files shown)


whites11 (approved):
LGTMIIWFY

e2e/setup/config.go
```diff
@@ -84,6 +88,10 @@ func NewConfig() (Config, error) {
 			Logger: logger,
 			SchemeBuilder: k8sclient.SchemeBuilder{
 				corev1alpha1.AddToScheme,
+				providerv1alpha1.AddToScheme,
```
fiunchinho, line 91: I don't think they were needed just yet, but I didn't want to hit this problem again in the future where some scheme is missing. That's why I'm adding them all now.

e2e/test/scaling/provider.go
```diff
@@ -94,30 +116,25 @@ func (p *Provider) NumMasters() (int, error) {
 }
 
 func (p *Provider) NumWorkers() (int, error) {
-	customObject, err := p.hostFramework.G8sClient().ProviderV1alpha1().AzureConfigs("default").Get(p.clusterID, metav1.GetOptions{})
+	ctx := context.TODO()
```
fiunchinho, line 119: I would add the context to the function signature but then I'd have to change e2e tests repositories, make sure legacy aws tests works, [and...](https://i.kym-cdn.com/photos/images/newsfeed/000/284/529/e65.gif)

e2e/test/scaling/provider.go
```diff
@@ -39,42 +44,56 @@ func NewProvider(config ProviderConfig) (*Provider, error) {
...
+		p.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("MachinePool CR for cluster id %q not found", p.clusterID))
+		return &v1alpha3.MachinePool{}, microerror.Maskf(notFoundError, fmt.Sprintf("MachinePool CR for cluster id %q not found", p.clusterID))
 	}
 
-	b, err := json.Marshal(patches)
+	return &crs.Items[0], nil
+}
+
+func (p *Provider) AddWorker() error {
+	ctx := context.TODO()
```
tuommaki, line 88:
```suggestion
	ctx := context.Background()
```
https://github.com/giantswarm/fmt/blob/master/go/context.md#guidelines

e2e/test/scaling/provider.go
```diff
@@ -94,30 +113,25 @@ func (p *Provider) NumMasters() (int, error) {
 }
 
 func (p *Provider) NumWorkers() (int, error) {
-	customObject, err := p.hostFramework.G8sClient().ProviderV1alpha1().AzureConfigs("default").Get(p.clusterID, metav1.GetOptions{})
+	ctx := context.TODO()
```
tuommaki, line 116:
```suggestion
	ctx := context.Background()
```
https://github.com/giantswarm/fmt/blob/master/go/context.md#guidelines

e2e/test/scaling/provider.go
```diff
@@ -94,30 +113,25 @@ func (p *Provider) NumMasters() (int, error) {
...
 func (p *Provider) RemoveWorker() error {
-	patches := []Patch{
-		{
-			Op:   "remove",
-			Path: "/spec/azure/workers/1",
-		},
-	}
-
-	b, err := json.Marshal(patches)
+	ctx := context.TODO()
```
tuommaki, line 126:
```suggestion
	ctx := context.Background()
```
Background

tuommaki (approved):
Nitpick for todo contexts. Otherwise lgtm.

PR #978, opened 2020-08-28
Make cluster deletion e2e test CAPI compatible

fiunchinho:


e2e/setup/bastion.go
```diff
@@ -77,9 +77,8 @@ func WaitForNetworkToBeCreated(ctx context.Context, logger micrologger.Logger, r
 
 		return nil
 	}
-	n := backoff.NewNotifier(logger, ctx)
 	b := backoff.NewConstant(backoff.LongMaxWait, backoff.LongMaxInterval)
-	err := backoff.RetryNotify(o, b, n)
+	err := backoff.Retry(o, b)
```
fiunchinho, line 81: it's an expected error while we wait, we don't want to log an error I think.

whites11 (approved):
seems reasonable

tuommaki (approved):
LGTM

PR #979, opened 2020-08-28
Little improvements

fiunchinho:


axbarsan (approved):
Nice

service/controller/resource/clusterdependents/resource.go
```diff
@@ -122,6 +122,11 @@ func (r *Resource) ensureMachinePoolCRsDeleted(ctx context.Context, cr capiv1alp
 	}
 
 	for _, mp := range mpList.Items {
+		if !mp.GetDeletionTimestamp().IsZero() {
```
fiunchinho, line 125: I saw it in CAPI code, I don't think it hurts?

whites11 (approved):

service/controller/resource/clusterdependents/resource.go
nprokopic: Sounds sane to me :)

service/controller/resource/ipam/resource.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	Persister          Persister
 }
 
+// Resource finds free IP ranges within a virtual network to create new subnets.
```
nprokopic, line 28: > finds free IP ranges within a virtual network to create new subnets.

This is for node pools. For `AzureConfig` (clusters) it would be something like: "finds free IP ranges within an installation range to create new virtual network for the tenant cluster."

nprokopic (approved):
One note regarding IPAM, otherwise all good.

service/controller/resource/ipam/resource.go
fiunchinho: Take a look now

PR #980, opened 2020-08-28
Update CAPI dependencies

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/13022

It looks like I'm updating a lot of dependencies, but compared to the node pools PR I'm actually just changing:

- cluster-api updated to v0.3.8
- cluster-api-provider-azure to v0.4.7
- removing `github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.4.0+incompatible`

After updating the dependencies some of the golden files need to be updated due fields being different in this CAPI version.


whites11 (approved):

tuommaki (approved):

nprokopic (approved):
LGTM

PR #981, opened 2020-08-31
Added azure clients

whites11:
This PR migrates azure's DNSRecordsClient and interfaceClient to the new factory system


nprokopic (approved):
LGTM

PR #982, opened 2020-08-31
Improve multi-az e2e test

nprokopic:
When checking tenant cluster availability zones, the test will use input provided by `AZURE_AZS` environment variable to check the number of availability zones and exact availability zones, instead of comparing to values hard-coded in the test.

This way custom test runs can use different multi-az settings.

### DIFF
--- a/e2e/env/azure.go
+++ b/e2e/env/azure.go
@@ -165,6 +165,21 @@ func AzureAvailabilityZonesAsStrings() []string {
 	return azs
 }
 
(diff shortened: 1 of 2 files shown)


whites11 (approved):
LGTMIIWFY

nprokopic:
Tested with:
- 1 zone: `AZURE_AZS="1"`
- 2 zones: `AZURE_AZS="1 3"`
- 3 zones: `AZURE_AZS="1 2 3"`

fiunchinho:
Nice!

PR #983, opened 2020-08-31
Add nat gateway for master instances

whites11:
This PR adds a new nat gateway to be used for master(s) instances.

This is currently not strictly needed, but it will be when we'll introduce HA masters.
Doing it now allows for easier testing and smaller PR when we'll merge HA masters.


axbarsan (approved):
Light years ahead, damn 😮

PR #985, opened 2020-09-01
Etcd public LB

whites11:
This PR changes the load balancer we use for the ETCD cluster from internal to public.
This is mandatory for HA masters, because otherwise 33% of the ETCD requests originating from the master nodes would fail (martian source error).
I am sorry the PR is quite large, but it touches a lot of stuff around the operator.

Tested both new cluster creation as well as 12.x upgrade.


service/controller/key/key_test.go
```diff
@@ -269,7 +269,7 @@ func Test_Functions_for_DNSKeys(t *testing.T) {
 }
 
 func Test_MasterNICName(t *testing.T) {
-	expectedMasterNICName := "3p5j2-Master-1-NIC"
+	expectedMasterNICName := "3p5j2-master-3p5j2-nic"
```
axbarsan, line 272: This is different compared to the source code

https://github.com/giantswarm/azure-operator/blob/4337e788945831cbcb691baf47abe64c050fd164/service/controller/key/key.go#L390

```suggestion
	expectedMasterNICName := "3p5j2-Master-3p5j2-nic"
```

axbarsan (approved):
Other than the name difference, everything looks good to me 👍

service/controller/key/key_test.go
whites11: good catch

service/controller/key/key.go
```diff
@@ -387,7 +387,7 @@ func MasterNatGatewayID(cr providerv1alpha1.AzureConfig, subscriptionID string)
 
 // MasterNICName returns name of the master NIC.
 func MasterNICName(customObject providerv1alpha1.AzureConfig) string {
-	return fmt.Sprintf("%s-Master-1-NIC", ClusterID(&customObject))
+	return fmt.Sprintf("%s-master-%s-nic", ClusterID(&customObject), ClusterID(&customObject))
```
fiunchinho, line 390: if we change the name, is it backwards compatible? can customers upgrade?

whites11: we didn't change the name. The function was unused. I needed it so I fixed it (the name returned by the function didn't match the actual name in the ARM template) and used it

PR #988, opened 2020-09-02
Skip CR mapping while cluster is being deleted

nprokopic:
Cluster deletion is slow, and we can have a race condition where CAPI CRs get recreated during cluster deletion.

We should skip the CR mapping during cluster creation.

### DIFF
--- a/service/controller/resource/capzcrs/create.go
+++ b/service/controller/resource/capzcrs/create.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/giantswarm/azure-operator/v4/service/controller/key"
 )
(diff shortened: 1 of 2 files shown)


service/controller/resource/capzcrs/create.go
```diff
@@ -38,6 +39,23 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+			Namespace: cr.Namespace,
+			Name:      cr.Name,
+		}
+		cluster := new(capiv1alpha3.Cluster)
+		err = r.ctrlClient.Get(ctx, objKey, cluster)
+		if errors.IsNotFound(err) {
+			// all good
+		} else if err != nil {
+			return microerror.Mask(err)
+		} else if cluster.GetDeletionTimestamp() != nil {
```
whites11, line 53: I think we usually use the IsZero() function on timestamps

service/controller/resource/capzcrs/create.go
```diff
@@ -40,6 +41,23 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
```
fiunchinho, line 55:
```suggestion
		} else if !cluster.GetDeletionTimestamp().IsZero() {
```

service/controller/resource/capzcrs/create.go
nprokopic: Already fixed by Jose. Go team!

fiunchinho (approved):

PR #989, opened 2020-09-02
Skip nodepool IPAM while the cluster is being deleted

nprokopic:
While the cluster is being deleted, there is no need to allocate a subnet for node pool. It's an edge case scenario that could in theory happen if a new node pool is added at roughly the same time while the cluster is being deleted.

### DIFF
--- a/service/controller/resource/ipam/azure_machinepool_subnet_checker.go
+++ b/service/controller/resource/ipam/azure_machinepool_subnet_checker.go
@@ -60,6 +60,11 @@ func (c *AzureMachinePoolSubnetChecker) Check(ctx context.Context, namespace str
 		if err != nil {
 			return false, microerror.Mask(err)
 		}
+
+		if !azureMachinePool.GetDeletionTimestamp().IsZero() {
+			c.logger.LogCtx(ctx, "level", "debug", "message", "AzureMachinePool is being deleted, skipping subnet allocation")
+			return false, nil
+		}
 	}
 
 	// Check if Cluster is being deleted. In that case we are skipping subnet allocation.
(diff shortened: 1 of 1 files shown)


fiunchinho:
Should we do the same thing for `AzureMachinePool`? Could it happen that we delete a node pool, but then before it's removed the controller `sync` period is due and the controller receives all the `AzureMachinePool` objects ?

nprokopic:
> Should we do the same thing for `AzureMachinePool`? Could it happen that we delete a node pool, but then before it's removed the controller `sync` period is due and the controller receives all the `AzureMachinePool` objects ?

Good point, is sounds like we should do that as well, will add it.

whites11 (approved):

service/controller/resource/ipam/azure_machinepool_subnet_checker.go
```diff
@@ -44,20 +46,46 @@ func NewAzureMachinePoolSubnetChecker(config AzureMachinePoolSubnetCheckerConfig
...
+			Name:      name,
+		}
+
+		azureMachinePool = &expcapzv1alpha3.AzureMachinePool{}
+		err = c.ctrlClient.Get(ctx, objectKey, azureMachinePool)
+		if err != nil {
+			return false, microerror.Mask(err)
+		}
+
+		if !azureMachinePool.GetDeletionTimestamp().IsZero() {
```
nprokopic, line 64: @fiunchinho This is checking the `AzureMachinePool` CR that is currently being reconciled, and bailing out if it is being deleted.

fiunchinho (approved):

PR #990, opened 2020-09-03
Label different node pools differently

fiunchinho:
We add a new label to the `Node` object in k8s so that it contains the node pool id. This will allow customers to schedule different workloads on different node pools using the label.

### DIFF
--- a/e2e/test/nodepool/nodepool_test.go
+++ b/e2e/test/nodepool/nodepool_test.go
@@ -89,7 +89,7 @@ func New(config Config) (*Nodepool, error) {
 	return s, nil
 }
 
-const LabelVmSize = "node.kubernetes.io/instance-type"
+const LabelVmSize = "beta.kubernetes.io/instance-type"
 
 func (s *Nodepool) Test(ctx context.Context) error {
 	clusterID := s.clusterID
--- a/e2e/test/nodepool/nodepool_test.go
+++ b/e2e/test/nodepool/nodepool_test.go
@@ -1,0 +1,384 @@
+// +build k8srequired
+
+package nodepool
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/label"
+	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/reference"
+	expcapzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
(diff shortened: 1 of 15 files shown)


e2e/test/nodepool/nodepool_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.assertRightNumberOfNodes(ctx, setupNodePoolID, newNodepoolID, setupNodePoolReplicas, newNodePoolReplicas, setupNodePoolVmSize, newNodePoolVmSize)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.provider.AddWorker(ctx, clusterID, newNodepoolID)
```
fiunchinho, line 111: This adds a new worker to the specified node pool, which means changing the `MachinePool` object. This tests that reconciliation works fine when changing that object.

e2e/test/nodepool/nodepool_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.assertRightNumberOfNodes(ctx, setupNodePoolID, newNodepoolID, setupNodePoolReplicas, newNodePoolReplicas+1, setupNodePoolVmSize, newNodePoolVmSize)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.provider.ChangeVmSize(ctx, clusterID, newNodepoolID, "Standard_D5_v2")
```
fiunchinho, line 126: This changes the virtual machine size used in the specified node pool, which means changing the `AzureMachinePool` object. This tests that reconciliation works fine when changing that object.

service/controller/key/key.go
```diff
@@ -347,6 +348,7 @@ func KubeletLabelsNodePool(getter LabelsGetter) string {
 
 	labels = ensureLabel(labels, label.Provider, "azure")
 	labels = ensureLabel(labels, label.OperatorVersion, OperatorVersion(getter))
+	labels = ensureLabel(labels, apiextensionslabels.MachinePool, MachinePoolID(getter))
```
fiunchinho, line 351: This is the important change. We add a new label to the `Node` object in k8s so that it contains the node pool id. This will allow customers to schedule different workloads on different node pools using the label.

e2e/test/nodepool/nodepool_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.WaitForNodesReady(ctx, newNodePoolReplicas+setupNodePoolReplicas)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = s.assertRightNumberOfNodes(ctx, setupNodePoolID, newNodepoolID, setupNodePoolReplicas, newNodePoolReplicas, setupNodePoolVmSize, newNodePoolVmSize)
```
fiunchinho, line 106: We assert that the two different node pools contain different number of nodes. Since normally node pools will have different virtual machine sizes, I'm selecting nodes by nodepool label or vm size label, and making sure there are the right number.

e2e/test/nodepool/nodepool_test.go
nprokopic: Maybe we could keep scaling up and scaling down as two separate tests.

Scaling up "simply" adds one node, while scaling down requires draining and moving workload to another node, so it's more likely to hit some issues. Having two separate tests for these two scenarios would make it easier to discover issues.

WDYT?

tuommaki (approved):
LGTM

PR #993, opened 2020-09-10
Use bootstrap cloudconfig generated by spark handler

fiunchinho:
We were still using the old cloudconfig files. We need to use the cloudconfig generated by the `spark` handler instead.

### DIFF
--- a/service/controller/cloudconfig/base_extension.go
+++ b/service/controller/cloudconfig/base_extension.go
@@ -30,6 +30,16 @@ func (e *baseExtension) templateData(certFiles []certs.File) templateData {
 		certsPaths = append(certsPaths, file.AbsolutePath)
 	}
 
+	primaryScaleSetName := key.WorkerVMSSName(e.customObject)
+	if e.azureMachinePool != nil {
+		primaryScaleSetName = key.NodePoolVMSSName(e.azureMachinePool)
+	}
+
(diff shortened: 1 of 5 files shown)


service/controller/resource/blobobject/desired.go
```diff
@@ -154,21 +154,5 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		output = append(output, containerObjectState)
 	}
 
-	{
-		b, err := cc.CloudConfig.NewWorkerTemplate(ctx, ignitionTemplateData, encrypter)
```
fiunchinho, line 158: We no longer need to render the old cloudconfig.

whites11 (approved):
Seems legit

service/controller/cloudconfig/base_extension.go
```diff
@@ -28,6 +30,16 @@ func (e *baseExtension) templateData(certFiles []certs.File) templateData {
 		certsPaths = append(certsPaths, file.AbsolutePath)
 	}
 
+	primaryScaleSetName := key.WorkerVMSSName(e.customObject)
```
fiunchinho, line 33: This piece of code is used for both masters and workers. We only use `MachinePool` when creating workers, so we need to make sure is not `nil`.

PR #995, opened 2020-09-11
Migrate built-in workers to node pool

tuommaki:
This is handler that migrates built-in workers to node pool by reading the
worker VMSS metadata, creating corresponding MachinePool and AzureMachinePool
CRs and eventully moving the workload and deleting the old VMSS.

The handler is developed in TDD fashion and contains initial sketch for Azure
API abstraction and use of gomock for integration testing the handler
implementation.


service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,62 @@
...
+			return nil
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	var machinePool expcapiv1alpha3.MachinePool
+	//var azureMachinePool expcapzv1alpha3.AzureMachinePool
+	{
+		// TODO: Ensure that MachinePool & AzureMachinePool CRs exist for builtinVMSS.
```
tuommaki, line 38: The handler is still very much WIP due to my TDD approach. Currently there's a failing test for this so that I can quickly catch the context on Monday and continue.

service/controller/resource/workermigration/create_test.go
```diff
@@ -0,0 +1,236 @@
...
+		"cluster.yaml",
+		"azureconfig.yaml",
+		"azurecluster.yaml",
+	})
+
+	o, err := loadCR("azureconfig.yaml")
+	if err != nil {
+		t.Fatal(err)
+	}
+	cr := o.(*providerv1alpha1.AzureConfig)
```
tuommaki, line 56: I feel like this is becoming a pattern. We should refactor these to external package for better reusability.

service/controller/resource/workermigration/create_test.go
```diff
@@ -0,0 +1,236 @@
...
+	m.
+		EXPECT().
+		GetVMSS(gomock.Any(), key.ResourceGroupName(*cr), key.WorkerVMSSName(*cr)).
+		Return(newBuiltinVMSS(3, key.WorkerVMSSName(*cr)), nil).
+		Times(1)
+
+	m.
+		EXPECT().
+		DeleteVMSS(gomock.Any(), gomock.Any(), gomock.Any()).
+		Times(0)
```
tuommaki, line 67: This is not necessarily the most beautiful or easiest to read, but I like how easily one can mock implementations and not spend time on writing those stubs.

service/controller/resource/workermigration/create_test.go
```diff
@@ -0,0 +1,236 @@
...
+	err = ctrlClient.List(context.Background(), mpList, opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(mpList.Items) == 0 {
+		t.Fatal("expected at least one MachinePool CR to exist. got 0.")
+	}
+
+	// gomock verifies rest of the assertions on exit.
```
tuommaki, line 87: This is also slightly nasty in terms of readability or explicitness, but I get the reasons why this is like it is. Probably not a problem once we get used to this?

service/controller/resource/workermigration/create_test.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	cr := o.(*providerv1alpha1.AzureConfig)
+
+	m.
+		EXPECT().
+		GetVMSS(gomock.Any(), key.ResourceGroupName(*cr), key.WorkerVMSSName(*cr)).
+		Return(nil, microerror.Mask(autorest.DetailedError{StatusCode: 404})).
+		Times(1)
```
tuommaki, line 118: :heart_eyes:

service/controller/resource/workermigration/internal/azure/spec.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+)
+
+type VMSS *compute.VirtualMachineScaleSet
+
+type API interface {
```
tuommaki, line 11: Look ma! No implementation yet!

service/controller/resource/workermigration/internal/mock_azure/api.go
```diff
@@ -0,0 +1,64 @@
+// Code generated by MockGen. DO NOT EDIT.
```
tuommaki, line 1: This gets generated by `go generate` w/ pragmas in `create_test.go`.

service/controller/resource/workermigration/create_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/yaml"
+
+	"github.com/giantswarm/azure-operator/v4/service/controller/key"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/workermigration/internal/azure"
+	"github.com/giantswarm/azure-operator/v4/service/controller/resource/workermigration/internal/mock_azure"
+)
+
+//go:generate mockgen -destination internal/mock_azure/api.go -source internal/azure/spec.go API
```
tuommaki, line 32: This takes care of mock generation.

service/controller/resource/workermigration/resource.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if config.CtrlClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CtrlClient must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	newResource := &Resource{
+		azureapi:   azure.NewAPI(config.ClientFactory),
```
tuommaki, line 40: Since the abstracted API implementation is still under `internal` I'm hiding it here. Doesn't affect testing since it's in the same package.

tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/13208

tuommaki:
Happy path tests are now more or less done. Next step is to thoroughly test in live cluster. Code should be more or less ready for review.

service/controller/resource/workermigration/resource.go
```diff
@@ -0,0 +1,64 @@
...
+	if config.Location == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Location must not be empty", config)
+	}
+
+	newResource := &Resource{
+		clientFactory: config.ClientFactory,
+		ctrlClient:    config.CtrlClient,
+		logger:        config.Logger,
+		wrapAzureAPI:  azure.GetAPI,
+
```
fiunchinho, line 55: you are not setting the `tenantClientFactory`?

service/controller/resource/workermigration/testdata/azuremachinepool.yaml
```diff
@@ -0,0 +1,22 @@
...
+    cluster.x-k8s.io/cluster-name: c6fme
+    giantswarm.io/cluster: c6fme
+    giantswarm.io/machine-pool: c6fme
+    giantswarm.io/organization: giantswarm
+    release.giantswarm.io/version: 12.0.0
+  name: c6fme
+  namespace: default
+spec:
+  location: westeurope
+  template:
```
fiunchinho, line 15: I believe we need the `datadisks`.

service/controller/resource/workermigration/internal/tenantclient/spec.go
```diff
@@ -0,0 +1,12 @@
+package tenantclient
```
fiunchinho, line 1: you can now use the new package in `pkg/tenantcluster`

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.0-dev"
+	version            = "5.0.0-migrate"
```
fiunchinho, line 8: Remember to change it back, I always forget :sweat_smile:

service/controller/resource/workermigration/internal/azure/spec.go
```diff
@@ -0,0 +1,17 @@
+package azure
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+)
+
+type VMSS *compute.VirtualMachineScaleSet
```
fiunchinho, line 9: I read `VMSS` in a different file and I started wondering why you weren't using a pointer. Then I realized `VMSS` is an alias for a pointer. Wouldn't be more explicit to keep the alias to `type VMSS compute.VirtualMachineScaleSet` so other places would use `*VMSS` making it clear that it's actually a pointer? I don't know if there are any Go guidelines about this.

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+	credentialSecret := &providerv1alpha1.CredentialSecret{
+		Name:      key.CredentialName(cr),
+		Namespace: key.CredentialNamespace(cr),
+	}
+	azureAPI := r.wrapAzureAPI(r.clientFactory, credentialSecret)
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring that built-in workers are migrated to node pool")
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding built-in workers VMSS")
+	var builtinVMSS azure.VMSS
```
fiunchinho, line 48: nitpick: s/builtin/legacy

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+
+		if builtinVMSS.Sku == nil {
+			return microerror.Maskf(executionFailedError, "built-in VMSS Sku is nil")
+		}
+		if builtinVMSS.Sku.Name == nil {
+			return microerror.Maskf(executionFailedError, "built-in VMSS Sku.Name is nil")
+		}
+		if builtinVMSS.Sku.Capacity == nil {
+			return microerror.Maskf(executionFailedError, "built-in VMSS Sku.Capacity is nil")
+		}
```
fiunchinho, line 68: Why would they be `nil`?

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring Spark CR exists for built-in workers VMSS")
+	_, err = r.ensureSparkExists(ctx, cr)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured Spark CR exists for built-in workers VMSS")
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding if node pool workers are ready")
+	if !machinePool.Status.InfrastructureReady || (machinePool.Status.Replicas != machinePool.Status.ReadyReplicas) {
```
fiunchinho, line 95: `machinePool.Status.InfrastructureReady` is set by some handler, copying whatever is in `azureMachinePool.Status.Ready`. Why not using `azureMachinePool.Status.Ready` directly instead?

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+	return nil
+}
+
+func (r *Resource) deleteDrainerConfigs(ctx context.Context, cr providerv1alpha1.AzureConfig) error {
+	o := client.MatchingLabels{
+		capiv1alpha3.ClusterLabelName: key.ClusterID(&cr),
+	}
+
+	var dcList corev1alpha1.DrainerConfigList
+	err := r.ctrlClient.List(ctx, &dcList, o)
```
fiunchinho, line 200: could we just delete all objects that match? Like here https://github.com/kubernetes-sigs/controller-runtime/blob/5757a389803ec368126bb1ff046ae3524dacbfcf/pkg/client/example_test.go#L238

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+		if kubeletDiskSizeGB <= 0 {
+			kubeletDiskSizeGB = 100
+		}
+	}
+
+	// CR didn't exist so it's created here.
+	azureMachinePool = expcapzv1alpha3.AzureMachinePool{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AzureMachinePool",
+			APIVersion: "exp.infrastructure.cluster.x-k8s.io/v1alpha3",
```
fiunchinho, line 253: Are there properties in the `azureMachinePool` variable that we can use instead of _hardcoding_ this here? I mean something like
```
Kind: azureMachinePool.Kind,
APIVersion: azureMachinePool.APIVersion,
```

That way, if version changes, we don't need to manually change it.

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+		return expcapiv1alpha3.MachinePool{}, microerror.Mask(err)
+	}
+
+	var infrastructureCRRef *corev1.ObjectReference
+	{
+		s := runtime.NewScheme()
+		err := expcapzv1alpha3.AddToScheme(s)
+		if err != nil {
+			panic(fmt.Sprintf("expcapzv1alpha3.AddToScheme: %+v", err))
+		}
```
fiunchinho, line 380: why is this needed here? isn't the `scheme` configured before?

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+	ys := make([]string, 0, len(xs))
+	for _, x := range xs {
+		ys = append(ys, strconv.Itoa(x))
+	}
+	return ys
+}
+
+func toInt32P(i int32) *int32 {
+	return &i
+}
```
fiunchinho, line 496: I've been using `autorest/to` for this. We have it as a dependency already.

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,496 @@
...
+	err = r.ctrlClient.Create(ctx, &azureMachinePool)
+	if err != nil {
+		return expcapzv1alpha3.AzureMachinePool{}, microerror.Mask(err)
+	}
+
+	return azureMachinePool, nil
+}
+
+func (r *Resource) ensureDrainerConfigsExists(ctx context.Context, cr providerv1alpha1.AzureConfig) error {
+	cluster, err := r.getClusterForAzureConfig(ctx, cr)
```
fiunchinho, line 297: Should we handle the case when we don't have a `Cluster` just yet and don't error in that case?

service/controller/resource/workermigration/create_test.go
fiunchinho: I think it's fine to read :+1:

service/controller/resource/workermigration/create_test.go
fiunchinho: You mean the fact that gomock asserts the number of calls, the arguments passed and so on? I'm used to this behavior from other languages, so I think it's fine, but I reckon is not as explicit as calling a `verify` method.

fiunchinho (commented):
Some questions, but it's looking good!

service/controller/resource/workermigration/create_test.go
tuommaki: I rather mean the fact that the test fails on mock verification failures only after the whole function executed.

service/controller/resource/workermigration/internal/azure/spec.go
tuommaki: Hmm... You have a point there. I was thinking about the most lightweight mechanism to make have an interface abstraction for Azure SDK types and this worked quite nicely without introduction of actual `interface` and all the boilerplate that comes with it.

I tried to change this but it seems that pointer alias gets "interface treatment" while struct alias actually creates a new type and it cannot be cast from one to another (because it's not an interface type).

service/controller/resource/workermigration/internal/tenantclient/spec.go
tuommaki: Done.

service/controller/resource/workermigration/testdata/azuremachinepool.yaml
tuommaki: Theoretically yeah, but since this isn't used for ignition rendering, it's not strictly necessary for testing purposes.

service/controller/resource/workermigration/resource.go
tuommaki: This is weird. It's definitely there, but somehow GH doesn't render it :scream:

service/controller/resource/workermigration/create.go
fiunchinho: No, this is fine.

tuommaki:
Fishing!

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,505 @@
...
+						DiskSizeGB: int32(kubeletDiskSizeGB),
+						Lun:        to.Int32Ptr(22),
+					},
+				},
+				VMSize: vmSize,
+			},
+		},
+	}
+
+	err = r.ctrlClient.Create(ctx, &azureMachinePool)
```
fiunchinho, line 299: wasn't the `capzcrs` handler creating the `MachinePool` and `AzureMachinePool`?

service/controller/resource/workermigration/create.go
```diff
@@ -0,0 +1,505 @@
...
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", "deleted legacy workers' deployment")
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "deleting legacy workers' VMSS")
+	err = azureAPI.DeleteVMSS(ctx, key.ResourceGroupName(cr), *legacyVMSS.Name)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", "deleted legacy workers' VMSS")
```
fiunchinho, line 174: what's deleting the `AzureConfig` once the migration is done?

service/controller/resource/workermigration/create.go
tuommaki: `capzcrs` takes care of the CP CRs. `AzureConfig` -> {`Cluster`, `AzureCluster`, `AzureMachine`}.

service/controller/resource/workermigration/create.go
tuommaki: Nothing yet, I hope. We reconcile the tenant cluster control plane by it. :slightly_smiling_face:

fiunchinho (approved):
LGTM

PR #996, opened 2020-09-12
Compute tenant cluster k8sclient instead of relying on context

fiunchinho:
Before nodepools we had a handler called `tenantcluster` that added the k8sclient for the tenant cluster to the `controllercontext`. Since `nodepool` handler reacts to `AzureMachinePool`, the `tenantcluster` handler wasn't being called, and we need the client. So I compute it for the `nodepool` handler, but I don't save it in the context.

Not using the `controllercontext` meant passing the tenant cluster client to a bunch of functions.

### DIFF
--- a/service/controller/resource/nodepool/resource.go
+++ b/service/controller/resource/nodepool/resource.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/Azure/go-autorest/autorest/azure/auth"
-	"github.com/giantswarm/errors/tenant"
 	"github.com/giantswarm/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/tenantcluster/v2/pkg/tenantcluster"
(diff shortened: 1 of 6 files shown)


whites11:
Would it be possible to add the tenant client variable as a field of `Resource` maybe?
If it is, it would be cleaner than passing the variable to every func.

fiunchinho:
The problem with that approach is that we can't pass the tenant client while constructing the `Resource`. That means that every function would have to make sure the field is not `nil`. I think it's cleaner passing the client as parameter.

tuommaki (approved):
LGTM

service/controller/resource/nodepool/resource.go
```diff
@@ -53,3 +64,23 @@ func New(config Config) (*Resource, error) {
...
+
+func (r *Resource) getTenantClusterK8sClient(ctx context.Context, cluster *capiv1alpha3.Cluster) (k8sclient.Interface, error) {
+	var k8sClient k8sclient.Interface
+	{
+		restConfig, err := r.tenantRestConfigProvider.NewRestConfig(ctx, key.ClusterID(cluster), cluster.Spec.ControlPlaneEndpoint.String())
+		if err != nil {
+			return k8sClient, microerror.Mask(err)
+		}
+
+		k8sClient, err := k8sclient.NewClients(k8sclient.ClientsConfig{
```
tuommaki, line 76: Just spot a bug here. You need to remove `:` from here. Otherwise the scope variable for success return won't be correct. Fix also error returns with `return nil, microerror.Mask(err)`.

fiunchinho: Fixing it here https://github.com/giantswarm/azure-operator/pull/999

PR #997, opened 2020-09-14
Adjust cluster-autoscaler related tags from VMSSes

whites11:
Removed cluster-autoscaler tags for master VMSS (we don't want to autoscale masters for now).
Fix cluster-autoscaler tags for worker VMSS.


service/controller/resource/masters/template/main.json
```diff
@@ -249,8 +249,6 @@
               "zones": "[if(greater(length(parameters('zones')),0), parameters('zones'), json('null'))]",
               "tags":{
                 "provider":"[toUpper(parameters('GiantSwarmTags').provider)]",
-                "cluster-autoscaler-enabled":"true",
```
whites11, line 252: Disable autoscaler for master nodes

service/controller/resource/instance/template/main.json
```diff
@@ -262,7 +265,7 @@
               "tags":{
                 "provider":"[toUpper(parameters('GiantSwarmTags').provider)]",
                 "cluster-autoscaler-enabled":"true",
-                "cluster-autoscaler-name":"[parameters('vmssName')]",
+                "cluster-autoscaler-name":"[parameters('clusterID')]",
```
whites11, line 268: the cluster autoscaler app uses just the cluster ID as name

whites11: See https://github.com/giantswarm/cluster-autoscaler-app/blob/master/helm/cluster-autoscaler-app/templates/deployment.yaml#L52

fiunchinho (approved):

PR #998, opened 2020-09-14
Order cert files so we have a deterministic cloud config output

fiunchinho:
Our cloud config files content was not deterministic, because it depended on the order of the files and units. The order of the files was changing because it depended on the order in which the go routines finished.


whites11 (approved):
sounds like a good idea

tuommaki (approved):
Very good catch!

PR #1001, opened 2020-09-14
Allow volume expansion for PVCs on the default storage classes

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13211

Allow to expand PVCs created using the default storage classes we deploy on tenant clusters.


fiunchinho (approved):
LGTM

PR #1002, opened 2020-09-14
Use datadisks defined in CR's

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/13240

ClusterAPI types contain fields to specify the disks that will be created in Azure for the docker and kubelet volumes. So far we've only given the customer the possibility to choose the disks sizes. With these changes customers can customize the disks way more.


whites11 (approved):
LGTM

service/controller/resource/nodepool/template/main.json
```diff
@@ -316,24 +322,19 @@
...
-                "caching": "ReadWrite",
-                "createOption": "Empty",
-                "diskSizeGB": "[parameters('kubeletVolumeSizeGB')]",
-                "managedDisk": {
-                  "storageAccountType": "[variables('storageAccountType')]"
-                },
-                "lun": 22
+                "name": "dataDisks",
+                "count": "[length(parameters('dataDisks'))]",
+                "input": {
```
tuommaki, line 329: Should there be `"name": "[copyIndex('dataDisks').NameSuffix]",` inside `input`?

fiunchinho: We didn't name the disks in the past, and @whites11 told me the name is probably ignored when creating disks from VMSS. We can still set it if you think it's best.

PR #1004, opened 2020-09-15
Add tags for cluster autoscaler to node pools ARM templates

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13200

This PR adds the needed tags to use cluster autoscaler with node pools


service/controller/resource/nodepool/template/main.json
```diff
@@ -260,16 +266,18 @@
       "zones": "[if(greater(length(parameters('zones')),0), parameters('zones'), json('null'))]",
       "tags": {
         "provider": "[toUpper(parameters('GiantSwarmTags').provider)]",
-        "cluster-autoscaler-enabled": "true",
-        "cluster-autoscaler-name": "[variables('vmssName')]",
+        "cluster-autoscaler-enabled":"true",
```
fiunchinho, line 269:
```suggestion
        "cluster-autoscaler-enabled": "true",
```

service/controller/key/key.go
```diff
@@ -49,6 +49,8 @@ const (
 
 	AnnotationEtcdDomain        = "giantswarm.io/etcd-domain"
 	AnnotationPrometheusCluster = "giantswarm.io/prometheus-cluster"
+	AnnotationNodePoolMinSize   = "cluster.k8s.io/cluster-api-autoscaler-node-group-min-size"
```
fiunchinho, line 52: We can do it after merging this PR if you prefer. But it'd be cool to have these annotations in [the apiextensions `pkg` package](https://github.com/giantswarm/apiextensions/tree/master/pkg/annotation) like we do with others.

whites11: Yeah I agree but we are still on a very old version of apiextensions and I don't know if that is going to change soon.

axbarsan: Why don't we get the label directly from the CAPI package, as we do with the cluster ID label?

https://github.com/giantswarm/azure-operator/blob/416215e88c0054035ee621bf5235e1650580276a/service/controller/resource/clusterownerreference/resource_test.go#L44

whites11: Yeah that's what Jose asked for.
Requires us to bump apiextensions to 2.2.0.

axbarsan: In my example we import the label from `sigs.k8s.io/cluster-api/api/v1alpha3`. Is updating `apiextensions` required for using that?

whites11: I didn't find that constant in the CAPI repo. Did you find it?

axbarsan: Ah my bad, I just assumed they were there. I now looked around couldn't find them, either 😞

whites11: this is done

fiunchinho (approved):

service/controller/resource/nodepool/template/main.json
```diff
@@ -261,15 +267,17 @@
       "tags": {
         "provider": "[toUpper(parameters('GiantSwarmTags').provider)]",
         "cluster-autoscaler-enabled": "true",
```
tuommaki, line 269: Would it make sense to reflect this flag based on whether specific `MachinePool` has `min` / `max` annotations set or not? Wouldn't `cluster-autoscaler` otherwise always auto-scale the VMSS?

tuommaki (commented):
Not urgent maybe, but related here, did you also check our friend 429 here? https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/azure#azure-api-throttling

service/controller/resource/nodepool/template/main.json
```diff
@@ -261,15 +267,17 @@
...
-        "cluster-autoscaler-name": "[variables('vmssName')]",
+        "cluster-autoscaler-name":"[parameters('clusterID')]",
         "gs-azure-operator.giantswarm.io-version": "[parameters('azureOperatorVersion')]",
         "capi-azure-machine-pool.giantswarm.io-version": "[parameters('azureMachinePoolVersion')]",
-        "capi-machine-pool.giantswarm.io-version": "[parameters('machinePoolVersion')]"
+        "capi-machine-pool.giantswarm.io-version": "[parameters('machinePoolVersion')]",
+        "min":"[int(parameters('minReplicas'))]",
+        "max":"[int(parameters('maxReplicas'))]"
```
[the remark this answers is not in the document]
whites11: Needed by the cluster-autoscaler to know the boundaries.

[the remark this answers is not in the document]
whites11: See https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_manager.go#L291

service/controller/resource/nodepool/template/main.json
whites11: It makes sense to enable it in any case. If the node pool has min == max set, the auto scaler will simply not scale that node pool. If the customer updates the min or max, then it will just work immediately which makes sense to me.

service/controller/resource/nodepool/template/main.json
```diff
@@ -261,15 +267,17 @@
...
       },
       "sku": {
         "name": "[parameters('vmSize')]",
         "tier": "Standard",
-        "capacity": "[parameters('replicas')]"
+        "capacity": "[parameters('minReplicas')]"
```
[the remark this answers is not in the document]
whites11: absolutely, in my todo list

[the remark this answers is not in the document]
whites11: done

PR #1010, opened 2020-09-21
Save VMs status from Azure into CR

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/13256

With this PR we save the VMSS and the VMs id's in the `AzureMachinePool` CR. This will help us save the current status of each node (whether the node is `ready` or not). These fields are set

- `azureMachinePool.Status.ProvisioningState`
- `azureMachinePool.Status.Ready`
- `azureMachinePool.Status.Replicas`
- `azureMachinePool.Spec.ProviderID`
- `azureMachinePool.Spec.ProviderIDList`

I had different options to implement this. I could've created a new handler to take care of this, but I think it makes sense to make the `nodepool` handler (the one creating the `vmss`) to handle saving the status of the `vmss` in Azure.


service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -136,11 +148,53 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
...
+	azureMachinePool.Status.Replicas = int32(len(instances))
+	azureMachinePool.Spec.ProviderID = fmt.Sprintf("azure://%s", *vmss.ID)
+
+	providerIDList := make([]string, len(instances))
+	for i, vm := range instances {
+		providerIDList[i] = fmt.Sprintf("azure://%s", *vm.ID)
+	}
+	azureMachinePool.Spec.ProviderIDList = providerIDList
+
+	err = r.CtrlClient.Update(ctx, azureMachinePool)
```
whites11, line 188: Does calling `Update` store the `Status` fields as well? I thought `UpdateStatus` was required in this case.

whites11:
LGTM when tested, even if it raises the 429 red flag to me because we're calling the vmss list instances API at every reconciliation loop, once more.

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -136,11 +148,53 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 		return currentState, nil
 	default:
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "template and parameters unchanged")
+
+		err := r.saveAzureIDsInCR(ctx, virtualMachineScaleSetsClient, virtualMachineScaleSetVMsClient, &azureMachinePool)
+		if err != nil {
```
fiunchinho, line 153: I'm making sure we don't show errors if we couldn't save the data in the CR. We log the error and it will try again on the next reconciliation loop.

service/controller/resource/nodepool/create_deployment_uninitialized.go
fiunchinho: You are right. I'll fix it.

tuommaki (approved):
LGTM

nprokopic:
> LGTM when tested, even if it raises the 429 red flag to me **because we're calling the vmss list instances API at every reconciliation loop**, once more.

Since `ProviderIDList` is set here, which gives us caching of Azure ID for every VMSS instance, we could use these IDs for per-instance Azure API operations, instead of fetching all instances every time. This would reduce the number of listing operations significantly.

We would call VMSS instance listing only when per-instance Azure API calls are failing because the instance info is stale.

PR #1015, opened 2020-09-22
Use typed ARM template parameters

fiunchinho:


service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -90,11 +91,18 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 		return currentState, microerror.Mask(err)
 	}
 
-	deploymentIsOutOfDate, nodesNeedToBeRolled, err := r.deploymentIsOutOfDate(ctx, currentDeployment, desiredDeployment)
+	deploymentIsOutOfDate, err := template.IsOutOfDate(currentDeployment, desiredDeployment)
```
tuommaki, line 94: :heart_eyes:

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -90,11 +91,18 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
...
 	if err != nil {
 		return currentState, microerror.Mask(err)
 	}
 
+	nodesNeedToBeRolled, err := template.NeedToRolloutNodes(currentDeployment, desiredDeployment)
```
tuommaki, line 99: I'm fine with this now, but would it make sense to have some kind of subtype or something so that nodes are not mixed with template? I feel they are conceptually two different things. I can't think of anything better right now, but maybe just as a food for a thought?

service/controller/resource/nodepool/template/parameters.go
```diff
@@ -0,0 +1,209 @@
...
+	Version   string
+}
+
+func (p Parameters) ToDeployParams() map[string]interface{} {
+	allParams := map[string]interface{}{}
+	allParams["azureOperatorVersion"] = struct {
+		Value interface{}
+	}{
+		Value: p.AzureOperatorVersion,
+	}
```
tuommaki, line 45:
```suggestion
func param(v interface{}) interface{} {
	return struct{
		Value interface{}
	}{
		Value: v,
	}
}

func (p Parameters) ToDeployParams() map[string]interface{} {
	allParams := map[string]interface{}{}
	allParams["azureOperatorVersion"] = param(p.AzureOperatorVersion)
```

Would it make any sense to do this sort of indirection to slightly improve readability / maintainability of parameters?

service/controller/resource/nodepool/template/parameters.go
```diff
@@ -0,0 +1,209 @@
...
+	allParams["zones"] = struct {
+		Value interface{}
+	}{
+		Value: p.Zones,
+	}
+
+	return allParams
+}
+
+var wrongTypeError = &microerror.Error{
```
tuommaki, line 130: Maybe move this to `errors.go`?

service/controller/resource/nodepool/template/template.go
```diff
@@ -23,3 +25,65 @@ func GetARMTemplate() (map[string]interface{}, error) {
...
+		return false, microerror.Mask(err)
+	}
+
+	desiredParameters, err := NewFromDeployment(desiredDeployment)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	rolloutnodes := currentParameters.AzureOperatorVersion != desiredParameters.AzureOperatorVersion ||
+		currentParameters.ClusterID != desiredParameters.ClusterID ||
```
tuommaki, line 77: Would it make any sense to inject logger to this function and log changed parameters by wrapping these comparisons with something like:

```go
func logChange(ctx context.Context, logger micrologger.Logger, annotation string, v bool) {
	if !v {
		logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("change detected for %q", annotation))
	}
}
```

Then each comparison line would be like:
```go
		logChange(ctx, logger, "ClusterID", (currentParameters.ClusterID != desiredParameters.ClusterID)) ||
```

tuommaki (commented):
I like this a lot. Really nice improvement! :+1:

service/controller/resource/nodepool/create_deployment_uninitialized.go
fiunchinho: Yeah, I agree but I don't know how to solve it either. Thinking out loud here, but we could make `IsOutOfDate` return an slice of strings with the name of the parameters that have changed. So you know is out of date if the length of the slice is greater than 0. Then, clients of this function like the `nodepool` handler can make decision based on the slice. Does it only contain the `scaling` field? Then I need to roll out nodes. But this way the node concept is removed from `template`.

service/controller/resource/nodepool/template/parameters.go
fiunchinho: Yeah, I think it's a good idea.

service/controller/resource/nodepool/template/template.go
fiunchinho: not sure I understand how the code would look like for `rolloutnodes`?

fiunchinho: I guess `logChange` needs to return `v`?

service/controller/resource/nodepool/create_deployment_uninitialized.go
fiunchinho: I pushed changes following that idea. Let me know what you think.

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.0-dev"
+	version            = "5.0.0-params"
```
whites11, line 8: Let's leave this as `-dev` if you want to merge it.

whites11 (approved):
I like it a lot. Well done Jose!

service/controller/resource/nodepool/template/template.go
tuommaki: Oh, yes, somehow missed that when writing code in separate comment :slightly_smiling_face:

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -70,37 +65,50 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
...
+		if err != nil {
+			return currentState, microerror.Mask(err)
+		}
+
+		// When customer is only scaling the cluster,
+		// we don't need to move to the next state of the state machine which will rollout all the nodes.
+		numberOfChangedParameters := len(changes)
+		deploymentNeedsToBeSubmitted = numberOfChangedParameters > 0
+		nodesNeedToBeRolled = numberOfChangedParameters > 1 || numberOfChangedParameters == 1 && !contains(changes, "scaling")
+		r.Logger.LogCtx(ctx, "level", "debug", "message", "Checking if deployment is out of date and needs to be re-submitted", "deploymentNeedsToBeSubmitted", deploymentNeedsToBeSubmitted, "nodesNeedToBeRolled", nodesNeedToBeRolled, "changedParameters", changes)
```
tuommaki, line 104: Just beautiful :heart:

tuommaki (approved):
This is just beautiful! Thank you! :heart_eyes:

PR #1017, opened 2020-09-22
Fix scaling error

whites11:
This PR fixes the scaling e2e test that was broken

4 tests are failing because I didn't run them.

### DIFF
--- a/e2e/test/scaling/provider.go
+++ b/e2e/test/scaling/provider.go
@@ -164,7 +164,7 @@ func (p *Provider) WaitForNodes(ctx context.Context, expectedNodes int) error {
 	p.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("waiting for %d k8s nodes to be in %#q state", expectedNodes, v1.NodeReady))
 
 	o := func() error {
-		// Get all all nodes from the kubernetes API.
+		// Get all nodes from the kubernetes API.
 		var nodesReady int
 		var allNodes int
 		{
--- a/e2e/test/scaling/provider.go
+++ b/e2e/test/scaling/provider.go
@@ -5,6 +5,7 @@ package scaling
 import (
 	"context"
 	"fmt"
+
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/e2e-harness/v2/pkg/framework"
--- a/e2e/test/scaling/provider.go
+++ b/e2e/test/scaling/provider.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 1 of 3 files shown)


e2e/test/scaling/provider.go
```diff
@@ -74,26 +77,31 @@ func (p *Provider) findMachinePool(ctx context.Context) (*v1alpha3.MachinePool,
...
 		return microerror.Mask(err)
 	}
 
-	machinePool.Spec.Replicas = to.Int32Ptr(*machinePool.Spec.Replicas + int32(1))
+	if len(machinePools) < 1 {
+		p.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("MachinePool CR for cluster id %q not found", p.clusterID))
+		return microerror.Maskf(notFoundError, fmt.Sprintf("MachinePool CR for cluster id %q not found", p.clusterID))
+	}
+
+	machinePool := machinePools[0]
```
tuommaki, line 97: I'm not sure if k8s api guarantees ordering. Since this is in test, it might make sense to sort the slice before picking up a specific one by index. Alternative option would be to pick it up by name.

e2e/test/scaling/provider.go
```diff
@@ -112,29 +120,83 @@ func (p *Provider) NumMasters(ctx context.Context) (int, error) {
...
+	}
+
+	return nil
+}
+
+func (p *Provider) WaitForNodes(ctx context.Context, expectedNodes int) error {
+	p.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("waiting for %d k8s nodes to be in %#q state", expectedNodes, v1.NodeReady))
+
+	o := func() error {
+		// Get all all nodes from the kubernetes API.
```
tuommaki, line 167:
```suggestion
		// Get all nodes from the kubernetes API.
```

tuommaki (approved):
LGTM; minor not for potential sorting.

e2e/test/scaling/provider.go
whites11: This is not important at all.
For the needs of the scaling test we can scale up and scale down *any* node pool and it should always work

fiunchinho (approved):
Approving but a question for y'all: should we keep running the `scaling` e2e test? The `nodepool` test already scales up, so we could remove that part from the `nodepool` test, or completely remove the `scaling` e2e test. Since they take so long to run, I think I'm leaning towards using just the `nodepool` test, but I understand the value of having different tests for different features.

whites11:
> Approving but a question for y'all: should we keep running the `scaling` e2e test? The `nodepool` test already scales up, so we could remove that part from the `nodepool` test, or completely remove the `scaling` e2e test. Since they take so long to run, I think I'm leaning towards using just the `nodepool` test, but I understand the value of having different tests for different features.

If a test is not there, a test can't break.
+1 for removing it if it's not useful any more.

PR #1021, opened 2020-09-24
Np type change fix

whites11:
This PR fixes a problem in the upgrade process during node rollout for a node pool.
The problem was about correctly detecting the old and new nodes when only the instance type was changed.
It also fixes a race condition between the azure operator and the cluster autoscaler that were fighting to control the number of replicas in a VMSS during nodes rollout.
Finally I changed the instance names of node pool instances from `nodepool-ab123000001` to `nodepool-ab123-000001` because my autism wasn't allowing me to accept it any longer :smile:


tuommaki (approved):
LGTM

PR #1028, opened 2020-09-24
Spot instances support for node pools

whites11:
This PR adds support for creating spot instances node pools.
Since I wasn't able to find any relevant field in the `AzureMachinePool` spec, I used 2 annotations in that CR for enabling node pools and specifying the max price desired.

The PR itself is quite easy, we can discuss tomorrow if we want to merge it or wait after the NP release.


service/controller/resource/nodepool/template/main.json
```diff
@@ -284,6 +299,11 @@
         },
         "singlePlacementGroup": "[parameters('vmssSinglePlacementGroup')]",
         "virtualMachineProfile": {
+          "evictionPolicy": "Delete",
+          "priority": "[if(parameters('spotInstancesEnabled'), 'Spot', 'Regular')]",
+          "billingProfile": {
```
fiunchinho, line 304: will this be ignored if `priority` is `regular`?

whites11: Documentation says:

> Specifies the eviction policy for the Azure Spot virtual machine and Azure Spot scale set.

So I assume it is ignored if you don't have spot instances.
Plus in my tests I didn't notice issues.

fiunchinho (approved):
LGTM

service/controller/key/key.go
```diff
@@ -620,6 +620,28 @@ func NodePoolDeploymentName(azureMachinePool *expcapzv1alpha3.AzureMachinePool)
...
+
+	return enabled
+}
+
+func NodePoolSpotInstancesMaxPrice(azuremachinePool *expcapzv1alpha3.AzureMachinePool) float64 {
+	maxPriceStr := azuremachinePool.Annotations[AnnotationSpotInstancesMaxPrice]
+	maxPrice, err := strconv.ParseFloat(maxPriceStr, 64)
+	if err != nil {
+		// Annotation not found or invalid.
+		return -1
```
tuommaki, line 639: What happens when Azure is given `maxPrice: -1`?

whites11: It automatically sets max price to the on demand price.
It's the default. See https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-12-01/virtualmachinescalesets#BillingProfile

tuommaki (approved):

marians:
I just passed by this PR and was wondering about the terminology (using the term "instance" here in an Azure context. So far we have used the cloud provider specific terminology, even if it created a difference within our product, between providers. Have you had an internal discussion, has using "instances" been a conscious decision?

See also [this Slack thread](https://gigantic.slack.com/archives/C02521EC0/p1601544197065500) for alignment.

whites11:
> I just passed by this PR and was wondering about the terminology (using the term "instance" here in an Azure context). So far we have used the cloud provider specific terminology, even if it created a difference within our product, between providers. Have you had an internal discussion, has using "instances" been a conscious decision?
> 
> See also [this Slack thread](https://gigantic.slack.com/archives/C02521EC0/p1601544197065500) for alignment.

Not that I paid any attention to the terminology here, but VMSS members are called instances in azure.

azure

MartaD:
Feature was renamed from low-priority VMs to Spot VMs - let's use this term in docs.

PR #1029, opened 2020-09-25
Accelerated networking

whites11:
This PR adds the accelerated networking feature to the node pools.
It's enough to set the flag on `AzureMachinePool` to true for the VMSS to use accelerated networking


service/controller/resource/nodepool/deployment.go
```diff
@@ -71,10 +71,11 @@ func (r Resource) getDesiredDeployment(ctx context.Context, storageAccountsClien
...
 
 	templateParameters := template.Parameters{
-		AzureOperatorVersion: project.Version(),
-		ClusterID:            azureCluster.GetName(),
-		DataDisks:            azureMachinePool.Spec.Template.DataDisks,
-		NodepoolName:         key.NodePoolVMSSName(azureMachinePool),
+		AzureOperatorVersion:        project.Version(),
+		ClusterID:                   azureCluster.GetName(),
+		DataDisks:                   azureMachinePool.Spec.Template.DataDisks,
+		EnableAcceleratedNetworking: key.EnableAcceleratedNetworking(*azureMachinePool),
```
fiunchinho, line 77: I'd drop the function in the `key` package and directly use the CR field. If the `AzureMachinePool` type changes, we'd need to change the `EnableAcceleratedNetworking` function anyway to accept the new type.

whites11: the problem is that's a pointer.
if you pass nil to azure it'll get very angry

fiunchinho: I see. I'd still use a local function, but I'm not blocking.

fiunchinho (approved):
LGTM

tuommaki (approved):

PR #1030, opened 2020-09-25
Use opinionated factory that knows how to fetch creds

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/12171

We are fetching the credential secret from different places. This secret is used to read credentials needed by the Azure clients.

I thought it would be more convenient if we had a wrapper for the clients factory that already knew how to fetch the credentials secret. That way we can centralize the logic required to fetch those credentials.

There still some duplication but
- `azureconfig` handler and `azureclusterconfig` handler will go away soon
- `spark` handler will go away


client/organization_factory.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (f *OrganizationFactory) getCredentialSecret(ctx context.Context, objectMeta v1.ObjectMeta) (*v1alpha1.CredentialSecret, error) {
+	f.logger.LogCtx(ctx, "level", "debug", "message", "finding credential secret")
+
+	secretList := &corev1.SecretList{}
+	{
+		err := f.ctrlClient.List(
+			ctx,
+			secretList,
+			client.InNamespace(objectMeta.Namespace),
```
fiunchinho, line 178: This is the important part. Instead of using the default namespace, here we are using the source CR namespace.

client/organization_factory.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		f.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found credential secret %s/%s", credentialSecret.Namespace, credentialSecret.Name))
+
+		return credentialSecret, nil
+	}
+
+	// If no credential secrets are found, we use the default.
+	credentialSecret := &v1alpha1.CredentialSecret{
+		Namespace: objectMeta.Namespace,
```
fiunchinho, line 211: Same here.

service/controller/resource/azureclusterconfig/credential.go
```diff
@@ -27,7 +26,7 @@ func (r *Resource) getCredentialSecret(ctx context.Context, cluster capiv1alpha3
 		err := r.ctrlClient.List(
 			ctx,
 			secretList,
-			client.InNamespace(credentialNamespace),
+			client.InNamespace(cluster.Namespace),
```
fiunchinho, line 29: Same here.

service/controller/resource/azureconfig/credential.go
```diff
@@ -27,7 +26,7 @@ func (r *Resource) getCredentialSecret(ctx context.Context, cluster capiv1alpha3
 		err := r.ctrlClient.List(
 			ctx,
 			secretList,
-			client.InNamespace(credentialNamespace),
+			client.InNamespace(cluster.Namespace),
```
fiunchinho, line 29: Same.

service/controller/resource/spark/resource.go
```diff
@@ -195,7 +195,7 @@ func (r *Resource) getCredentialSecret(ctx context.Context, azureMachinePool key
 		err := r.ctrlClient.List(
 			ctx,
 			secretList,
-			client.InNamespace(credentialNamespace),
+			client.InNamespace(azureMachinePool.Namespace),
```
fiunchinho, line 198: Etc

whites11 (approved):
seems nice and clean.
If it's tested properly seems great to me

tuommaki (approved):

PR #1043, opened 2020-09-29
Ensure orphaned azureconfig finalizer gets deleted on upgrade

tuommaki:
With node pools azure-operator got several new controllers and in order to make
it easier to maintain, the naming of controllers was changed a bit. This
resulted in orphaned finalizer when upgrading cluster from pre-NP release where
azureconfig controller didn't have any suffix.

This change adds simple handler that ensures this orphaned finalizer gets
deleted.

Closes https://github.com/giantswarm/giantswarm/issues/13500


service/controller/resource/azureconfigfinalizer/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	var exists bool
+	for i, v := range cr.Finalizers {
+		if strings.TrimSpace(v) == legacyFinalizer {
+			exists = true
+
+			// Drop it.
+			cr.Finalizers = append(cr.Finalizers[:i], cr.Finalizers[i+1:]...)
+			break
```
whites11, line 74: Instead of break here, you could update the CR straight away and return I guess.
But not important

whites11 (approved):

nprokopic (approved):
LGTM

service/controller/resource/azureconfigfinalizer/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring azureconfig doesn't have orphaned azure-operator finalizer present")
+
+	{
+		// Refresh the CR object.
+		err := r.ctrlClient.Get(ctx, client.ObjectKey{cr.Name, cr.Namespace}, &cr)
```
nprokopic, line 61: CircleCI build failed here:

```
service/controller/resource/azureconfigfinalizer/resource.go:61:32: k8s.io/apimachinery/pkg/types.NamespacedName composite literal uses unkeyed fields
```

service/controller/resource/azureconfigfinalizer/resource.go
tuommaki: Sure, but I feel the loop would get a bit too messy with all the logging & stuff. I prefer it this way. :slightly_smiling_face:

fiunchinho (approved):
One comment is not right, LGTM otherwise.

service/controller/resource/azureconfigfinalizer/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	// Finalizer of old operator's controller.
+	legacyFinalizer = "operatorkit.giantswarm.io/azure-operator"
+)
+
+type Config struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+}
+
+// Resource manages the cluster ID label in the AzureConfig CR.
```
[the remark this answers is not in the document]
tuommaki: you didn't see anything

tuommaki:
Wokrs

PR #1044, opened 2020-09-29
Credential secrets may live in the giantswarm namespace for a while

fiunchinho:
While the migration to organization scope CR's is being done, the organization credentials will live in the `giantswarm` namespace. When the migration is complete, the credentials will live in the organization namespace.

So I change the code so that it tries to find the credentials in the organization namespace first. In case it can't find them, it will try to find them in the `giantswarm` namespace.


tuommaki (approved):
LGTM

client/organization_factory.go
```diff
@@ -192,27 +217,61 @@ func (f *OrganizationFactory) getCredentialSecret(ctx context.Context, objectMet
...
+	}
+
+	f.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found credential secret %s/%s", credentialSecret.Namespace, credentialSecret.Name))
+
+	return credentialSecret, nil
+}
+
+// getLegacyCredentialSecret tries to find a Secret in the default credentials namespace but labeled with the organization name.
+// This is needed while we migrate everything to the org namespace and org credentials are created in the org namespace instead of the default namespace.
+func (f *OrganizationFactory) getLegacyCredentialSecret(ctx context.Context, objectMeta v1.ObjectMeta) (*v1alpha1.CredentialSecret, error) {
```
nprokopic, line 239: `getOrganizationCredentialSecret` and `getLegacyCredentialSecret` look the same, maybe use one function with namespace as a parameter. Not a blocker.

nprokopic (approved):

PR #1046, opened 2020-09-29
Accelerated networking nil

whites11:
This PR aims at enabling the AcceleratedNetworking feature when the CAPZ flag is set to nil, only when the instance type supports it.

### DIFF
--- a/service/controller/internal/vmsku/vmsku.go
+++ b/service/controller/internal/vmsku/vmsku.go
@@ -91,8 +91,6 @@ func (v *Interface) initCache(ctx context.Context) error {
 	for iterator.NotDone() {
 		sku := iterator.Value()
 
-		v.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Adding SKU %s in cache", *sku.Name))
-
 		skus[*sku.Name] = &sku
 
 		err := iterator.NextWithContext(ctx)
@@ -100,6 +98,8 @@ func (v *Interface) initCache(ctx context.Context) error {
 			return microerror.Mask(err)
 		}
 	}
+
+	v.skus = skus
 
 	v.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Number of SKUs in cache: '%d'", len(skus)))
 
--- a/service/controller/internal/vmsku/vmsku.go
+++ b/service/controller/internal/vmsku/vmsku.go
@@ -2,10 +2,13 @@ package vmsku
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
 	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/v4/client"
 )
 
 const (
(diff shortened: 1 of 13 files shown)


whites11:
This is still untested.

service/controller/internal/vmsku/vmsku.go
```diff
@@ -0,0 +1,107 @@
...
+				}
+			}
+		}
+	}
+	return false, nil
+}
+
+func (v *Interface) initCache(ctx context.Context) error {
+	v.logger.LogCtx(ctx, "level", "debug", "message", "Initializing cache for VMSKU")
+	cl, err := v.clientFactory.GetResourceSkusClient("giantswarm", "credential-default")
```
whites11, line 77: Is there a way to get the location of this secret in a better way?

nprokopic: I believe we attempt to get in from the org namespace, and then fallback to the default one, e.g. https://github.com/giantswarm/azure-operator/blob/master/client/organization_factory.go#L178-L180

service/controller/internal/vmsku/vmsku.go
```diff
@@ -0,0 +1,107 @@
...
+	CapabilityAcceleratedNetworking = "AcceleratedNetworkingEnabled"
+)
+
+type Config struct {
+	ClientFactory *client.Factory
+	Location      string
+	Logger        micrologger.Logger
+}
+
+type Interface struct {
```
tuommaki, line 28: Huh? Struct called `Interface`? Should you maybe define the interface in `spec.go` and then have implementation here for that? Then constructor below would be `func New(config Config) (Interface, error) {`

service/controller/internal/vmsku/vmsku.go
```diff
@@ -0,0 +1,107 @@
...
+	}, nil
+}
+
+func (v *Interface) HasCapability(ctx context.Context, vmType string, name string) (bool, error) {
+	if len(v.skus) == 0 {
+		err := v.initCache(ctx)
+		if err != nil {
+			return false, microerror.Mask(err)
+		}
+	}
```
tuommaki, line 58: Would it make sense to wrap this in a function like `v.ensureInititialized()` which would then do `v.initCache()` call once inside an instance specific mutex to make it read nicer and prevent concurrent init calls?

service/controller/internal/vmsku/vmsku.go
```diff
@@ -0,0 +1,107 @@
...
+				if capability.Value != nil && strings.EqualFold(*capability.Value, string(CapabilitySupported)) {
+					return true, nil
+				}
+			}
+		}
+	}
+	return false, nil
+}
+
+func (v *Interface) initCache(ctx context.Context) error {
```
tuommaki, line 75: I think this should be somehow guarded with a mutex to only be initialized once. Alternatively one could also utilize [`sync.Once`](https://golang.org/pkg/sync/#Once).

service/controller/internal/vmsku/vmsku.go
whites11: here I don't really care about the organization. The outcome of this module is shared between all organizations (VM sku details do not change within the same region).
So I am really ok using the default credentials here.

service/controller/internal/vmsku/vmsku.go
whites11: Let me just rename it.

service/controller/internal/vmsku/vmsku.go
fiunchinho: So there are two ways of using credentials in the operator. You either use the control plane credentials, or the tenant cluster credentials.

## Control plane creds

These are easy. They are built at the very beginning and just passed around. If you want to use these, just wire the object when is built.

## Tenant cluster creds

Without the CR we can't know which credentials to use (BYOC). So we can't just build these credentials and pass them around. They need to be built when reacting to a watched object. To build them, we need the CR to infer the credential secret.


So basically, if you don't care about which credentials are used for this particular call, just use the control plane ones. If you want to make the call with the tenant cluster credentials, your interface needs to either receive the credentials object, or the credential secret, or the CR.

service/controller/internal/vmsku/vmsku.go
whites11: done

service/controller/internal/vmsku/vmsku.go
whites11: done with a mutex

service/controller/internal/vmsku/vmsku.go
whites11: sounds good, can you link some place where we use this strategy?

fiunchinho: Not sure which one are you asking for. If you don't care about credentials and using the control plane is fine, maybe you can use this `AzureClientSet`? https://github.com/giantswarm/azure-operator/blob/5504970f61319f91a1e392295e85a53f79156331/service/service.go#L440

Pass it to your object in creation time and that's it.

whites11: Yeah I was referring to the first strategy. Thanks

service/controller/internal/vmsku/vmsku.go
```diff
@@ -0,0 +1,129 @@
...
+	Location      string
+	Logger        micrologger.Logger
+}
+
+type VMSKUs struct {
+	clientFactory *client.Factory
+	location      string
+	skus          map[string]*compute.ResourceSku
+	logger        micrologger.Logger
+	mux           sync.Mutex
```
tuommaki, line 34: Minor nitpick: _mux_ as term is often used in context of HTTP handlers path routing e.g. [stdlib](https://golang.org/pkg/net/http/#ServeMux). Perhaps `initMutex` ?

tuommaki (approved):
I'm good with this now once you settle with Jose. :+1:

service/controller/internal/vmsku/vmsku.go
whites11: should be done

service/controller/internal/vmsku/vmsku.go
whites11: makes a lot of sense. Done, thanks

fiunchinho (approved):

PR #1047, opened 2020-09-29
Use AzureClientSet for tenant clusters, fetching org credentials

fiunchinho:
Every time we create a Azure API client inside a handler we need to search the secret that contains the credentials for the organization. This means that we are hammering the k8s API with every `GetDeploymentsClient` or `GetStorageAccountClient`, which can be executed several times on every handler.

In this PR I wanted to try out using the `AzureClientSet` object that we use when working with the Azure credentials for the Control Plane.
That way, every handler just needs to create the `AzureClientSet` and use whatever clients it needs, and the k8s `Secret` is only searched once.

With this approach we wouldn't use the factories anymore.


whites11:
> Every time we create a Azure API client inside a handler we need to search the secret that contains the credentials for the organization. This means that we are hammering the k8s API with every `GetDeploymentsClient` or `GetStorageAccountClient`, which can be executed several times on every handler.
> 
> In this PR I wanted to try out using the `AzureClientSet` object that we use when working with the Azure credentials for the Control Plane.
> That way, every handler just needs to create the `AzureClientSet` and use whatever clients it needs, and the k8s `Secret` is only searched once.
> 
> With this approach we wouldn't use the factories anymore.

I like the idea a lot

nprokopic:
> Every time we create a Azure API client inside a handler we need to search the secret that contains the credentials for the organization. This means that we are hammering the k8s API with every GetDeploymentsClient or GetStorageAccountClient, which can be executed several times on every handler.

Wait, what, no, we are not hammering anything :smile: We get it from factory cache, that was kinda one of the main points to use the factory :) 

https://github.com/giantswarm/azure-operator/blob/master/client/factory.go#L230-L233

fiunchinho:
We'll try a different approach.

PR #1048, opened 2020-09-30
Use organization credential secret fallback everywhere

fiunchinho:
Using the same fallback approach in the handlers that own their own logic to fetch the credentials.


whites11 (approved):
let's do it, thanks

PR #1049, opened 2020-09-30
Default credential path in key functions

tuommaki:


tuommaki:
I'm not 100% sure about this. Can you intentionally do something stupid as a customer by leaving credentials not defined? I can't think of anything, but I'm also wondering why we didn't do this already?

whites11 (approved):
Seems legit

PR #1052, opened 2020-10-01
Remove randomkeys dependency implementing it locally

fiunchinho:


service/controller/cloudconfig/cloud_config.go
```diff
@@ -114,3 +130,84 @@ func newCloudConfig(template string, params k8scloudconfig.Params) (string, erro
...
+	}{
+		{RandomKey: &cluster.APIServerEncryptionKey, Type: EncryptionKey},
+	}
+
+	for _, k := range keys {
+		err := search(ctx, ctrlClient, k.RandomKey, clusterID, namespace, k.Type)
+		if err != nil {
+			return Cluster{}, microerror.Mask(err)
+		}
+	}
```
tuommaki, line 149: Is there going to be more elements in `keys` in the future or could we drop this indirection and inline `search()` here?

fiunchinho: I have no idea, tbh. I tried to keep it more or less like it was before, but I don't think the current implementation makes much sense now.

tuommaki: Your call, either way. :slightly_smiling_face:

tuommaki (approved):

nprokopic (approved):
LGTM

PR #1053, opened 2020-10-01
Simplify how we find encryption secret

fiunchinho:
I didn't want to block the testing, that's why I didn't add these changes in the previous PR. This PR simplifies the code that we copied from randomkeys. I don't think most of it was necessary.


whites11 (approved):
Looks good

tuommaki (approved):

PR #1056, opened 2020-10-02
Don't expect VMSS to exist on creation

tuommaki:


tuommaki:
Going with a ping to test this most straightforward way.

PR #1058, opened 2020-10-02
Safe scale up

whites11:
When there is a change that requires the rollout of all nodes in a node pool, we apply the deployment and then we proceed with the rollout (we double the size of the NP, cordon old nodes and delete them once drained).
This PR ensures the deployment is applied successfully before it proceeds with the scaling operations, because it happened that we rolled out the nodes despite the deployment was not applied (thus doing the rollout for no reason).


tuommaki (approved):
LGTM

PR #1070, opened 2020-10-05
Return early if reconciled CR is being deleted

fiunchinho:


whites11 (approved):
<3

service/controller/resource/cloudconfigblob/create.go
```diff
@@ -27,6 +27,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if !azureMachinePool.GetDeletionTimestamp().IsZero() {
```
nprokopic, line 30: Do we need the same check for `MachinePool` CR here?

nprokopic: Is it even possible that `EnsureCreated` is called for an `AzureMachinePool` CR that is being deleted?

service/controller/resource/nodepool/create.go
```diff
@@ -41,6 +41,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if !azureMachinePool.GetDeletionTimestamp().IsZero() {
```
nprokopic, line 44: Also here, should we also check `MachinePool` CR?

service/controller/resource/spark/create.go
```diff
@@ -54,6 +54,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if !azureMachinePool.GetDeletionTimestamp().IsZero() {
```
nprokopic, line 57: Also here, should we also check `MachinePool` CR?

service/controller/resource/subnet/resource.go
```diff
@@ -78,6 +78,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if !azureCluster.GetDeletionTimestamp().IsZero() {
```
nprokopic, line 81: Should we also check `Cluster` CR?

service/controller/resource/cloudconfigblob/create.go
fiunchinho: I think so. IIUC controllers get the full list of items every X minutes. That may include deleted items? I can't confirm 100% though.

nprokopic:
I believe we need some unified implementation of some funcs like

```
IsClusterBeingDeleted(cluster capi.Cluster, azureCluster capz.AzureCluster)
// or
IsClusterBeingDeleted(obj metav1.ObjectMeta) // uses cluster labels

IsNodePoolBeingDeleted(machinePool capi.MachinePool, azureMachinePool capz.AzureMachinePool)
// or
IsNodePoolBeingDeleted(obj metav1.ObjectMeta) // uses node pool labels
```

nprokopic:
Not blocking, if this solves the current node pool deletion bugs we found :)

service/controller/resource/cloudconfigblob/create.go
tuommaki: No. `EnsureCreated()` won't be called on object that is deleted (i.e. has non-zero `deletionTimestamp`).

https://github.com/giantswarm/operatorkit/blob/master/pkg/controller/controller.go#L461-L489

tuommaki (commented):
I don't think this is needed?

fiunchinho:
> I don't think this is needed?

I believe it is. Now that we know that operatorkit takes care of checking the deletion timestamp of the reconciled CR, we can remove that part, but we still have to check for other CR's. Otherwise we will get
```
E 10/05 08:43:21 /apis/exp.cluster.x-k8s.io/v1alpha3/namespaces/org-giantswarm/machinepools/h6xs8 failed to reconcile | operatorkit/v2/pkg/controller/controller.go:305 | controller=azure-operator-machine-pool-controller | event=update | loop=112 | version=209552020
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/controller/controller.go:485
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/controller/controller.go:509
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/controller/controller.go:591
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/resource/wrapper/metricsresource/basic_resource.go:43
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/resource/wrapper/retryresource/basic_resource.go:64
        /go/pkg/mod/github.com/giantswarm/backoff@v0.2.0/retry.go:23
        /go/pkg/mod/github.com/giantswarm/operatorkit/v2@v2.0.1/pkg/resource/wrapper/retryresource/basic_resource.go:52
        /root/project/service/controller/resource/machinepoolownerreference/resource.go:72
        unknown: azuremachinepools.exp.infrastructure.cluster.x-k8s.io "h6xs8" not found
```

service/controller/resource/cloudconfigblob/create.go
fiunchinho: I removed the check for the reconciled CR and added checks for other CRs that are used but are not the ones being watched.

fiunchinho:
I was wrong, the root cause was not handling the "Not found" error gracefully. I added a check to make sure we don't error in that situation.

I don't think the other code harms? I mean the code to return early if CR is being deleted.

tuommaki (approved):
> I don't think there is harm in the other changes? I mean the code to return early if CR is being deleted.

It doesn't harm, but it's kind of dead code. OperatorKit is wrong if that code is needed / gets executed. I'd maybe drop it, but your call.

fiunchinho:
> > I don't think there is harm in the other changes? I mean the code to return early if CR is being deleted.
> 
> It doesn't harm, but it's kind of dead code. OperatorKit is wrong if that code is needed / gets executed. I'd maybe drop it, but your call.

But operatorkit only checks for the reconciled CR. For example, for the `azuremachinepool_controller` it only checks if the `AzureMachinePool` is being deleted. We are checking if the `Cluster` is being deleted, because there is no point in doing anything else if the whole `Cluster` is being deleted.

nprokopic:
> > > I don't think there is harm in the other changes? I mean the code to return early if CR is being deleted.
> > 
> > 
> > It doesn't harm, but it's kind of dead code. OperatorKit is wrong if that code is needed / gets executed. I'd maybe drop it, but your call.
> 
> But operatorkit only checks for the reconciled CR. For example, for the `azuremachinepool_controller` it only checks if the `AzureMachinePool` is being deleted. We are checking if the `Cluster` is being deleted, because there is no point in doing anything else if the whole `Cluster` is being deleted.

When we are reconciling CR `X`, my understanding is that checking some CR "up the dependency tree" for deletion may be necessary, or at least a form of optimization. Examples:
- Check `Cluster` when reconciling `AzureCluster`
- Check `Cluster` and/or `MachinePool` when reconciling `AzureMachinePool`

Does it make sense?

PR #1072, opened 2020-10-06
Gracefully handle if owner reference is not set

fiunchinho:
Instead of printing out an error in this case, let's just cancel the whole reconciliation and wait until it's done.


tuommaki (approved):
Makes sense.

PR #1073, opened 2020-10-07
Gracefully handle if certificates are not found yet

fiunchinho:


service/controller/resource/spark/create.go
```diff
@@ -114,6 +114,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "ignition blob rendering requirements not met")
 			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
 			reconciliationcanceledcontext.SetCanceled(ctx)
+			return nil
+		} else if certs.IsTimeout(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "waited too long for certificates")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
```
whites11, line 120: are you maybe missing a call to `reconciliationcanceledcontext.SetCanceled(ctx)`?

fiunchinho: Didn't think of it. It makes sense so I'm adding it.

tuommaki (approved):

PR #1074, opened 2020-10-07
Use cluster CIDR for security groups & configuration with node pools

tuommaki:
Since there are no built-in workers anymore but each node pool gets own subnet,
it makes more sense to relax security rules to whole tenant cluster CIDR.


whites11 (approved):

service/controller/resource/deployment/deployment.go
```diff
@@ -46,7 +46,7 @@ func (r Resource) newDeployment(ctx context.Context, customObject providerv1alph
 		"virtualNetworkName":         key.VnetName(customObject),
 		"vnetGatewaySubnetName":      key.VNetGatewaySubnetName(),
 		"vpnSubnetCidr":              vpnSubnet.String(),
-		"workerSubnetCidr":           key.WorkersSubnetCIDR(customObject),
+		"workerSubnetCidr":           key.AzureConfigNetworkCIDR(customObject),
```
fiunchinho, line 49: we should probably change the name of the parameter then

PR #1075, opened 2020-10-07
Adjust the worker SSH rule more precisely to avoid side-effects

tuommaki:


service/controller/resource/deployment/template/main.json
```diff
@@ -428,7 +428,7 @@
                       "sourcePortRange":"*",
                       "destinationPortRange":"22",
                       "sourceAddressPrefix":"[parameters('hostClusterCidr')]",
-                      "destinationAddressPrefix":"[parameters('workerSubnetCidr')]",
+                      "destinationAddressPrefix":"[parameters('virtualNetworkCidr')]",
```
whites11, line 431: You changed a .json file but I don't see the pkged.go file. Did you run pkger?

whites11: Plus it doesn't seem this parameter is defined in the ARM template

tuommaki: :scream: 

Totally forgot the whole thing.

tuommaki: Done. Tested. W0krs. Reverted `project.go`.

tuommaki:
Going with a ping :bell:

PR #1090, opened 2020-10-13
Add normalization function for organization namespaces

tuommaki:
In order to normalize free form organization names into namespaces, it must be
converted into lower case string and non-alphanumeric (+ dash) characters must
be converted / stripped. The length of the string must also be limited.

### DIFF
--- a/pkg/normalize/normalize.go
+++ b/pkg/normalize/normalize.go
@@ -1,0 +1,51 @@
+package normalize
+
+import (
+	"strings"
+	"unicode"
+)
+
+const maxDNSLabelLength = 63
+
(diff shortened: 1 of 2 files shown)


tuommaki:
This ain't used yet but if we migrate CAPI + CAPZ CRs into organization namespace when upgrading cluster, then we need this. As I wrote this code already, I figured it makes sense to commit it as well now.

whites11 (approved):
Nice job!

whites11:
@axbarsan might have sense to use this approach in the API as well maybe?

axbarsan:
> @axbarsan might have sense to use this approach in the API as well maybe?

Yes! I'll steal it for kubectl-gs and the api

PR #1091, opened 2020-10-13
Storage account type

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13699

This PR avoids changing the storage account type from standard to premium when the AzureMachinePool gets edited and the instance type for the node pool changes from one not supporting premium storage to one that supports it.

This is needed because azure does not support such change.


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-dev"
+	version            = "5.0.0-dev"
```
tuommaki, line 8: Accidental or intentional?

tuommaki (commented):
Would it make sense to maintain SA type of node pool in `AzureMachinePool` CR?

There's field for that under `OSDisk`: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha3/types.go#L324

That way we could prevent changing this field in admission controller. WDYT?

pkg/project/project.go
whites11: accidental

whites11: fixed

tuommaki (approved):
Code wise LGTM

PR #1093, opened 2020-10-13
Encryption secret should live in the same namespace than the CR

fiunchinho:
We were using the wrong namespace name, because org namespaces are all lowercase, but the organization label is not lowercase.

So I figured: this was the only part in azure-operator where the `organization` namespace shows up. So what about removing that and just use the CR namespace? That way azure-operator doesn't care about the org namespace , or its format: it just uses whatever namespace the CR is in.


T-Kukawka (approved):
Seems legit imo

whites11 (approved):
Absolutely agree

PR #1101, opened 2020-10-14
Instrument Azure API client with basic metrics collection

tuommaki:
In order to get better insight into azure-operator behavior regarding Azure API
calls, this change instruments Azure API clients with send decorator that
records basic API call metrics.

Towards https://github.com/giantswarm/giantswarm/issues/13758


client/senddecorator/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	titleName := strings.ReplaceAll(lowerName, "_", " ")
+	titleName = strings.Title(titleName)
+	titleName = strings.ReplaceAll(titleName, "Dns", "DNS")
+	titleName = strings.ReplaceAll(titleName, "Gw", "GW")
+	titleName = strings.ReplaceAll(titleName, "Ip", "IP")
+	titleName = strings.ReplaceAll(titleName, "Nat", "NAT")
+	titleName = strings.ReplaceAll(titleName, "Skus", "SKUs")
+	titleName = strings.ReplaceAll(titleName, "Vms", "VMs")
+
+	totalCallsOpts := prometheus.Opts{Namespace: metricsNamespace, Name: fmt.Sprintf("%s_total_calls", lowerName), Help: fmt.Sprintf("Total number of %s API calls", titleName), ConstLabels: prometheus.Labels{"subscription_id": subscriptionID}}
```
whites11, line 30: Would it be better or worse to have one single metric per type and the lowerName as a label?

tuommaki: That's a good shout. I need to think about it how to get it done, because the current prom interfaces don't have `With()` method to supply additional labels. I'll dig around a bit.

whites11 (approved):
LGTM

service/collector/azure_api_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return nil
+}
+
+func (c *AzureAPIMetricsCollector) GetCounterVec(opts prometheus.Opts, labelNames []string) *prometheus.CounterVec {
+	k := opts.Namespace + "/" + opts.Name
+	counter, exists := c.counters[k]
+	if !exists {
+		c.mutex.Lock()
```
fiunchinho, line 68: why not use the mutex at the beginning of the function and only have one `if` statement?

client/senddecorator/configurator.go
```diff
@@ -2,19 +2,14 @@ package senddecorator
...
+// autorest Client.
 //
 // Existing SendDecorators are preserved, but moved to end of slice.
-func ConfigureClient(g *backpressure.Backpressure, c *autorest.Client) {
-	c.SendDecorators = append([]autorest.SendDecorator{
-		// NOTE: Order matters here since these decorators are executed in
-		// order. See: https://godoc.org/github.com/Azure/go-autorest/autorest#Client
-		RateLimitCircuitBreaker(g),
-	}, c.SendDecorators...)
+func WrapClient(c *autorest.Client, decorators ...autorest.SendDecorator) {
```
fiunchinho, line 11: I'm missing something. When is this function used?

service/collector/azure_api_metrics.go
tuommaki: It's only needed once when creating the object.

client/senddecorator/configurator.go
tuommaki: Here: https://github.com/giantswarm/azure-operator/pull/1101/files#diff-ed5adad93d3328a35e51172ed8aed66598cfef9e2c769de9c92287bf99bc2c65R192-R200

fiunchinho (approved):
LGTM, well done

PR #1102, opened 2020-10-14
Try to send only one request to VMSS Azure API from nodepool handler

fiunchinho:
I realized that we were sending several different requests from the `nodepool` handler to fetch the same VMSS. I tried to fetch it once and pass it around so we can save some requests on every reconciliation loop.

I haven't tested it yet. What do you think?

### DIFF
--- a/service/controller/resource/nodepool/create_deployment_uninitialized.go
+++ b/service/controller/resource/nodepool/create_deployment_uninitialized.go
@@ -91,7 +91,7 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	}
 
 	// Compute desired state for Azure ARM Deployment.
-	desiredDeployment, err := r.getDesiredDeployment(ctx, storageAccountsClient, release, machinePool, &azureMachinePool, cluster, azureCluster, vmss)
+	desiredDeployment, err := r.getDesiredDeployment(ctx, storageAccountsClient, release, machinePool, &azureMachinePool, azureCluster, vmss)
 	if IsNotFound(err) {
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "Azure resource not found", "stack", microerror.JSON(err))
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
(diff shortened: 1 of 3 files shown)


service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -187,14 +194,9 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	}
 }
 
-func (r *Resource) saveAzureIDsInCR(ctx context.Context, virtualMachineScaleSetsClient *compute.VirtualMachineScaleSetsClient, virtualMachineScaleSetVMsClient *compute.VirtualMachineScaleSetVMsClient, azureMachinePool *capzexpv1alpha3.AzureMachinePool) error {
+func (r *Resource) saveAzureIDsInCR(ctx context.Context, virtualMachineScaleSetsClient *compute.VirtualMachineScaleSetsClient, virtualMachineScaleSetVMsClient *compute.VirtualMachineScaleSetVMsClient, azureMachinePool *capzexpv1alpha3.AzureMachinePool, vmss compute.VirtualMachineScaleSet) error {
```
whites11, line 197: vmss can be nil, does this even compile?

whites11: I mean, shouldn't it be a pointer?

whites11: Oh, now I got how it works.
Maybe I would use a pointer and set it to nil in case of 404, then check for pointer != nil where you want to use the `vmss`.
But I like the idea in general

tuommaki: If we can use non-pointer value (especially given that e.g. autorest `Response` provides functionality), I feel it's better. Less unnecessary nil checks and less potential for null pointer dereferencing, but your call. I'm not blocking on this.

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -153,7 +160,7 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "ARM deployment has failed, re-applying")
 		r.Debugger.LogFailedDeployment(ctx, currentDeployment, err)
 
-		err := r.saveAzureIDsInCR(ctx, virtualMachineScaleSetsClient, virtualMachineScaleSetVMsClient, &azureMachinePool)
+		err := r.saveAzureIDsInCR(ctx, virtualMachineScaleSetsClient, virtualMachineScaleSetVMsClient, &azureMachinePool, vmss)
```
tuommaki, line 163: Do we need the vmss clients when we pass `vmss` instead?

fiunchinho: Good point, I'll remove it.

service/controller/resource/nodepool/create_deployment_uninitialized.go
fiunchinho: Yeah, I'd rather not to deal with pointers.

tuommaki (approved):

PR #1103, opened 2020-10-14
Use operatorkit version that fixes the pause bug

fiunchinho:


whites11 (approved):
When it works seems legit

PR #1104, opened 2020-10-15
Do not use ssh key

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13783

This PR ignores the `SSHPublicKey` field in the `AzureMachine` and `AzureMachinePool` CRs when creating an azure VMSS.
The ssh key is now hardcoded to a public key who's private key is not available.
(We need to hardcode the value because it is required by Azure VMSS ARM template spec)

Please note that we never used the ssh key provisioned by the azure deployment to ssh into tenant cluster nodes.
Instead we use `vault` and, as a backup, public ssh keys that we provision via ignition.
So from a functional point of view this change has no impact at all.


service/controller/resource/masters/template/main.json
```diff
@@ -280,7 +277,7 @@
                       "ssh":{
                         "publicKeys":[
                           {
-                            "keyData":"[parameters('vmssSshPublicKey')]",
+                            "keyData":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCuJvxy3FKGrfJ4XB5exEdKXiqqteXEPFzPtex6dC0lHyigtO7l+NXXbs9Lga2+Ifs0Tza92MRhg/FJ+6za3oULFo7+gDyt86DIkZkMFdnSv9+YxYe+g4zqakSV+bLVf2KP6krUGJb7t4Nb+gGH62AiUx+58Onxn5rvYC0/AXOYhkAiH8PydXTDJDPhSA/qWSWEeCQistpZEDFnaVi0e7uq/k3hWJ+v9Gz0qqChHKWWOYp3W6aiIE3G6gLOXNEBdWRrjK6xmrSmo9Toqh1G7iIV0Y6o9w5gIHJxf6+8X70DCuVDx9OLHmjjMyGnd+1c3yTFMUdugtvmeiGWE0E7ZjNSNIqWlnvYJ0E1XPBiyQ7nhitOtVvPC4kpRP7nOFiCK9n8Lr3z3p4v3GO0FU3/qvLX+ECOrYK316gtwSJMd+HIouCbaJaFGvT34peaq1uluOP/JE+rFOnszZFpCYgTY2b4lWjf2krkI/a/3NDJPnRpjoE3RjmbepkZeIdOKTCTH1xYZ3O8dWKRX8X4xORvKJO+oV2UdoZlFa/WJTmq23z4pCVm0UWDYR5C2b9fHwxh/xrPT7CQ0E+E9wmeOvR4wppDMseGQCL+rSzy2AYiQ3D8iQxk0r6T+9MyiRCfuY73p63gB3m37jMQSLHvm77MkRnYcBy61Qxk+y+ls2D0xJfqxw== fake",
```
tuommaki, line 280: Is this the original one or just random garbage?

pkg/checksum/checksum_test.go
```diff
@@ -93,13 +92,15 @@ func Test_getDeploymentParametersChecksum(t *testing.T) {
 				t.Fatalf("Unable to construct a deployment: %v", err)
 			}
 
+			fmt.Printf("%+v\n", deployment)
```
tuommaki, line 95: Debug print?

service/controller/resource/masters/template/main.json
whites11: original one

pkg/checksum/checksum_test.go
whites11: yes, thanks

service/controller/resource/masters/template/main.json
tuommaki: Not sure if this should be just random garbage or the original one. I guess there's not that much of a difference. Your call.

tuommaki (approved):

whites11:
@tuommaki I tried to use a garbage key, but microsoft actually validates it so I need a valid public key.
Rather than generating a new one, I decided to keep using the same we always used.

PR #1105, opened 2020-10-15
Ssh users

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13783

This PR uses the list of ssh public keys coming from draughtsman (as an helm parameter in the configmap) to generate the ignition config file, ignoring the field in AzureConfig CR.

We need to do this otherwise the customer could inject their SSH keys in tenant clusters.


service/controller/azure_machine_pool.go
```diff
@@ -54,7 +55,7 @@ type AzureMachinePoolConfig struct {
 	OIDC                      setting.OIDC
 	RegistryDomain            string
 	SentryDSN                 string
-	SSHUserList               string
+	SSHUserList               []v1alpha1.ClusterKubernetesSSHUser
```
fiunchinho, line 58: Can't we keep the type to `string`? I think there is value in having `azure_machine_pool` controller not depending on `v1alpha1` types, and only use go types or CAPI/CAPZ types.

whites11: we can, at the cost of duplicating the parsing function in two places (or moving it into a shared location).
What about having a new local type to hold this information?

fiunchinho: Duplication is fine sometimes, specially having in mind that `AzureConfig` related stuff will be removed in the following months. Anyway, your call.

whites11: I created an external module for parsing the flag and pass around a new custom type between controllers.
WDYT?

fiunchinho (approved):
Awesome!

PR #1106, opened 2020-10-16
bumped operatorkit

whites11:
this PR bumps operatorkit to latest patch to solve a problem in cluster deletion


whites11:
going with a ping

PR #1110, opened 2020-10-19
Removed instance watchdog to save on VMSS API calls

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13828

This PR removes the instance watchdogs we used to monitor the health of VMSS instances during operations like scaling or updating of a VMSS.

This seems not needed any more and allows us to save a good number of VMSS API calls


service/controller/internal/vmsscheck/vmss.go
```diff
@@ -13,6 +13,10 @@ import (
 )
 
 const (
+	// Provisioning States.
+	provisioningStateFailed    = "Failed"
+	provisioningStateSucceeded = "Succeeded"
```
fiunchinho, line 18: Where do these come from? Are they used?

fiunchinho (approved):
Just one comment, looks good otherwise.

service/controller/internal/vmsscheck/vmss.go
whites11: they are used, they were set in a file that was deleted

fiunchinho (approved):

PR #1113, opened 2020-10-21
Disable autoscaler

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13841

This PR changes the nodepool ARM deployment in order to set the VMSS' `cluster-autoscaler-enabled` tag to false when the min and max replicas are equal.
This stops the `cluster-autoscaler` to query azure API about that VMSS and saves some API calls.


fiunchinho (approved):
LGTM

PR #1127, opened 2020-10-29
Upgrade from NP release to NP release

tuommaki:
This change provides first version of implementation that propagates
cluster upgrade in controlled manner over all cluster node pools.

It ensures that Cluster -> AzureCluster CRs have matching release labels
and then propagates these labels to MachinePools one-by-one while
ensuring that there's only one node pool upgrading at a time.

MachinePool labels are then directly ensured on respective
AzureMachinePool CR and corresponding CR's conditions are used to
monitor when given cluster-subset has been upgraded.


service/controller/resource/clusterupgrade/create.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (r *Resource) propagateReleaseToMachinePools(ctx context.Context, cr capiv1alpha3.Cluster) error {
+	lst := expcapiv1alpha3.MachinePoolList{}
+	err := r.ctrlClient.List(ctx, &lst, client.MatchingLabels{capiv1alpha3.ClusterLabelName: key.ClusterName(&cr)})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if !hasMachinePoolsUpgradeCompleted(lst.Items, cr.Labels[label.ReleaseVersion]) {
```
nprokopic, line 120: Is this the correct place to do this check?

It returns if some `MachinePool` did not finish upgrading, but the upgrade itself is triggered after this in code, when setting the release labels. So if the upgrade check is correct, it will always initially return `false`, as we would never start the upgrade.

I understood we would check if `MachinePool` `n` is upgraded before triggering the upgrade for `MachinePool` `n+1`, but I don't see that here.

Or I'm missing something?

nprokopic: Or not...

I mean, this will be new azure-operator that will be running and reconciling CR anyway, so the nodepool handler would just pick up the `AzureMachinePool` CRs anyway and start doing whatever it is doing with them, which includes submitting Azure deployments and doing the entire upgrade.

So in order to enforce the upgrade order, I believe we would need to change `nodepool` resource, so it does not proceed with upgrades until some criteria is met.

So what you implemented here, setting release and azure-operator labels, that would trigger the setting of `Upgrading` to `True` (what I have in another branch), and then in `nodepool` handler we need to check if `Upgrading` condition is set, and proceed with upgrade only it is set to `True`, otherwise wait in the same state.

tuommaki: I don't fully follow.

My strategy / thinking here is following:

- First ensure that CP gets upgraded. This is currently ensured by checking master nodes' `azure-operator` version label from tenant api directly. Slightly ugly, but works for now until we get specific types for CP.
- Then proceed to node pools as follows:
  - To proceed, all `MachinePool` CRs must have `Upgrading` condition false for to-be-upgraded release version. We also need to check here if labels were already updated (i.e. `machine-pool-controller` didn't pick up the CR update yet).
  - When no node pool is upgrading, pick one and update labels.
    - Now `machine-pool-controller` gets event (eventually), sets the `Upgrading` condition and reflects the change to `AzureMachinePool`.
      - `azure-machine-pool` controller is straightforward and works by detecting drift in desired infrastructure vs. present infrastructure and therefore it proceeds upgrading the VMSS.

My reasoning for this is that upgrade strategy should be controlled on abstract high level since it's nothing provider specific. I'd also strongly discourage towards dealing with sibling CRs because for given controller they are fully different things working in different context and coordination between them is responsibility of component that manages them, by that very definition. There are few exceptions to this, such as IPAM where installation-specific uniqueness must be guaranteed, but even that one should probably be implemented by a separate operator.

Does that make any sense?

fiunchinho: Maybe we can talk tomorrow about the whole flow, it's not easy to understand. At the end I'd like to have a comment in the code explaining the whole thing.

tuommaki:
Going with this now. Pretty much all functionality here has been reviewed in here or in sub-PRs. This has been tested in several ways and no remaining upgrade issues are known.

PR #1129, opened 2020-10-30
Pass dockerhub token for authorized pulls

corest:
Towards https://github.com/giantswarm/giantswarm/issues/13981


whites11 (approved):
LGTMIIWFY

fiunchinho:
Does this mean that we will pay DockerHub to be able to pull images? or authenticating using a free account give us bigger limits than anonymous pulls?

corest:
@fiunchinho we pay dockerhub

service/controller/cloudconfig/cloud_config.go
```diff
@@ -78,6 +80,7 @@ func New(config Config) (*CloudConfig, error) {
 
 		azure:                  config.Azure,
 		azureClientCredentials: config.AzureClientCredentials,
+		dockerhubToken:         config.DockerhubToken,
```
fiunchinho, line 83: should we check if the token is empty and fail in that case?

fiunchinho (approved):
for the CAPI/CAPZ versions we use the `spark` resource to generate the worker cloud config. You'd need to update it too. Also remember to update the changelog.

fiunchinho (changes requested):
for the CAPI/CAPZ versions we use the spark resource to generate the worker cloud config. You'd need to update it too. Also remember to update the changelog.

fiunchinho (approved):
Ok, Vol just made me realize this is not targeting the master branch so no need to change the `spark` handler.

service/controller/cloudconfig/cloud_config.go
corest: yes, added

fiunchinho (approved):

PR #1137, opened 2020-10-30
Add MachinePool Upgrading condition

nprokopic:
What's in the box:
- Setting `MachinePool` `Upgrading` condition in `machinepoolconditionshandler`
  - Initializing to `False`, when the `MachinePool` is just created
  - Setting to `True` when the upgrade has been started, which is determined by checking desired release and azure-operator version, as well as actual azure-operator version set on Nodes.
  - Setting to `False` when the upgrade has been completed, which is determined by checking desired release and azure-operator version, as well as actual azure-operator version set on Nodes.
- Added new subpackage `pkg/conditions` with common condition functionalities (more stuff will be moved here).

### DIFF
--- a/pkg/conditions/check.go
+++ b/pkg/conditions/check.go
@@ -1,0 +1,35 @@
+package conditions
+
+import (
+	aeconditions "github.com/giantswarm/apiextensions/v3/pkg/conditions"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


nprokopic:
Scratch this, I need these changes on top of Tuomas' upgrade branch.

PR #1138, opened 2020-10-30
MachinePool Upgrading condition

nprokopic:
What's in the box:
- Setting `MachinePool` `Upgrading` condition in `machinepoolconditionshandler`
  - Initializing to `False`, when the `MachinePool` is just created
  - Setting to `True` when the upgrade has been started, which is determined by checking desired release and azure-operator version, as well as actual azure-operator version set on Nodes.
  - Setting to `False` when the upgrade has been completed, which is determined by checking desired release and azure-operator version, as well as actual azure-operator version set on Nodes.
- Added new subpackage `pkg/conditions` with common condition functionalities (more stuff will be moved here).
- Before triggering a node pool upgrade, checking if previous node pool upgrades have been completed.


service/controller/resource/clusterupgrade/create.go
```diff
@@ -111,18 +111,37 @@ func (r *Resource) ensureMasterHasUpgraded(ctx context.Context, cluster capiv1al
...
-		return nil
-	}
+	for i, mp := range lst.Items {
+		if cr.Labels[label.ReleaseVersion] == mp.Labels[label.ReleaseVersion] &&
+			cr.Labels[label.AzureOperatorVersion] == mp.Labels[label.AzureOperatorVersion] {
+			continue
+		}
+
+		if i > 0 {
+			previousMachinePool := &lst.Items[i-1]
```
tuommaki, line 126: If possible, I'd propose to avoid anything that is based on sequentiality. Sorting of items is missing here, but even with that the correctness is not guaranteed if `MachinePool` is created with "unlucky" name/ID between reconciliation rounds (newly created CR would become `previousMachinePool`).

service/controller/resource/nodepool/create_deployment_uninitialized.go
```diff
@@ -133,6 +139,13 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	if deploymentNeedsToBeSubmitted {
 		r.Logger.LogCtx(ctx, "level", "debug", "message", "template or parameters changed")
 
+		if thisIsUpgrade && !conditions.IsUpgradingTrue(machinePool) {
```
tuommaki, line 142: As I wrote in the NP upgrade PR's comment, I'd propose to keep the reconciliation logic straightforward and let the handler here only consider desired vs. present state, if possible.

My current thinking goes that upgrade logic should be possible to keep at higher level while provider specific implementation then ensures that running infrastructure matches to desired state that CR specification represents.

That way it should be hopefully easier to reason the overall logic of reconciliation. WDYT?

service/controller/resource/clusterupgrade/create.go
nprokopic: > Sorting of items is missing here

Yeah, I wasn't sure about this one, thanks for confirming :+1:

service/controller/resource/nodepool/create_deployment_uninitialized.go
nprokopic: > My current thinking goes that upgrade logic should be possible to keep at higher level while provider specific implementation then ensures that running infrastructure matches to desired state that CR specification represents.
> That way it should be hopefully easier to reason the overall logic of reconciliation. WDYT?

Yeah, that would be easier to reason about definitely, but I believe that we would have to change current `nodepool` handler implementation then.

Correct me if I'm wrong, this is my understanding of the current upgrade process on an example (for `nodepool` handler):
- we have a cluster with 5 node pools
- upgrade is triggered from `v20.0.0` to `v21.0.0`
- new azure-operator version starts to reconcile the cluster
- for every `AzureMachinePool`, `nodepool` handler compares current Azure ARM deployment/template to the desired ARM deployment/template
    - it finds the difference for at least azure-operator version
    - it submits a new VMSS deployment
    - it does the above deployment checks/submitting for all `AzureMachinePool` CRs
    - so effectively all node pools are upgraded in parallel (so rolling, draining and the whole shebang), which could disrupt customer's workload, and this is something that we wanted to avoid with sequential node pool upgrades (one node pool upgrade at a time), if I understood correctly how we wanted to have NP to NP upgrade.

So my understanding is that we would have to change current `nodepool` handler implementation, if we want to upgrade one node pool at a time.

service/controller/resource/clusterupgrade/create.go
nprokopic: Here then, in every reconciliation loop, we should check that every node pool is either already upgraded, or the upgraded did not even start. And after checking them all, we take one and update its version labels.

service/controller/resource/nodepool/create_deployment_uninitialized.go
tuommaki: > it does the above deployment checks/submitting for all AzureMachinePool CRs
> so effectively all node pools are upgraded in parallel

No, the trick is that operator version label in `MachinePool` / `AzureMachinePool` CR guards this. As long as the labels are not updated, the old operator is still reconciling those CRs. Only once the labels are updated, the new operator picks up with them.

nprokopic: If we can achieve to handle upgrades on a higher non-provider level (so I guess on CAPI level), that would really be more understandable.

I'm just not sure if that's possible in every scenario, as we sometimes have very specific upgrade steps to do, especially when upgrading/migrating to a new release where we have bigger changes, so it can be difficult to have a "dumb" handler that just "simply" ensures desired state that is finds in the CR.

service/controller/resource/clusterupgrade/create.go
tuommaki: Yes, and as I wrote in the other thread, we need to account for situation where `MachinePool` labels are updated to new release, but the conditions are not updated yet. In that case the `Upgrading` should have the previous release version (as discussed last week).

nprokopic: > we need to account for situation where MachinePool labels are updated to new release, but the conditions are not updated yet. In that case the Upgrading should have the previous release version (as discussed last week).

Yup, `upgrade.IsNodePoolUpgradeInProgressOrPending(...)` func is checking condition and the actual state.

After updating `MachinePool` labels, the `machinepoolconditions` handler should almost immediately kick in and update the `Upgrading` condition as well, so this time frame where labels are updated, but `Upgrading` condition is still not set to `True`. should really narrow.

service/controller/resource/clusterupgrade/create.go
```diff
@@ -111,57 +112,70 @@ func (r *Resource) ensureMasterHasUpgraded(ctx context.Context, cluster capiv1al
...
 			}
 
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updated release to machinepool %q", mp.Name))
+			if nodePoolUpgradeInProgressOrPending {
+				// This node pool is still being upgraded, so we want to wait
+				// for this upgrade to be completed.
+				allStartedNodePoolUpgradesAreCompleted = false
+				break
+			}
+		}
```
nprokopic, line 146: If some node pool upgrade was started, and not completed, we wait before starting the upgrade process for the next node pool.

service/controller/resource/clusterupgrade/create.go
```diff
@@ -111,57 +112,70 @@ func (r *Resource) ensureMasterHasUpgraded(ctx context.Context, cluster capiv1al
...
+
+		if machinePoolUpgradeCandidate == nil &&
+			(cr.Labels[label.ReleaseVersion] != machinePool.Labels[label.ReleaseVersion] ||
+				cr.Labels[label.AzureOperatorVersion] != machinePool.Labels[label.AzureOperatorVersion]) {
+
+			// We did not already find next node pool to be upgraded, and this
+			// one does not have version labels updated, so this node pool will
+			// be upgraded next.
+			machinePoolUpgradeCandidate = &machinePool
```
nprokopic, line 155: Here we pick the node pool candidate for the next upgrade. (so a potential upgrade, since we still iterate through the rest of the nodes to check if some of them is not done upgrading)

service/controller/resource/machinepoolconditions/conditionupgrading.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Set initial Upgrading condition status to false, since the MachinePool
+	// is just created.
+	if capiconditions.IsUnknown(machinePool, aeconditions.UpgradingCondition) {
+		err = conditions.MarkUpgradingNotStarted(machinePool)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		return nil
+	}
```
nprokopic, line 33: Current logic for setting the `MachinePool` `Upgrading` condition in the first reconciliation loop - it just gets set to `False`. This should be good enough now, as every node pool is a new node pool (new clusters get new node pools from the start, legacy clusters get a first node pool which is also new, therefore no upgrades in the beginning), so this would practically initialize `Upgrading` condition for all node pools.

However, it could be too simple for other scenarios we talked about, e.g. restoring CRs where we would need to recalculate everything in the `Status`, including conditions.

nprokopic:
@tuommaki @fiunchinho not tested, as it is targeting another dev branch where we could test everything together.

service/controller/resource/machinepoolconditions/conditionupgrading.go
nprokopic: > However, it could be too simple for other scenarios we talked about, e.g. restoring CRs where we would need to recalculate everything in the Status, including conditions.

This is flaw in the implementation for the `Cluster` `Creating` and `Cluster` / `MachinePool` `Upgrading` conditions, while all other `*Ready` conditions are being recalculated all the time, based on the state of resources/components.

pkg/conditions/error.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+var unexpectedConditionStatusError = &microerror.Error{
+	Kind: "UnexpectedConditionStatus",
+}
+
+func UnexpectedConditionStatusError(cr CR, t capi.ConditionType) error {
+	c := capiconditions.Get(cr, t)
+
+	return microerror.Maskf(
```
tuommaki, line 16: I'm not sure if this is a good idea. The return stack trace looses the real place of error origin, when used e.g. like below:
```go
	// Let's make sure that the condition status is set to a supported value.
	if conditions.IsUnexpected(machinePool, aeconditions.UpgradingCondition) {
		return conditions.UnexpectedConditionStatusError(machinePool, aeconditions.UpgradingCondition)
	}
```

The last two entries in returned stack trace would be the caller of function from where this snippet is from, given it used `microerror.Mask()` and then the line from this error constructor.

Maybe this could be changed so that this function is reduced to:
```go
func UnexpectedConditionStatusErrorMsg(cr CR, t capi.ConditionType) string {
	c := capiconditions.Get(cr, t)
	return fmt.Sprintf("Unexpected status for condition %s, got %s", t, c.Status)
}
```

And the call site would be:
```go
	// Let's make sure that the condition status is set to a supported value.
	if conditions.IsUnexpected(machinePool, aeconditions.UpgradingCondition) {
		return microerror.Maskf(conditions.UnexpectedConditionStatusError, conditions.UnexpectedConditionStatusErrorMsg(machinePool, aeconditions.UpgradingCondition))
	}
```

That is not so beautiful, I admit, but it would preserve the expected stack trace.

pkg/upgrade/checks.go
```diff
@@ -0,0 +1,103 @@
...
+func AllNodePoolNodesUpToDate(ctx context.Context, c client.Client, machinePool *capiexp.MachinePool, desiredAzureOperatorVersion string) (bool, error) {
+	nodes := &corev1.NodeList{}
+	var labelSelector client.MatchingLabels
+	{
+		labelSelector = map[string]string{
+			label.MachinePool: machinePool.Name,
+		}
+	}
+
+	err := c.List(ctx, nodes, labelSelector)
```
tuommaki, line 68: Minor nitpick, but would it make sense to just reduce these into:
```suggestion
	err := c.List(ctx, nodes, client.MatchingLabels{label.MachinePool: machinePool.Name})
```

service/controller/resource/clusterupgrade/create.go
```diff
@@ -111,57 +112,70 @@ func (r *Resource) ensureMasterHasUpgraded(ctx context.Context, cluster capiv1al
...
-			mp.Labels[label.ReleaseVersion] = cr.Labels[label.ReleaseVersion]
-			mp.Labels[label.AzureOperatorVersion] = cr.Labels[label.AzureOperatorVersion]
-
-			err = r.ctrlClient.Update(ctx, &mp)
-			if apierrors.IsConflict(err) {
-				r.logger.LogCtx(ctx, "level", "debug", "message", "conflict trying to save object in k8s API concurrently", "stack", microerror.JSON(microerror.Mask(err)))
-				r.logger.LogCtx(ctx, "level", "debug", "message", "cancelling resource")
-				return nil
-			} else if err != nil {
+		machinePool := mp
```
tuommaki, line 124: Any reason why not just rename the iterator variable?

service/controller/resource/clusterupgrade/create.go
tuommaki: While reviewing this function, I'd prefer it to be split into smaller components. One possibility could be with following three functions:
- One that checks if any of the present `MachinePools` is upgrading.
  - If yes, then we can just return at that point since we don't want to start next one.
- One that filters out already upgraded node pools from `MachinePoolList`.
  - In case `len(leftovers) == 0` => Upgrade is done.
  - Otherwise one can pick `1:n` elements from `leftovers` for next upgrade.
- One that ensures given `MachinePool` has up-to-date labels.

WDYT?

service/controller/resource/machinepoolconditions/conditionupgrading.go
tuommaki: >  This should be good enough now, as every node pool is a new node pool (new clusters get new node pools from the start, legacy clusters get a first node pool which is also new, therefore no upgrades in the beginning), so this would practically initialize Upgrading condition for all node pools.

We need to remember this in testing so that we test from a cluster that has been created with this operator version.

tuommaki (commented):
The noted error handling should be fixed so that resulting stack traces are correct. Otherwise minor nitpicks more or less. In general this looks good. :+1:

service/controller/resource/clusterupgrade/create.go
nprokopic: Yeah, that sounds good.

pkg/conditions/error.go
nprokopic: Good point :+1:

service/controller/resource/clusterupgrade/create.go
nprokopic: Further down in code `machinePool` pointer is saved to be used outside of `for` loop. I believe there would be issues if using a pointer to the var declared by `for` loop, that issue where the var `machinePoolUpgradeCandidate` outside of func would point to the last element in the slice (but now when I look at it, that I guess would not be the issue here, as we are not iterating all the way to the end).

Anyway, should not matter if this gets refactored and split into smaller components like described in another comment.

pkg/conditions/upgrading.go
```diff
@@ -3,12 +3,67 @@ package conditions
...
+	corev1 "k8s.io/api/core/v1"
 	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
 	capiconditions "sigs.k8s.io/cluster-api/util/conditions"
 
+	"github.com/giantswarm/azure-operator/v5/pkg/label"
 	"github.com/giantswarm/azure-operator/v5/service/controller/key"
 )
 
+func IsUpgradingInProgress(cr CR, desiredRelease string) (bool, error) {
+	c := capiconditions.Get(cr, aeconditions.UpgradingCondition)
```
nprokopic, line 15: `Get` can return `nil`, so this can panic. There is a `capiconditions.IsTrue(cr, aeconditions.UpgradingCondition)` upstream func which checks that the condition is set and that it is set to `True`.

pkg/conditions/upgrading.go
```diff
@@ -3,12 +3,67 @@ package conditions
...
+	// in progress.
+	return true, nil
+}
+
+func IsUpgraded(cr CR, desiredRelease string) (bool, error) {
+	c := capiconditions.Get(cr, aeconditions.UpgradingCondition)
+	if c.Status == corev1.ConditionTrue {
+		// When Upgrading == True => Upgrading is still in progress.
+		return false, nil
+	}
```
nprokopic, line 50: Same as above for panic and use of `capiconditions.IsTrue(cr, aeconditions.UpgradingCondition)`

fiunchinho (approved):
LET'S DO IT

PR #1139, opened 2020-11-02
Use reflection to construct new zero value runtime.Object

tuommaki:
There is a corner case of empty values with `omitempty` annotation which
combined with DeepCopy() clone of destination object can result with
wrong outcome when reading k8s objects for desired vs. present state
comparison.

In this particular case it happened that desired version AzureCluster
VNET CIDR block contained value, but existing object in k8s API didn't
and due to `omitempty` annotation, the desired value also ended up in
"freshly read" object. This broke detection for update.

By creating new instance of given object via reflection, the Go runtime
ensures that new object is empty zero value and this guarantees correct
behavior for comparison.


service/controller/resource/capzcrs/create.go
```diff
@@ -238,8 +237,8 @@ func (r *Resource) mapAzureConfigToAzureMachine(ctx context.Context, cr provider
 
 func (r *Resource) updateCRs(ctx context.Context, crmappings []crmapping) error {
 	for _, m := range crmappings {
-		// Construct new instance by creating deep copy of desired object.
-		readCR := m.obj.DeepCopyObject()
+		// Construct new instance via reflection to ensure clean zero value object.
+		readCR := reflect.New(reflect.TypeOf(m.obj).Elem()).Interface().(runtime.Object)
```
tuommaki, line 241: To decipher this:
- Get type of `m.obj`.
- Since we know it's a `runtime.Object`, it's guaranteed to be a pointer and we need to dereference it.
- Construct a new instance of it and get a pointer to that value (`New()`).
- Extract `interface{}` from `Value` object.
- Cast `interface{}` to `runtime.Object`.

fiunchinho: Maybe this explanation is worth a comment in the code.

service/controller/resource/capzcrs/create.go
```diff
@@ -318,103 +324,30 @@ func detectAzureClusterUpdate(orig, desired runtime.Object) (bool, error) {
...
-	}
-
-	if !cmp.Equal(d.GetAnnotations(), o.GetAnnotations()) {
-		return true, nil
-	}
-
-	if !cmp.Equal(d, o, cmpopts.IgnoreTypes(&metav1.ObjectMeta{}), cmpopts.IgnoreTypes(&metav1.TypeMeta{}, cmpopts.IgnoreTypes(&capzv1alpha3.Subnets{}))) {
-		return true, nil
-	}
+	fmt.Println("Original cidrblock: ", o.Spec.NetworkSpec.Vnet.CidrBlock)
```
fiunchinho, line 327: heads up

service/controller/resource/capzcrs/create.go
tuommaki: I rewrote it in steps and added comments.

fiunchinho (approved):

PR #1140, opened 2020-11-02
Fix firewall rules

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/14084

This PR fixes the firewall rules for both workers and master nodes that was preventing node to node communication in 13.0.0 releases


MarcelMue (approved):

fiunchinho:
I believe you should remove the parameter from the `security_groups_setup` deployment.

PR #1142, opened 2020-11-05
Simplify master nodes' upgrade process

tuommaki:
When upgrading master nodes, it is not needed to first update the VM
instance (this triggers reboot), then drain the node and eventually
re-image (this reboots the node again) but it should be sufficient to
just directly proceed with reimaging given that all nodes are `Ready` in
k8s api and VMSS instances are in `Successful` state.

This is expected to make master nodes upgrade a lot faster while
maintaining same disruption level as before. Because the initial VM
update rebooted the node already, the following draining doesn't have
any positive effect and hence this whole process can be removed.

Towards giantswarm/giantswarm#13847


service/controller/resource/masters/create.go
```diff
@@ -23,7 +23,6 @@ func (r *Resource) createStateMachine() state.Machine {
 			ProvisioningSuccessful:         r.provisioningSuccessfulTransition,
 			ClusterUpgradeRequirementCheck: r.clusterUpgradeRequirementCheckTransition,
 			MasterInstancesUpgrading:       r.masterInstancesUpgradingTransition,
-			WaitForMastersToBecomeReady:    r.waitForMastersToBecomeReadyTransition,
```
tuommaki, line 26: Master instances upgrade process doesn't proceed if nodes aren't `Ready` and VMs in `Successful` state so this is not needed anymore.

service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -25,26 +24,28 @@ func (r *Resource) masterInstancesUpgradingTransition(ctx context.Context, obj i
 		return "", microerror.Mask(err)
 	}
 
-	versionValue := map[string]string{}
+	r.Logger.LogCtx(ctx, "level", "debug", "message", "finding out if all tenant cluster master nodes are Ready")
 	{
-		for _, node := range cr.Status.Cluster.Nodes {
-			versionValue[node.Name] = node.Version
+		allMasterNodesAreReady, err := r.areNodesReadyForUpgrading(ctx)
```
tuommaki, line 29: Diff here is a mess. Might be easier to compare side-by-side diff.

service/controller/resource/masters/create_cluster_upgrade_requirement_check.go
```diff
@@ -32,5 +32,5 @@ func (r *Resource) clusterUpgradeRequirementCheckTransition(ctx context.Context,
 	}
 
 	// Skip instance rolling by default.
-	return WaitForMastersToBecomeReady, nil
+	return DeploymentCompleted, nil
```
tuommaki, line 35: I'll check if this works. It's affecting cluster creation.

whites11:
> When upgrading master nodes, it is not needed to first update the VM
> instance (this triggers reboot), then drain the node and eventually
> re-image (this reboots the node again) but it should be sufficient to
> just directly proceed with reimaging given that all nodes are `Ready` in
> k8s api and VMSS instances are in `Successful` state.
> 
> This is expected to make master nodes upgrade a lot faster while
> maintaining same disruption level as before. Because the initial VM
> update rebooted the node already, the following draining doesn't have
> any positive effect and hence this whole process can be removed.
> 
> Towards [giantswarm/giantswarm#13847](https://github.com/giantswarm/giantswarm/issues/13847)

Didn't check the code, but according to this comment there might be a problem.
I think the `update` phase is still needed.
Update should apply the VMSS desired state to the instance.
If we skip it, reimaging the instance would not pick up any updates in the VMSS itself IIUC.

tuommaki:
> If we skip it, reimaging the instance would not pick up any updates in the VMSS itself IIUC.

I'm testing / verifying this, but initially I interpreted the documentation that it would perform the update as well:
https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/reimage
> Reimages (upgrade the operating system) one or more virtual machines in a VM scale set which don't have a ephemeral OS disk, for virtual machines who have a ephemeral OS disk the virtual machine is reset to initial state.

whites11:
> > If we skip it, reimaging the instance would not pick up any updates in the VMSS itself IIUC.
> 
> I'm testing / verifying this, but initially I interpreted the documentation that it would perform the update as well:
> [docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/reimage](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/reimage)
> 
> > Reimages (upgrade the operating system) one or more virtual machines in a VM scale set which don't have a ephemeral OS disk, for virtual machines who have a ephemeral OS disk the virtual machine is reset to initial state.

alright, let me know the outcome

tuommaki:
> alright, let me know the outcome

You are right. Specific `Update` is required first and only then it can be reimaged. I'll change that now.

tuommaki:
Successfully tested this. Now we only need to _update_ the master node and it's done after that (so only one reboot). Reimaging is there as a backup, but normally it's not needed and theoretically it could be removed. Now only change pending is to get a new `k8scloudconfig` release w/ patch.

service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -25,26 +24,28 @@ func (r *Resource) masterInstancesUpgradingTransition(ctx context.Context, obj i
...
 		}
-	}
 
-	var drainerConfigs []corev1alpha1.DrainerConfig
-	{
-		n := v1.NamespaceAll
-		o := metav1.ListOptions{
-			LabelSelector: fmt.Sprintf("%s=%s", label.Cluster, key.ClusterID(&cr)),
+		if !allMasterNodesAreReady {
+			r.Logger.LogCtx(ctx, "level", "debug", "message", "found out that all tenant cluster master nodes are not Ready")
```
whites11, line 38: We only have one master node so this is technically true, but the reality is that we `found out that at least one master node is not ready`

tuommaki: Good catch :+1:

service/controller/resource/masters/create_master_instances_upgrading.go
```diff
@@ -57,143 +58,95 @@ func (r *Resource) masterInstancesUpgradingTransition(ctx context.Context, obj i
...
+					continue
+				}
+				if desiredVersion == instanceVersion {
+					continue
+				}
+
+				masterUpgradeInProgress = true
+
+				// Ensure that VM has latest VMSS configuration (includes ignition template etc.).
+				if !*vm.VirtualMachineScaleSetVMProperties.LatestModelApplied {
```
whites11, line 84: We need to test this very carefully in germany west central.
We used to rely on `LatestModelApplied` before and we had to get rid of it because it was not working as expected in GWC.

whites11 (approved):
In general seems good. 
I have one doubt about using `LatestModelApplied` and the `GermanyWestCentral` region, but if it's tested than LGTM

service/controller/resource/masters/create_master_instances_upgrading.go
tuommaki: We did rely on that before as well:
https://github.com/giantswarm/azure-operator/pull/1142/files#diff-aa9eaeec1e181519c60c3144c3c2aca5687f17dbe148b50506cc92ad18b8b2deL345-L351

tuommaki: But in case it doesn't make the node reboot or so, there's the catch w/ reimaging.

PR #1157, opened 2020-11-11
Roll nodes on major changes during release upgrade

tuommaki:
When certain critical / major components such as kubernetes or flatcar
version changes during release upgrade, roll the node. Typical example
of this requirement is to apply security fix with release upgrade. With
current giant swarm approach, the kubernetes version is "baked" into
ignition template so VMs must be recycled in order to apply the patch.

This change adds release version label to cluster nodes and it's
currently not updated if release changes don't require rolling nodes.
This is preferable so that when there is next upgrade[s] in the future,
the change detection knows the components used to build the given node.
In case of release absence, nodes are always rolled.


whites11 (approved):
Seems reasonable

fiunchinho (changes requested):
You need to run `pkger`, also update the CHANGELOG

service/controller/resource/nodestatus/create.go
```diff
@@ -76,12 +77,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	err = r.deleteRetiredNodes(ctx, tenantClusterK8sClient, machinePool.Status.NodeRefs, machinePool.Spec.ProviderIDList)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	nodeRefsResult, err := r.getNodeReferences(ctx, tenantClusterK8sClient, machinePool.Spec.ProviderIDList)
+	nodeRefsResult, err := r.getNodeReferences(ctx, tenantClusterK8sClient, machinePool.Name, machinePool.Spec.ProviderIDList)
```
tuommaki, line 80: @fiunchinho Do you think this specific change is ok?

fiunchinho: Yes, totally.

service/controller/resource/nodes/upgrade.go
```diff
@@ -26,7 +27,9 @@ func AnyOutOfDate(ctx context.Context, destinationRelease string, releases []rel
 		return false, clientNotFoundError
 	}
 
-	nodeList, err := cc.Client.TenantCluster.K8s.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	nodeList, err := cc.Client.TenantCluster.K8s.CoreV1().Nodes().List(ctx, metav1.ListOptions{
```
fiunchinho, line 30: I'd be cool to use the controller client, but I don't want to delay the release, so I'll guess we can change it later.

fiunchinho (approved):
LGTM

PR #1162, opened 2020-11-16
Copy AzureConfig cluster description to Cluster

tuommaki:
For some reason this was missing. Needed when upgrading from
pre-nodepool cluster to nodepool cluster.


service/controller/resource/capzcrs/create.go
```diff
@@ -114,6 +114,9 @@ func (r *Resource) mapAzureConfigToCluster(ctx context.Context, cr providerv1alp
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cr.Name,
 			Namespace: key.OrganizationNamespace(&cr),
+			Annotations: map[string]string{
+				annotation.ClusterDescription: cr.Annotations[annotation.ClusterDescription],
```
tuommaki, line 118: No clue how this was missing, but I spotted this while testing.

whites11 (approved):

PR #1177, opened 2020-11-18
Move handlers inside each controller package

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/14403
Let's not merge this until we release `v13.0.0`.

Every controller is defined in a file called `controller.go`, each inside its own package. The handlers used in that controller got moved inside that package as well. All controllers are called `Controller`,  their config `ControllerConfig`, and there is a function called `NewController` to create them. This is the same on every controller package.

I also moved some packages to the `internal` package so the `controller` folder is cleaner.


service/controller/azureconfig/workermigration/create.go
```diff
@@ -480,7 +480,7 @@ func (r *Resource) isMasterUpgrading(obj providerv1alpha1.AzureConfig) (bool, er
 	var status string
 	{
 		for _, r := range cr.Status.Cluster.Resources {
-			if r.Name != masters.Name {
+			if r.Name != "masters" {
```
fiunchinho, line 483: I wanted to avoid the dependency on the `masters` handler.

service/controller/azureconfig/workermigration/create_test.go
```diff
@@ -76,7 +75,7 @@ func TestMigrationWaitsForMasterUpgradeToFinish(t *testing.T) {
 			}
 			cr := o.(*providerv1alpha1.AzureConfig)
 
-			err = setStatusCondition(ctrlClient, cr, masters.Name, Stage, s)
+			err = setStatusCondition(ctrlClient, cr, "masters", Stage, s)
```
fiunchinho, line 78: Same here.

service/controller/azuremachinepool/nodepool/create.go
```diff
@@ -90,5 +89,5 @@ func isMasterUpgrading(getter key.AnnotationsGetter) bool {
 		return false
 	}
 
-	return masterUpgrading != masters.DeploymentCompleted
+	return masterUpgrading != "DeploymentCompleted"
```
fiunchinho, line 92: I wanted to remove the dependency on the `masters` handler.

service/controller/azuremachinepool/spark/create.go
```diff
@@ -575,8 +575,8 @@ func (r *Resource) newCluster(cluster *capiv1alpha3.Cluster, azureCluster *capzv
 			return providerv1alpha1.Cluster{}, microerror.Mask(err)
 		}
 
-		commonCluster.Calico.CIDR = r.calico.CIDRSize
-		commonCluster.Calico.MTU = r.calico.MTU
+		commonCluster.Calico.CIDR = r.calicoCIDRSize
```
fiunchinho, line 578: Instead of using `CalicConfig` which is inside a handler, I unwrapped them and made the parameters explicit.

nprokopic (approved):
LGTM

Let's take it even further (in other PRs):
- move all non-controller packages from `controllers`
- move many subpackages to top-level `pkg`, I believe we have decent number of candidates:
    - `client` -> `pkg/azure` (here we can have the new API interface)
    - `e2e` -> `tests/e2e`
    - `integration` -> `tests/integration`
    - some stuff currently under `service`
- `pkg/handlers` or `service/handlers` for handlers shared across controllers

PR #1178, opened 2020-11-18
Only submit subnet ARM deployment when needed

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/13590

We were submitting the ARM deployment every time the `AzureCluster` CR changed (because we were using the CR `Generation` field to decide whether or not to submit the deployment). This caused the subnet ARM deployment to be re-submitted frequently, causing `409 Conflict` errors on Azure.

Now we check the subnet name and CIDR and only submit the ARM deployment if the deployment does not exist or those two fields are different from the ARM deployment parameters.


fiunchinho:
I'm testing these changes. This fixes `StatusCode=409 -- Original Error: Code="DeploymentActive"`, which  happened when submitting an ARM deployment that was still in progress.

We still see some errors that are recoverable. So I believe it's fine if we leave them.

### VPN

If we add a node pool while the cluster is being created, the subnet ARM deployment will be `Running` and showing code `429` until the VPN is created and ready. Once is ready, the subnet ARM deployment will become `Succeeded`.

### AnotherOperationInProgress

Apparently only one subnet can be created concurrently. When more than one subnet is being created the ARM deployment will fail with error code `AnotherOperationInProgress`. The `subnet` handler removes failed ARM deployments, so this will be fixed on next reconciliation loop. This only happens when trying to create several subnets at the same time, like when adding several node pools at the same time, or during the creation of the cluster.

service/controller/resource/subnet/resource.go
```diff
@@ -367,41 +379,44 @@ func isSubnetAllowedToStorageAccount(ctx context.Context, storageAccount storage
...
 	if !ok {
-		return false, microerror.Maskf(wrongTypeError, "expected 'string', got '%T'", currentDeployment.Properties.Parameters)
+		return false, microerror.Maskf(wrongTypeError, "expected 'string', got '%T'", currentParams["nodepoolName"].(map[string]interface{})["value"])
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", "Checking if deployment is out of date", "azureClusterVersion", crVersion, "deploymentParameter", deploymentVersion)
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Checking if deployment is out of date for %#q", nodepoolName), "desiredSubnetName", allocatedSubnet.Name, "deploymentSubnetName", nodepoolName, "desiredSubnetCidr", allocatedSubnet.CIDRBlocks[0], "deploymentSubnetCidr", subnetCidr)
 
-	return crVersion != deploymentVersion, nil
+	return allocatedSubnet.Name != nodepoolName || allocatedSubnet.CIDRBlocks[0] != subnetCidr, nil
```
fiunchinho, line 401: This is the main change that makes the deployment to be submitted only when any of the two fields change.

service/controller/resource/subnet/resource.go
```diff
@@ -106,9 +106,13 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	err = r.ensureSubnets(ctx, deploymentsClient, storageAccountsClient, natGatewaysClient, azureCluster)
-	if IsNatGatewayNotReadyError(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "Nat Gateway needs to be in state 'Succeeded' before subnets can be created")
+	err = r.ensureSubnets(ctx, deploymentsClient, storageAccountsClient, natGatewaysClient, &azureCluster)
+	if IsStorageAccountNotFound(err) {
```
tuommaki, line 110: Good catch!

tuommaki (approved):
LGTM

PR #1181, opened 2020-11-19
Refactor ipam.Check() to return existing subnet

tuommaki:
As a preparation for fixing subnet releasing on deletion of node pool,
refactor `ipam.Check()` to return allocated subnet if it exists.

### DIFF
--- a/service/controller/resource/ipam/azure_machinepool_subnet_checker.go
+++ b/service/controller/resource/ipam/azure_machinepool_subnet_checker.go
@@ -2,12 +2,12 @@ package ipam
 
 import (
 	"context"
+	"net"
 
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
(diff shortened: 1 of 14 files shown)


tuommaki:
This is first part of changes towards https://github.com/giantswarm/giantswarm/issues/14411

service/controller/resource/ipam/azure_machinepool_subnet_checker.go
```diff
@@ -58,44 +58,31 @@ func (c *AzureMachinePoolSubnetChecker) Check(ctx context.Context, namespace str
...
-	}
-
-	// Check if Cluster is being deleted. In that case we are skipping subnet allocation.
-	{
-		cluster, err := util.GetClusterFromMetadata(ctx, c.ctrlClient, azureMachinePool.ObjectMeta)
-		if err != nil {
-			return false, microerror.Mask(err)
-		}
-
-		if !cluster.GetDeletionTimestamp().IsZero() {
```
tuommaki, line 77: In general these aren't necessary since OperatorKit takes care of the logic on deletion and in the future case this specific case is actually actively harmful because this code needs to work in `EnsureDeleted()` as well :slightly_smiling_face:

fiunchinho (approved):
comment not up to date anymore, rest LGTM

PR #1182, opened 2020-11-19
Decrease usage of typed clients in favour of controller-runtime client

fiunchinho:


tuommaki (approved):
The change itself looks fine, but not sure if it matches branch name/PR title?

fiunchinho:
> The change itself looks fine, but not sure if it matches branch name/PR title?

True. Title/branch are too optimistic. These changes help going in that direction but we still have places where they are used.

fiunchinho:
I pushed more changes to cover some other places where the typed clients were used.

service/controller/resource/nodepool/create_drain_old_worker_nodes.go
```diff
@@ -57,20 +58,19 @@ func (r *Resource) drainOldWorkerNodesTransition(ctx context.Context, obj interf
 	r.Logger.LogCtx(ctx, "level", "debug", "message", "finding all drainerconfigs")
 
 	drainerConfigs := make(map[string]corev1alpha1.DrainerConfig)
+	var labelSelector client.MatchingLabels
```
tuommaki, line 61: Could this be just:
```go
var labelSelector := client.MatchingLabels{label.Cluster: key.ClusterID(&azureMachinePool)}
```

tuommaki (approved):

service/controller/resource/nodepool/create_drain_old_worker_nodes.go
fiunchinho: it could!

PR #1183, opened 2020-11-19
Implement releasing of allocated subnet on deletion

tuommaki:
When deleting node pool, the subnet allocation must be released in order
to make it available for later node pools.

This change adds an interface for `Releaser` and NOP implementation for
AzureConfig (because cluster resource allocations are released on
deletion automatically) and default implementation for
`AzureMachinePool` which removes the allocated subnet from
`AzureCluster` subnets.

### DIFF
--- a/service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
+++ b/service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
@@ -1,0 +1,81 @@
+package ipam
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"github.com/giantswarm/azure-operator/v5/pkg/helpers"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type AzureMachinePoolSubnetReleaserConfig struct {
+	CtrlClient client.Client
+	Logger     micrologger.Logger
+}
+
+// AzureMachinePoolSubnetReleaser is a Releaser implementation that releases an
+// allocated subnet for a node pool by removing it from AzureCluster CR.
+type AzureMachinePoolSubnetReleaser struct {
+	ctrlClient client.Client
+	logger     micrologger.Logger
+}
+
+func NewAzureMachinePoolSubnetReleaser(config AzureMachinePoolSubnetReleaserConfig) (*AzureMachinePoolSubnetReleaser, error) {
+	if config.CtrlClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CtrlClient must not be empty", config)
+	}
+	if config.Logger == nil {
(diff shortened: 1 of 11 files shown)


tuommaki:
Fixes https://github.com/giantswarm/giantswarm/issues/14411

service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	for i, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
+		if subnet.Name == azureMachinePool.Name {
+			azureCluster.Spec.NetworkSpec.Subnets = append(azureCluster.Spec.NetworkSpec.Subnets[:i], azureCluster.Spec.NetworkSpec.Subnets[i+1:]...)
+		}
+	}
+
+	err = r.ctrlClient.Update(ctx, azureCluster)
+	if err != nil {
```
fiunchinho, line 77: There is a chance this would return a conflict, let's catch that to avoid returning an error.

service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
```diff
@@ -0,0 +1,82 @@
...
+	azureCluster, err := helpers.GetAzureClusterFromMetadata(ctx, r.ctrlClient, azureMachinePool.ObjectMeta)
+	if err != nil {
+		errorMessage := fmt.Sprintf("error while getting AzureCluster CR from AzureMachinePool CR metadata")
+		r.logger.LogCtx(ctx, "level", "warning", "message", errorMessage)
+		return microerror.Mask(err)
+	}
+
+	for i, subnet := range azureCluster.Spec.NetworkSpec.Subnets {
+		if subnet.Name == azureMachinePool.Name {
+			azureCluster.Spec.NetworkSpec.Subnets = append(azureCluster.Spec.NetworkSpec.Subnets[:i], azureCluster.Spec.NetworkSpec.Subnets[i+1:]...)
```
fiunchinho, line 72: should we `break` after this line? not really important regarding efficiency, just to make explicit that only one subnet can be found

fiunchinho (commented):
few comments, looks good overall

service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
tuommaki: Yes, good catch.

service/controller/resource/ipam/azure_machinepool_subnet_releaser.go
tuommaki: Good point. Thanks! :+1:

tuommaki:
I ran conformance tests on this and they passed. :tada:

https://github.com/giantswarm/releases/pull/527
https://tekton.rfjh2.k8s.gorilla.eu-central-1.aws.gigantic.io/#/namespaces/test-workloads/pipelineruns/34c20dc7-2af2-11eb-8876-d23d29000256?pipelineTask=run-tests&step=run-tests

fiunchinho (approved):
niiiiice

PR #1186, opened 2020-11-20
Add `terminate-unhealthy-node` feature

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/14186

This PR adds azure support for the automated node termination feature based on the 
[Bad Node Detector Library](https://github.com/giantswarm/badnodedetector).

It is disabled by default (there is a doc in docs.giantswarm.io about how to enable it, I'm about to update it to consider Azure as well).


pkg/annotation/annotation.go
```diff
@@ -6,4 +6,7 @@ const (
 
 	// UpgradingToNodePools is set to True during the first cluster upgrade to node pools release.
 	UpgradingToNodePools = "release.giantswarm.io/upgrading-to-node-pools"
+
+	// NodeTerminateUnhealthy the name of the annotation needed in the Cluster CR to enable the terminate-unhealthy feature.
+	NodeTerminateUnhealthy = "alpha.node.giantswarm.io/terminate-unhealthy"
```
calvix, line 11: you can import that from apiextensions annotation package

whites11: thanks

service/controller/resource/terminateunhealthynode/create.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Scale VMSS up by one.
+	var vmssName string
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "Retrieving AzureMachinePool CR")
+		amp := capzexpv1alpha3.AzureMachinePool{}
+		err := ctrlClient.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: node.Labels[label.MachinePool]}, &amp)
```
tuommaki, line 124: Here you get the error for unknown kind because the passed `ctrlClient` is the client to tenant cluster which doesn't have CRDs for our types.

service/controller/resource/terminateunhealthynode/create.go
```diff
@@ -0,0 +1,178 @@
...
+		restConfig, err := r.tenantRestConfigProvider.NewRestConfig(ctx, key.ClusterID(cluster), cluster.Spec.ControlPlaneEndpoint.Host)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		k8sClient, err = k8sclient.NewClients(k8sclient.ClientsConfig{
+			Logger:     r.logger,
+			RestConfig: rest.CopyConfig(restConfig),
+			SchemeBuilder: k8sclient.SchemeBuilder{
+				capzexpv1alpha3.AddToScheme,
```
tuommaki, line 98: Here you probably want to register the core types to be able to list `Nodes` etc. from tenant cluster. CAPI/CAPZ types won't exist in tenant cluster.

service/controller/resource/terminateunhealthynode/create.go
whites11: thanks

service/controller/resource/terminateunhealthynode/create.go
```diff
@@ -0,0 +1,175 @@
...
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	var err error
+	cr, err := key.ToCluster(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// check for annotation enabling the node auto repair feature
+	if _, ok := cr.Annotations[annotation.NodeTerminateUnhealthy]; !ok {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "node auto repair is not enabled for this cluster, cancelling")
```
calvix, line 35: maybe this string is outdated and you want to change it to `termiante unhealthy node is not enabled ....`

service/controller/resource/terminateunhealthynode/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+const (
+	gaugeValue float64 = 1
+)
+
+var (
+	nodeAutoRepairTermination = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "aws_operator_node_auto_repair_termination",
```
calvix, line 14: aws -> azure

calvix: maybe adjust metric nam to `azure_operator_terminate_unealthy_node`

calvix (approved):
LGTM

service/controller/key/key.go
```diff
@@ -308,6 +310,26 @@ func DNSZones(customObject providerv1alpha1.AzureConfig) providerv1alpha1.AzureC
 	return customObject.Spec.Azure.DNSZones
 }
 
+func InstanceIDFromNode(node v1.Node) (string, error) {
+	// Extract "000001" from "nodepool-w8kjg-000001".
+	var base36 string
+	{
+		m := regexp.MustCompile(`-[^-]{6}$`)
```
tuommaki, line 317: This could be better when moved to package-global variable:
- The regex needs to be compiled only once because it's static.
- `MustCompile()` is good, but it should be executed when binary is started to get the panic immediately instead of only triggering when there's unhealthy node.

service/controller/resource/terminateunhealthynode/create.go
```diff
@@ -0,0 +1,177 @@
...
+		newCapacity := *vmss.Sku.Capacity + 1
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("Scaling up VMSS %q to %d replicas", vmssName, newCapacity))
+
+		update := compute.VirtualMachineScaleSetUpdate{
+			Sku: &compute.Sku{
+				Capacity: &newCapacity,
+			},
+		}
+
+		_, err = vmssClient.Update(ctx, cluster.Name, vmssName, update)
```
tuommaki, line 146: Does termination below decrease the VMSS capacity back to desired?

tuommaki (commented):
I noticed this is already merged, but left couple notes anyway :slightly_smiling_face:

service/controller/resource/terminateunhealthynode/create.go
whites11: deleting one node automatically decreses the size of VMSS, hence the scale up above

service/controller/resource/terminateunhealthynode/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+
+var (
+	nodeAutoRepairTermination = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "azure_operator_unhealthy_node_termination",
+			Help: "Gauge representing node termination due to node auto repair feature.",
+		},
+		[]string{"cluster_id", "terminated_node", "terminated_instance_id"},
+	)
+)
```
fiunchinho, line 19: IIUC this will add a new metric with value `1` whenever a new node is terminated, and it will be there until the operator is restarted/terminated? or until when?

service/controller/terminate_unhealthy_node.go
```diff
@@ -0,0 +1,192 @@
...
+				return ctx, nil
+			},
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
+			// Name is used to compute finalizer names. This results in something
+			// like operatorkit.giantswarm.io/azure-operator-machine-pool-controller.
+			Name: project.Name() + "-terminate-unhealthy-node-controller",
+			NewRuntimeObjectFunc: func() runtime.Object {
+				return new(v1alpha3.Cluster)
+			},
```
fiunchinho, line 93: Does `operatorkit` allow creating controllers that don't watch any CRD? That way it will work effectively like a cronjob, and we would know that it would be executed every X minutes.

whites11: but we need a crd in this case, beacuse the check has to be executed in the tenant cluster not the control plane

PR #1190, opened 2020-11-24
Add pkger workaround to e2e test packages

tuommaki:
E2e tests has a build constraint `k8srequired`. This makes `go list` to
error when enumerating code inside e2e test packages and as a
consequence, this makes `pkger` to fail because underneath it uses
`github.com/gobuffalo/here` which runs `go list -json`.

This change adds `nop` variable as a separate source file without build
constraint into each test package so that Go compiler can find some
non-restricted code for them and therefore prevent the error.


whites11 (approved):
LGTM when build error is solved

fiunchinho:
image

fiunchinho (approved):

PR #1191, opened 2020-11-24
Moved regexp initialization at package global level

whites11:
Moved regexp MustCompile at global package level to avoid doing it everytime we call the key function


tuommaki (approved):
:rocket:

PR #1193, opened 2020-11-24
Implement migration for worker network fw rules

tuommaki:
Due to change prevention in network security group rules (to guard
customers' custom rules) the firewall rules don't get updated on cluster
upgrade and therefore the destination CIDR for some rules is wrong.

This change implements a migration path to replace built-in worker
subnet CIDR with VNET one.

### DIFF
--- a/client/factory.go
+++ b/client/factory.go
@@ -222,6 +222,17 @@ func (f *Factory) GetNatGatewaysClient(credentialNamespace, credentialName strin
 	return toNatGatewaysClient(client), nil
 }
 
+// GetNetworkSecurityGroupsClient returns *network.SecurityGroupsClient that is used for management of Network Security Groups.
+// The created client is cached for the time period specified in the factory config.
+func (f *Factory) GetNetworkSecurityGroupsClient(credentialNamespace, credentialName string) (*network.SecurityGroupsClient, error) {
+	client, err := f.getClient(credentialNamespace, credentialName, "NetworkSecurityGroupsClient", newNetworkSecurityGroupsClient)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return toNetworkSecurityGroupsClient(client), nil
+}
+
 // GetResourceSkusClient returns *compute.ResourceSkusClient that is used for reading VM instance types.
 // The created client is cached for the time period specified in the factory config.
 func (f *Factory) GetResourceSkusClient(credentialNamespace, credentialName string) (*compute.ResourceSkusClient, error) {
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-pike"
+	version            = "5.0.1-dev"
 )
 
 func Description() string {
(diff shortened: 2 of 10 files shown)


client/factory.go
```diff
@@ -222,6 +222,17 @@ func (f *Factory) GetNatGatewaysClient(credentialNamespace, credentialName strin
 	return toNatGatewaysClient(client), nil
 }
 
+// GetNetworkSecurityGroupsClient returns *network.SecurityGroupsClient that is used for management of Network Security Groups.
+// The created client is cached for the time period specified in the factory config.
+func (f *Factory) GetNetworkSecurityGroupsClient(credentialNamespace, credentialName string) (*network.SecurityGroupsClient, error) {
```
tuommaki, line 227: We need new Azure API client :slightly_smiling_face:

tuommaki (commented):
This is now initial version that hasn't been tested yet. I skipped writing thorough unit tests for this functionality as I don't think it pays off at this point. It's quicker to test against live cluster and that reveals potential real problems with Azure API.

whites11 (approved):
LGTM when tested

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-dev"
+	version            = "5.0.1-porcula"
```
fiunchinho, line 8: image

service/controller/resource/workermigration/security_groups.go
```diff
@@ -0,0 +1,79 @@
...
+	var securityGroups azure.SecurityGroups
+	{
+		securityGroups, err = azureAPI.ListNetworkSecurityGroups(ctx, key.ResourceGroupName(cr))
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	for _, sg := range securityGroups {
+		if sg.Name != nil && *sg.Name == key.WorkerSecurityGroupName(cr) {
```
fiunchinho, line 35: we have only one security group for all node pools (which is the worker security group), right?

service/controller/resource/workermigration/create_test.go
```diff
@@ -570,6 +606,12 @@ func TestFinishedMigration(t *testing.T) {
 	}
 	cr := o.(*providerv1alpha1.AzureConfig)
 
+	m.
```
fiunchinho, line 609: should we also expect `CreateOrUpdate` to be called to update the rules?

fiunchinho: I just read your comment about skipping tests for now. Moving on.

pkg/project/project.go
tuommaki: trololoo

service/controller/resource/workermigration/security_groups.go
tuommaki: Yeah

service/controller/resource/workermigration/create_test.go
tuommaki: Not in existing test cases since the mock returns `nil` for security groups. Therefore none of the loops ever gets executed and therefore nothing needs update.

tuommaki:
Successfully upgraded `12.1.2` into this. Critical worker security rules are now updated accordingly.

tuommaki:
I did another round of tests this morning:
- `12.1.2` -> `13.0.0-beta1` -> b0rked fw rules -> `13.0.0-beta2` -> fw rules fixed
- `13.0.0-beta1` -> `13.0.0-beta2` fw rules were good from beginning and they stay that way; no problems
- `13.0.0-beta2` from scratch; all good

fiunchinho (approved):

PR #1194, opened 2020-11-25
Add CP public IPs to master firewall rules to allow etcd access

whites11:
This PR updates one security rule in the master security group to allow access to the etcd cluster from the control plane nodes.

This is needed by the etcd backup operator and by prometheus.

WARNING: this applies to new clusters only.
Upgrade support TBD


fiunchinho (approved):
LGTM

PR #1197, opened 2020-11-25
Implement migration for worker network fw rules (#1193)

tuommaki:
Due to change prevention in network security group rules (to guard
customers' custom rules) the firewall rules don't get updated on cluster
upgrade and therefore the source and destination CIDR for some
rules is wrong.

This change implements a migration path to replace built-in worker
subnet CIDR with VNET one.


tuommaki:
Cherry-picking the fix from `release-v5.0.x` branch.

whites11 (approved):

PR #1200, opened 2020-11-26
Migrate master's security group rules on upgrade 

tuommaki:
As part of preparation work towards HA masters the etcd LB was changed
from internal to external. This has the effect that control plane
cluster's clients are arriving from different source IPs and these must
be now updated into etcd LB ingress rules when upgrading a pre-NP
cluster.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-snowfall"
+	version            = "5.0.1-dev"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-dev"
+	version            = "5.0.1-snowfall"
```
whites11, line 8: Remember to revert this before merging

tuommaki: Sure, but I need to test this first :)

whites11 (approved):

PR #1208, opened 2020-11-30
Fetch IPs from CP instead of TC to allow CP into TC

fiunchinho:


tuommaki (approved):
Oh wow! To make me understand: The credentials were to TC which doesn't have rights to do the call to CP subscription and that's why it was wrong?

fiunchinho:
> Oh wow! To make me understand: The credentials were to TC which doesn't have rights to do the call to CP subscription and that's why it was wrong?

exactly

PR #1210, opened 2020-11-30
Fetch IPs from CP instead of TC to allow CP into TC

fiunchinho:


whites11 (approved):

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.0.1-dev"
+	version            = "5.0.0-beta7"
```
tuommaki, line 8: Probably revert this before merge.

tuommaki (approved):

PR #1213, opened 2020-12-04
Hardcode sentry DSN because it makes no sense to inject it

whites11:
It doesn't make sense to inject the DNS from the installations repo.
All versions of azure operator should send data to the same project in sentry otherwise having a central error collector doesn't make sense.
For this reason I harcdoded the DSN in the helm chart.

I also bumped to latest microerror


tuommaki (approved):
Does Sentry integration differentiate the operator instance by version somehow?

whites11:
> Does Sentry integration differentiate the operator instance by version somehow?

It probably doesn't. I consider this to be a starting point and to iterate on the iteration by adding features like the one you suggested as we feel the need.

PR #1221, opened 2020-12-07
Improve k8s client user-agent header

tuommaki:
Use `azure-operator/<version>` as user-agent header when calling k8s api. This
is useful when investigating k8s api request call metrics exported by
`k8s-audit-metrics`.


tuommaki:
> Works on `ghost` :trollface: 

Screenshot from 2020-12-07 14-42-53

whites11 (approved):

PR #1223, opened 2020-12-07
Define secret as an included template

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/14650


tuommaki (approved):
Seems legit when tested.

PR #1227, opened 2020-12-10
Fix incorrect Upgrading condition transition from True to False

nprokopic:
Towards https://github.com/giantswarm/giantswarm/issues/14140

`Cluster` `Upgrading` condition is being updated based on `release.giantswarm.io/last-deployed-version` annotation (by comparing it to desired release version).
This fixes how the annotation is updated, and it waits for all node pools to be upgraded before setting/updating the annotation on the `Cluster` CR. That will make `Cluster` `Upgrading` condition to switch from `True` to `False` only when all node pools are upgraded.


tuommaki (approved):
LGTM

PR #1229, opened 2020-12-10
Fix nodepool handler creation

nprokopic:
This fixes nodepool handler creation, which should fix the panic discovered during latest upgrade testing.


whites11 (approved):
LGTM when tested

fiunchinho (approved):

PR #1248, opened 2020-12-21
Added termination events to nodepool VMSSes

whites11:

this PR enables the termination events on the nodepools' VMSSes.

If we merge it, the only consequence as of now is that each node will wait 5 minutes from when the termination command is issues to when the termination actually happens.

But on the bright side we lay the foundations to let nodes drain themselves when a termination is issued to them.

This might happen:

- when cluster autoscaler downscales
- during an upgrade
- during ops activities.

WDYT?


axbarsan (approved):
Makes sense to me ✅

whites11:
@nprokopic since you worked on this, do you think I can merge this PR?
Using the new upgrade strategy will require 2 releases (one with this PR's change to enable events, the other to change the upgrade code) so I guess it makes sense to merge this ASAP (13.1.0) to have the next one merged shortly after and released with 13.2.0

nprokopic:
> @nprokopic since you worked on this, do you think I can merge this PR?
> Using the new upgrade strategy will require 2 releases (one with this PR's change to enable events, the other to change the upgrade code) so I guess it makes sense to merge this ASAP (13.1.0) to have the next one merged shortly after and released with 13.2.0

I am completely on board for the new upgrade strategy (https://github.com/giantswarm/azure-operator/pull/1250), as it simplifies the upgrades and makes use of Azure VMSS automatic updates.

I'm just not sure what are the benfits to merge this now, before completing/merging new upgrade strategy, or more precisely, will it slow down upgrades (as we want to make them quicker in general).
That being said, if you have tested it and you are certain that this will not noticeably slow down upgrades, and there is a benefit in having this merged before finishing/merging automatic VMSS updates with self-draining, feel free to proceed with this.

tuommaki (approved):
I'd wait with this until we handle the events. Otherwise the change is good.

whites11:
> I'd wait with this until we handle the events. Otherwise the change is good.

Handle the events == installing [the app](https://github.com/giantswarm/azure-scheduledevents-manager-app).

My suggestion is to merge this and install the app for new clusters (i.e. add the app in the `Release` CR) already so the event will be managed.

This will allow us to simplify the VMSS state machine for upgrades.

tuommaki:
> My suggestion is to merge this and install the app for new clusters (i.e. add the app in the Release CR) already so the event will be managed.

Works for me. I went through the PRs in wrong order initially so hence I wrote the note earlier. After reviewing the other one and realizing this, this works strategy works for me.

PR #1250, opened 2020-12-22
Added selfdrain feature

whites11:

This PR implements a simplified upgrade state machine for node pools, by leveraging `azure scheduled events` and the [`azure-scheduledevents-manager` app](https://github.com/giantswarm/azure-scheduledevents-manager-app) for draining nodes.

This PR requires termination events to be enabled on the existing cluster, so it can't land in 13.1.0 but it will have to wait until 13.2.0

See ticket for more details


service/controller/resource/nodepool/template/main.json
```diff
@@ -223,6 +223,12 @@
               }
             }
           },
+          "scheduledEventsProfile": {
+            "terminateNotificationProfile": {
+              "notBeforeTimeout": "PT15M",
```
tuommaki, line 228: Is this leftover from testing? I think the other PR had `PT5M` here?

tuommaki (approved):
Few notes that you might want to consider. Otherwise looks very good!

service/controller/resource/nodepool/create_terminate_old_workers.go
```diff
@@ -103,5 +109,62 @@ func (r *Resource) terminateOldWorkersTransition(ctx context.Context, obj interf
...
+		// We'll ignore this node for now and wait for it to bootstrap correctly.
+		return nil, nil
+	}
+
+	nodeVersion := semver.New(v)
+	if nodeVersion.LessThan(*myVersion) {
+		return &t, nil
+	} else {
+		// Check if instance type is up to date.
+		if *instance.Sku.Name != *vmss.Sku.Name {
```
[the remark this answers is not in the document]
fiunchinho: I believe having this check here to allow changing the instance type is confusing. I'd move it out. That way `waitForOldWorkersToBeGoneTransition` checks if `isWorkerInstanceFromPreviousRelease` or the instance type has changed.

nprokopic:
> by leveraging `azure scheduled events` and the [`azure-scheduledevents-manager` app](https://github.com/giantswarm/azure-scheduledevents-manager-app) for draining nodes

I would like us to talk about this part, as it is using 3rd party service (https://github.com/webdevops/azure-scheduledevents-manager) that we would have to deploy to all workload clusters.

whites11:
> > by leveraging `azure scheduled events` and the [`azure-scheduledevents-manager` app](https://github.com/giantswarm/azure-scheduledevents-manager-app) for draining nodes
> 
> I would like us to talk about this part, as it is using 3rd party service ([webdevops/azure-scheduledevents-manager](https://github.com/webdevops/azure-scheduledevents-manager)) that we would have to deploy to all workload clusters.

this is already happening in 13.1.0

nprokopic:
> > > by leveraging `azure scheduled events` and the [`azure-scheduledevents-manager` app](https://github.com/giantswarm/azure-scheduledevents-manager-app) for draining nodes
> > 
> > 
> > I would like us to talk about this part, as it is using 3rd party service ([webdevops/azure-scheduledevents-manager](https://github.com/webdevops/azure-scheduledevents-manager)) that we would have to deploy to all workload clusters.
> 
> this is already happening in 13.1.0

[This app](https://github.com/webdevops/azure-scheduledevents-manager) being deployed in the workload cluster? How come when this is not merged?

whites11:
> > > > by leveraging `azure scheduled events` and the [`azure-scheduledevents-manager` app](https://github.com/giantswarm/azure-scheduledevents-manager-app) for draining nodes
> > > 
> > > 
> > > I would like us to talk about this part, as it is using 3rd party service ([webdevops/azure-scheduledevents-manager](https://github.com/webdevops/azure-scheduledevents-manager)) that we would have to deploy to all workload clusters.
> > 
> > 
> > this is already happening in 13.1.0
> 
> [This app](https://github.com/webdevops/azure-scheduledevents-manager) being deployed in the workload cluster? How come when this is not merged?

This is a managed app, it is declared in the release manifest

https://github.com/giantswarm/releases/blob/azure-13.1.0/azure/v13.1.0/release.yaml#L32

nprokopic:
> > [This app](https://github.com/webdevops/azure-scheduledevents-manager) being deployed in the workload cluster? How come when this is not merged?
> 
> This is a managed app, it is declared in the release manifest
> 
> https://github.com/giantswarm/releases/blob/azure-13.1.0/azure/v13.1.0/release.yaml#L32

I mean how come we are including the app for self-draining in 13.1.0 (what is the purpose), when support on azure-operator side (this PR) is not merged?

azure-operator is currently carrying out the upgrade process in `nodepool` handler, so IMO self-draining (some app to do the draining and azure-operator support) and automatic VMSS updates should be implemented and released all at once.

fiunchinho:
I believe that's the plan. Take into account that [13.1.0 is not released yet](https://github.com/giantswarm/releases/tree/master/azure), so when it's released, it will include the required changes in azure-operator and the new app.

whites11:
> > > [This app](https://github.com/webdevops/azure-scheduledevents-manager) being deployed in the workload cluster? How come when this is not merged?
> > 
> > 
> > This is a managed app, it is declared in the release manifest
> > [https://github.com/giantswarm/releases/blob/azure-13.1.0/azure/v13.1.0/release.yaml#L32](https://github.com/giantswarm/releases/blob/azure-13.1.0/azure/v13.1.0/release.yaml?rgh-link-date=2021-01-11T15%3A56%3A45Z#L32)
> 
> I mean how come we are including the app for self-draining in 13.1.0 (what is the purpose), when support on azure-operator side (this PR) is not merged?
> 
> azure-operator is currently carrying out the upgrade process in `nodepool` handler, so IMO self-draining (some app to do the draining and azure-operator support) and automatic VMSS updates should be implemented and released all at once.

This PR removes the manual draining of nodes during an upgrade, delegating draining to the external app.
For this to work, termination events have to be enabled already on the VM.
Since we are working with "the double the size - delete old workers" strategy to upgrade the VMSSes, this has to be done in two steps.

Step 1 was https://github.com/giantswarm/azure-operator/pull/1248

this PR is step2

service/controller/azuremachinepool/handler/nodepool/template/main.json
```diff
@@ -229,7 +229,7 @@
         "singlePlacementGroup": "[parameters('vmssSinglePlacementGroup')]",
         "virtualMachineProfile": {
           "evictionPolicy": "[if(parameters('spotInstancesEnabled'), 'Delete', json('null'))]",
-          "priority": "[if(parameters('spotInstancesEnabled'), 'Spot', 'Regular')]",
```
whites11, line 232: Before spot instances we didn't specify this field so it was empty.
The field is not editable, so we have to keep it as "" for non-spot instance VMSS

service/controller/azuremachinepool/handler/nodepool/create_wait_for_old_workers_to_be_gone.go
```diff
@@ -0,0 +1,97 @@
...
+	nodePoolVMSSName := key.NodePoolVMSSName(&azureMachinePool)
+	vmss, err := virtualMachineScaleSetsClient.Get(ctx, resourceGroupName, nodePoolVMSSName)
+	if err != nil {
+		return currentState, microerror.Mask(err)
+	}
+
+	var oldWorkersCount int
+	{
+		for _, i := range allWorkerInstances {
+			old, err := r.isWorkerInstanceFromPreviousRelease(ctx, tenantClusterK8sClient, azureMachinePool.Name, i, vmss)
```
tuommaki, line 74: I'd suggest to re-design this a bit. Given I read the code correctly, this seems to now fetch VMSS instances, then iterate them and for each instance list WC nodes to find the corresponding pair.

Alternative proposal:

After fetching VMSS instances, `List()` WC worker nodes (w/ label selector) and convert slice into `map[string]corev1.Node` that is keyed by node name. Then you could change this function call into:
```go
n, exists := nodes[key.NodePoolInstanceName(azureMachinePool.Name, *instance.InstanceID)]
if !isWorkerInstanceFromPreviousRelease(n) || !isWorkerInstanceWithLatestSKU(i, vmss) {
  oldWorkersCount += 1
}
```

Similar approach then for `sortNodesByTenantVMState()` below.

whites11: Thanks for the suggestion, I didn't like that part either and was about to refactor it today.
I am actually thinking of closing this PR and open a new one once finished the refactoring, but don't want you to be upset because you did the review already. WDYT?

tuommaki: I'm totally fine both ways. Whatever works better for you. :+1:

PR #1270, opened 2021-01-19
Ensure the MC's network space is never used for WCs

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15264

this PR enforces the fact that the management cluster's address space is not available for workload clusters.

### DIFF
--- a/service/controller/resource/ipam/virtualnetwork_collector.go
+++ b/service/controller/resource/ipam/virtualnetwork_collector.go
@@ -29,7 +29,8 @@ type VirtualNetworkCollectorConfig struct {
 	K8sClient             k8sclient.Interface
 	Logger                micrologger.Logger
 
-	NetworkRange net.IPNet
+	NetworkRange  net.IPNet
+	ReservedCIDRs []net.IPNet
 }
 
 type VirtualNetworkCollector struct {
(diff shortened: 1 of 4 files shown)


whites11:
Tested, it works as expected

tuommaki (approved):
Code here LGTM, but let's talk about this in standup before merging. I want to understand this better.

service/controller/resource/ipam/virtualnetwork_collector.go
```diff
@@ -79,7 +82,7 @@ func NewVirtualNetworkCollector(config VirtualNetworkCollectorConfig) (*VirtualN
 func (c *VirtualNetworkCollector) Collect(ctx context.Context, _ interface{}) ([]net.IPNet, error) {
 	var err error
 	var mutex sync.Mutex
-	var reservedVirtualNetworks []net.IPNet
+	reservedVirtualNetworks := c.reservedCIDRs
```
nprokopic, line 85: Small nitty detail :) 

Maybe better copy `c.reservedCIDRs` to `reservedVirtualNetworks`, so we don't accidentally modify `c.reservedCIDRs` via `reservedVirtualNetworks`, as it is the same reference this way (which would cause to treat some range as reserved as long as azure-operator pod is running).

Something like:
```suggestion
	reservedVirtualNetworks := append([]whateverTypeHere{}, c.reservedCIDRs...)
```

Makes sense?

nprokopic (approved):
Minor detail. otherwise LGTM.

service/controller/resource/ipam/virtualnetwork_collector.go
whites11: yep, done

PR #1272, opened 2021-01-21
Enable VMSS termination events

fiunchinho:

this PR enables the termination events on the nodepools' VMSSes.

If we merge it, the only consequence as of now is that each node will wait 5 minutes from when the termination command is issues to when the termination actually happens.

image


whites11 (approved):

nprokopic (approved):

calvix:
yeah, everything on azure is already slow co what's another 5 minutes :trollface:

PR #1289, opened 2021-01-25
Move nodes handler

tuommaki:
This change moves `nodes` base handler together with state machine
support functionality, combining them together under
`pkg/handler/nodes`.

Towards https://github.com/giantswarm/giantswarm/issues/14403


nprokopic (approved):
LGTM

PR #1322, opened 2021-01-27
Don't use 'stack' with Debugf

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/15203

### DIFF
--- a/service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
+++ b/service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
@@ -9,6 +9,7 @@ import (
 	azureresource "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/v4/pkg/controller/context/reconciliationcanceledcontext"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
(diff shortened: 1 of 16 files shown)


service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
```diff
@@ -94,12 +95,13 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
...
 	if IsNotFound(err) {
-		r.Logger.Debugf(ctx, "Azure resource not found", "stack", microerror.JSON(err))
+		r.Logger.Debugf(ctx, "Azure resource not found")
 		r.Logger.Debugf(ctx, "canceling resource")
 		return currentState, nil
 	} else if IsSubnetNotReadyError(err) {
-		r.Logger.Debugf(ctx, "subnet is not Ready, it's probably still being created", "stack", microerror.JSON(err))
+		r.Logger.Debugf(ctx, "subnet is not Ready, it's probably still being created")
 		r.Logger.Debugf(ctx, "canceling resource")
+		r.Logger.Debugf(ctx, microerror.JSON(err))
```
fiunchinho, line 104: In this case, it may be useful to see what happened, that's why I tried logging the error anyway. WDYT?

whites11 (approved):

service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
tuommaki: Would it make sense to log it before `canceling resource` ?

tuommaki (approved):

service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
fiunchinho: Done

PR #1325, opened 2021-01-28
Spot instances

whites11:
This is preliminary work to support spot instances on azure opeator.
It's been tested quite a bit, but surely requires more testing from y'all.
I suggest we begin reviewing the PR and merge as soon as we're confident so we can keep testing it as part of the other modifications we'll develop for 14.1.0.

This PR also contains the modifications needed to use latest release of CAPI and CAPZ deps.

WARNING: Because of [this bug](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1151) it is required to edit the `azuremachinepools.exp.infrastructure.cluster.x-k8s.io` CRD to change the type of the `maxPrice` field from `number` to `string` or it is not going to be possible to create spot instance node pools. Regular node pools can be created normally even without the change.


fiunchinho:
> WARNING: Because of [this bug](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1151) it is required to edit the `azuremachinepools.exp.infrastructure.cluster.x-k8s.io` CRD to change the type of the `maxPrice` field from `number` to `string` or it is not going to be possible to create spot instance node pools. Regular node pools can be created normally even without the change.

do we know if it will take long for upstream to fix it?

service/controller/azuremachinepool/handler/nodepool/template/main.json
```diff
@@ -185,7 +205,8 @@
         "gs-azure-operator.giantswarm.io-version": "[parameters('azureOperatorVersion')]",
         "kubernetes-version": "[parameters('kubernetesVersion')]",
         "min": "[int(parameters('minReplicas'))]",
-        "max": "[int(parameters('maxReplicas'))]"
+        "max": "[int(parameters('maxReplicas'))]",
+        "spot": "[if(parameters('spotInstancesEnabled'), 'true', 'false')]"
```
fiunchinho, line 209: I think this line can be removed

fiunchinho (changes requested):

whites11:
> > WARNING: Because of [this bug](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1151) it is required to edit the `azuremachinepools.exp.infrastructure.cluster.x-k8s.io` CRD to change the type of the `maxPrice` field from `number` to `string` or it is not going to be possible to create spot instance node pools. Regular node pools can be created normally even without the change.
> 
> do we know if it will take long for upstream to fix it?

Adrian just opened the bug and nobody answered yet.
The timeframe I'm expecting is in the order of magnitude of weeks, if we are lucky.

service/controller/azuremachinepool/handler/nodepool/template/main.json
whites11: Which line?

fiunchinho: oh it's a tag, I thought it was a parameter. Never mind.

fiunchinho (approved):

tuommaki (approved):
LGTM

PR #1326, opened 2021-01-29
added script to migrate the master node from one az to another one

whites11:


scripts/migrate_master_azs.sh
```diff
@@ -0,0 +1,152 @@
...
+cluster=$2
+
+read -p "Installation: $installation, Cluster: $cluster.
+Press enter to continue, ctrl+c to abort"
+
+opsctl create kubeconfig -i $installation
+gsctl create kubeconfig --cluster=$cluster --certificate-organizations system:masters -e $installation
+
+# get org name
+org_namespace="$(kubectl --context=giantswarm-$installation -n org-giantswarm get cluster -l giantswarm.io/cluster=${cluster} -o yaml -A|yq -r '.items[0].metadata.namespace')"
```
tuommaki, line 15: In the future when writing these, using JSONPath selector directly with `kubectl` would be better so no additional tooling is required. I don't e.g. personally have `yq` installed. See https://kubernetes.io/docs/reference/kubectl/jsonpath/

No need to rewrite these now since that would be quite a big effort for no gain, but in the future the native approach w/ `kubectl` would be nice :slightly_smiling_face:

scripts/migrate_master_azs.sh
```diff
@@ -0,0 +1,152 @@
...
+org="$(kubectl --context=giantswarm-$installation -n org-giantswarm get cluster -l giantswarm.io/cluster=${cluster} -o yaml -A|yq -r '.items[0].metadata.labels["giantswarm.io/organization"]')"
+
+# get azure operator version
+azure_operator_version="$(kubectl --context=giantswarm-$installation -n ${org_namespace} get cluster ${cluster} -o yaml| yq -r '.metadata.labels["azure-operator.giantswarm.io/version"]')"
+
+# find subscription ID
+subscription="$(kubectl --context=giantswarm-$installation -n giantswarm get secret -l giantswarm.io/organization="${org}" -o yaml | yq -r '.items[0].data["azure.azureoperator.subscriptionid"]'| base64 -d)"
+
+master_ip="$(kubectl --context=giantswarm-$cluster get no $cluster-master-$cluster-000000 -o yaml| yq -r '.status.addresses[] | select(.type == "InternalIP").address')"
```
tuommaki, line 24: I'm not sure if it's always guaranteed that master node has `-000000` suffix. Could be better to rely on `role=master` label?

tuommaki (approved):
Mostly fine. Few notes.

scripts/migrate_master_azs.sh
whites11: thanks, done

scripts/migrate_master_azs.sh
fiunchinho: Can we at least use `jq` with `-o json` ? :sweat_smile:

scripts/migrate_master_azs.sh
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
+
+set -e
```
fiunchinho, line 3: Maybe it would be a good idea to use `pipefail` so the script doesn't continue if some `yq` parsing fail.

scripts/migrate_master_azs.sh
whites11: the only reason this PR exists is because Tobi asked for it. It's really meant to be run once by myself only sweat_smile

tuommaki: > Can we at least use jq with -o json ? ￼

I guess. I confess that I use that instead of JSONPath w/ adhoc commands. The JSONPath would just look better in scripts as it doesn't strictly require any additional tooling, but then again, the `az` don't have that support so `jq` is pretty mandatory anyway.

fiunchinho (approved):

scripts/migrate_master_azs.sh
whites11: done

scripts/migrate_master_azs.sh
whites11: switched to jq, thanks for the JSONPath hint, will try to remember for next time

fiunchinho:
https://twitter.com/iamdevloper/status/397664295875805184

PR #1336, opened 2021-02-02
Use latest CAPZ fork

axbarsan:
Let's actually be able to use spot instances ✅


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "azure-operator"
 	source      string = "https://github.com/giantswarm/azure-operator"
-	version            = "5.3.1-dev"
+	version            = "5.3.1-spot"
```
axbarsan, line 8: Remove this!

service/controller/key/key.go
```diff
@@ -727,12 +727,16 @@ func WorkersSubnetCIDR(customObject providerv1alpha1.AzureConfig) string {
...
+	return azureMachinePool.Spec.Template.SpotVMOptions != nil
+}
+
 func NodePoolSpotInstancesMaxPrice(azureMachinePool *expcapzv1alpha3.AzureMachinePool) string {
 	if azureMachinePool.Spec.Template.SpotVMOptions == nil || azureMachinePool.Spec.Template.SpotVMOptions.MaxPrice == nil {
 		return ""
 	}
 
-	return *azureMachinePool.Spec.Template.SpotVMOptions.MaxPrice
+	return azureMachinePool.Spec.Template.SpotVMOptions.MaxPrice.AsDec().String()
```
whites11, line 739: What does AsDec do? It feels like it's extracting a decimal and not a floating point number

axbarsan: `AsDec()` returns the value as a [`*inf.Dec`, which is a format used for "infinite precision" decimals](https://github.com/go-inf/inf/blob/v0.9.1/dec.go).

Serializing the `inf.Dec` type into a string basically creates a [stringified float value.](https://github.com/go-inf/inf/blob/d2d2541c53f18d2a059457998ce2876cc8e67cbf/dec.go#L50)

The reason for converting the value to `inf.Dec`, is that when serializing the `resource.Quantity` type into a string, it uses its own serialization format (e.g. `0.1` is `100m`)

This is also the logic in [upstream CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8df269291bae09a1873fa13619c03617e0d7e71f/cloud/converters/spotinstances.go#L35)

Btw, this branch was already tested yesterday, and everything worked smoothly 👍🏻

fiunchinho (approved):

tuommaki (approved):

PR #1337, opened 2021-02-02
AzureMachine Creating and Upgrading conditions

nprokopic:
Closes https://github.com/giantswarm/giantswarm/issues/15430

What's in the box:

- Updating `release.giantswarm.io/last-deployed-version` on `AzureMachine` CR when the control plane creation or upgrade is done.
- Setting AzureMachine Creating and Upgrading conditions. Existing condition handlers Creating and Upgrading (that are already used for Cluster and MachinePool) are used.


service/controller/azuremachine/handler/azuremachinemetadata/create.go
```diff
@@ -0,0 +1,41 @@
...
+	// We need this for existing clusters, and we do it both in Cluster
+	// controller (clusterupgrade) and in AzureMachine controller, because we
+	// do not know which controller/handler will be executed first, and we need
+	// annotation set correctly in both places.
+	err = helpers.InitAzureMachineAnnotations(ctx, r.ctrlClient, r.logger, &azureMachine)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = r.ctrlClient.Update(ctx, &azureMachine)
```
whites11, line 29: So basically we update the cr at every reconciliation loop, right?
Even if there are no changes being made.
If the update call makes a change in the timestamps we're going to end up with an infinite loop maybe?

nprokopic: If there were no actual changes in the CR, nothing should happen, i.e. update call would have no effect. We have code like this in many places in azure-operator.

whites11: good

whites11 (approved):
lgtm

service/controller/azuremachine/handler/azuremachinemetadata/create.go
```diff
@@ -0,0 +1,41 @@
...
+	if errors.IsConflict(err) {
+		r.logger.Debugf(ctx, "conflict trying to save object in k8s API concurrently", "stack", microerror.JSON(microerror.Mask(err)))
```
tuommaki, line 31: Yo! That thingie doesn't work as expected. :grimacing: 

I'd propose to start discussion on adding `Warningf()` here: https://github.com/giantswarm/micrologger/blob/master/spec.go#L7 to get the desired outcome. :slightly_smiling_face:

PR #1338, opened 2021-02-03
Remove Cluster Ready condition dependency from Cluster Upgrading condition

nprokopic:
Closes https://github.com/giantswarm/giantswarm/issues/15252

Motivation:

Upgrading condition handler is simply comparing `release.giantswarm.io/version` label (desired release) and `release.giantswarm.io/last-deployed-version` annotation (last actually deployed release), so it is not checking Cluster `Ready` directly.
However, at the end of the upgrade process `release.giantswarm.io/last-deployed-version` annotation is updated only when Cluster `Ready` condition becomes True. Therefore Cluster `Ready` condition is sort of a transitive dependency for Cluster `Upgrading` condition.)

What's in the box:
- Do not check Cluster `Ready` condition when updating Cluster `release.giantswarm.io/last-deployed-version` annotation.
- Check AzureMachine (control plane) `Upgrading` condition when updating Cluster `release.giantswarm.io/last-deployed-version` annotation. This effectively means that Cluster `Upgrading` will now (transitively) depend on only AzureMachine (control plane) `Upgrading` and MachinePools `Upgrading`, which is much more reliable than the previous solution.


service/controller/cluster/handler/clusterreleaseversion/checkupgradeprogress.go
```diff
@@ -56,72 +92,52 @@ func (r *Resource) isUpgradeCompleted(ctx context.Context, cluster *capi.Cluster
...
-	// This is basically upgrade happy path.
-	clusterIsReadyAndUpgraded := becameReadyWhileUpgrading && allNodePoolsUpgraded
-	if clusterIsReadyAndUpgraded {
-		r.logger.Debugf(ctx, "cluster became ready during the upgrade, all cluster descendants are upgraded, upgrade completed")
-		return true, nil
+	lastDeployedReleaseVersion, ok := obj.GetAnnotations()[annotation.LastDeployedReleaseVersion]
+	if !ok {
+		// Control plane or a node pool is still not created. This should be
+		// caught above in Creating check, but let's err on the side of caution
+		// here.
```
fiunchinho, line 131: is this because the annotation is added when `MachinePool` is `Ready`?

nprokopic: The annotation would be added for the first time when the control plane or node pool is created, and in the case of MachinePool, I believe that it checks the nodes directly (not Ready condition).

If the annotation is not set, that case would be probably caught in the `IsCreatingTrue(obj)` check above, because if the creation is still in progress (Creating==True), the annotation is not set.

That being said, it could happen that we reorder the handlers, or due to some async nature, so that Creating condition would still not be set, so `IsCreatingTrue(...)` would return false (rare case, but not impossible). Hence this additional check if the annotation is actually set.

P.S. If the annotation is not set, the bellow `lastDeployedReleaseVersion != desiredClusterReleaseVersion` check would also return false, but I think I added this before so it's more explicit.

nprokopic: The story around cluster upgrades, labels and annotation used for versioning and upgrades, and also conditions around those things is due for a nice refactoring and making a nicer and cleaner story there, since now some parts are a bit weirdly entangled :)

nprokopic: This PR is also towards that untangling :)

fiunchinho (approved):
LGTM

PR #1339, opened 2021-02-03
Refactor upgrade process

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15484

this PR refactors the upgrade state machine for node pools by removing the manual draining process lead by the node operator and replacing it with automated draining thanks to azure termination events.


pkg/handler/nodes/instances.go
```diff
@@ -5,49 +5,23 @@ import (
...
 	azureresource "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
-	providerv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/microerror"
+	"sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
 
-	"github.com/giantswarm/azure-operator/v5/pkg/handler/nodes/scalestrategy"
 	"github.com/giantswarm/azure-operator/v5/service/controller/key"
 )
 
-func (r *Resource) AllInstances(ctx context.Context, customObject providerv1alpha1.AzureConfig, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string) ([]compute.VirtualMachineScaleSetVM, error) {
```
whites11, line 15: diff for this file is a mess, but I just moved some public functions to a different module

service/controller/azureconfig/handler/masters/instances.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/azure-operator/v5/service/controller/key"
+)
+
+func (r *Resource) allInstances(ctx context.Context, customObject providerv1alpha1.AzureConfig, deploymentNameFunc func(customObject providerv1alpha1.AzureConfig) string) ([]compute.VirtualMachineScaleSetVM, error) {
```
whites11, line 13: moved here from a different package because it's not used elsewhere

service/controller/azuremachinepool/handler/nodepool/create_scale_workers.go
```diff
@@ -139,42 +122,148 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
...
+		return nil, nil, microerror.Mask(err)
+	}
 
-	strategy := scalestrategy.Quick{}
-	err = r.ScaleVMSS(ctx, virtualMachineScaleSetsClient, key.ClusterID(&azureMachinePool), key.NodePoolVMSSName(&azureMachinePool), int64(desiredWorkerCount), strategy)
+	var oldInstances []compute.VirtualMachineScaleSetVM
+	var newInstances []compute.VirtualMachineScaleSetVM
+	{
+		for _, i := range allWorkerInstances {
+			old, err := r.isWorkerInstanceFromPreviousRelease(ctx, cluster, azureMachinePool.Name, i)
```
tuommaki, line 169: I think this part is still the same as earlier: Listing WC nodes for each VMSS instance.

It could be a good idea though to go with this implementation now and then refactor that separately so the change is clearer in diff.

tuommaki (approved):

service/controller/azuremachinepool/handler/nodepool/create_scale_workers.go
whites11: it's slightly changed but the main point is similar, yes.
I agree with a separate refactoring

PR #1346, opened 2021-02-08
Vnet peering

whites11:
This PR replaces VPN gateway with VNet peering


service/controller/azureconfig/handler/vnetpeering/create.go
```diff
@@ -0,0 +1,225 @@
+package vnetpeering
```
nprokopic, line 1: Can this be implemented under `azurecluster` controller? since `azureconfig` should go away soon

whites11: I guess it's possible, yeah. 
I just backported the old PR to the current master to avoid being too much behind when we'll finally merge this.
Will request a review and rework what's needed when we'll be ready with the SP update.

service/controller/azureconfig/handler/vnetpeering/resource.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	Name = "vnetpeering"
+)
+
+type Config struct {
+	ClientFactory          client.OrganizationFactory
+	CPAzureClientSet       *client.AzureClientSet
+	HostResourceGroup      string
+	HostVirtualNetworkName string
```
tuommaki, line 18: While here, should these be changed to `ManagementCluster` prefix (from `Host`)?

tuommaki (commented):
Seems good in general

service/controller/azurecluster/handler/azureclusterconditions/conditionvpngatewayready.go
```diff
@@ -15,107 +13,61 @@ import (
 )
 
 const (
-	vpnDeploymentName = "vpn-template"
```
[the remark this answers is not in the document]
whites11: thanks, renamed

service/controller/azureconfig/handler/vnetpeering/resource.go
whites11: renamed

fiunchinho (approved):

service/service.go
```diff
@@ -252,13 +255,38 @@ func New(config Config) (*Service, error) {
...
+			config.Viper.GetString(config.Flag.Service.Azure.ClientSecret),
+			mcTenant,
+		)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		// Need to set all Organization Credentials' tenant ID as aux tenant.
+		secretList := &v1.SecretList{}
+		err = k8sClient.CtrlClient().List(context.Background(), secretList, ctrl.MatchingLabels{label.App: "credentiald"})
```
tuommaki, line 273: Don't these secrets ever get added, deleted or modified? This implementation runs forever with the state that existed when the program started. Shouldn't these be loaded on-demand when needed, in order to have fresh configuration of current state? Now if one removes credentials from MC, the running `azure-operator` still keeps using them.

fiunchinho: I believe this force us to restart the operator every time a Organization with a new tenant is created. I think it's not the best solution, but taking into account that I don't think this happens a lot, and that eventually we will stop running azure-operator, I'd say it's ok-ish?

whites11: you are both right.
The current setup of azure operator creates one single azure client set with the MC credentials rather than a factory like it's done with the organization secrets.
It would be a huge refactoring to change how that works and I don't think it's worth the effort considering, as Jose said, that azure operator will eventually disappear.
I understand the need to restart azure operator when org secrets change to be an issue but:
- secret change is handled by SEs only
- this should be a pretty rare situation.

Are you ok with this compromise or do you want me to work on a proper solution?

tuommaki: I'm ok with leaving this now as-is. Let's hope we don't hit issues with this. A workaround to make this reliable could be to add a controller watching these specific secrets and when it notices that known list of credentials changes, it would `os.Exit(1)` in order to restart the operator.

tuommaki (approved):

service/service.go
whites11: yeah that was my thinking as well.
You know what? it's should be easy enough, I will give it a try

tuommaki: Just a wish: Can you submit it in a separate PR so it's easier to grasp? :hugs:

PR #1349, opened 2021-02-09
Azure cluster identity

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15597

This PR adds a new handler that reconciles `Secret` CRs owned by `credentiald` and creates the CAPZ types.


service/controller/credential/handler/azureclusteridentity/create.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func newSecretName(legacySecret corev1.Secret) string {
+	name := strings.TrimPrefix(legacySecret.Name, "credential-")
+
+	return fmt.Sprintf("org-credential-%s", name)
+}
+
+func newSecretNamespace(legacySecret corev1.Secret) string {
+	return legacySecret.Namespace
```
whites11, line 183: Not sure about this. Maybe the secret and CR should be created in the org namespace?

service/controller/credential/handler/azureclusteridentity/create.go
```diff
@@ -0,0 +1,184 @@
...
+
+	desiredSpec := v1alpha3.AzureClusterIdentitySpec{
+		Type:     v1alpha3.ServicePrincipal,
+		ClientID: clientID,
+		ClientSecret: corev1.SecretReference{
+			Name:      newName,
+			Namespace: newNamespace,
+		},
+		TenantID:          tenantID,
+		AllowedNamespaces: make([]string, 0),
```
whites11, line 121: This means the AzureClusterIdentity can be used by any AzureCluster in any namespace.
Should we maybe restrict to the org namespace?

service/controller/credential/controller.go
```diff
@@ -0,0 +1,125 @@
...
+			// Name is used to compute finalizer names. This results in something
+			// like operatorkit.giantswarm.io/azure-operator-credential-controller.
+			Name: project.Name() + "-credential-controller",
+			NewRuntimeObjectFunc: func() runtime.Object {
+				return new(v1.Secret)
+			},
+			Resources: resources,
+			Selector: labels.SelectorFromSet(map[string]string{
+				label.App: "credentiald",
+			}),
```
tuommaki, line 72: I'd drop the separate controller here and just run the handler from `AzureClusterController` as part of the general reconciliation, before executing the upcoming handler that patches `AzureClusterIdentity` into it.

This implementation now causes the actual live race between operators because there's not version matching here (and there shouldn't be) so all operators jump in as soon as they can to update the related CRs. There is also potential for bigger problem in case we would ever implement any kind of garbage collection. The older operator possibly running would drop the finalizer immediately while the newer operator would expect it to be kept until dependent subresources are deleted. This would be very wrong.

The functionality is very simple in the end so it's fine to be part of `AzureClusterController` IMO.

whites11: OK, agreed will move it

whites11: done. What's missing now is the cleanup of `AzureCluterIdentity` objects that have been created from a legacy secret that was deleted. How do you suggest doing it?

tuommaki: I'd do something roughly like this (context: handler reconciling `AzureCluster`):
- Compute `legacySecretName` from referenced `AzureClusterIdentity`
- Find if `legacySecretName` still exists.
- If not, list all `AzureClusters` and check if given `AzureClusterIdentity` is referenced from there.
- If not, delete the given `AzureClusterIdentity` and set `AzureCluster` CR `IdentityRef` to nil.

This way we should minimize the chance that some other cluster is about-to-use the given to-be-deleted `AzureClusterIdentity`.

service/controller/azurecluster/handler/azureclusteridentity/create.go
```diff
@@ -0,0 +1,198 @@
...
+		err = r.ensureNewSecret(ctx, legacySecret)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		err = r.ensureAzureClusterIdentity(ctx, legacySecret)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
```
tuommaki, line 43: Instead of always processing all credentials, would it make sense to only process the ones that are in use? More specifically, only ensure `AzureClusterIdentity` for the credential that reconciled CR is actually using? This would reduce number of unnecessary API calls from the k8s apiserver. In present state this is listing all credentials and invoking two separate GETs for each legacy credential, performed on every CRs reconciliation loop.

This could be roughly something like following:
- Check if reconciled `AzureCluster` has `IdentityRef` filled. If yes, bail out early (no further actions).
- When `IdentityRef` is not populated, check if `AzureClusterIdentity` for given legacy credential exists.
- If not, proceed ensuring that related new secret & azure cluster identity CR exists.
- Update reconciled CR's `IdentityRef` with given `AzureClusteridentity`.

WDYT?

fiunchinho: That sounds like a good approach.

fiunchinho: Doing that would already solve https://github.com/giantswarm/giantswarm/issues/15576

whites11: on it

service/controller/credential/controller.go
whites11: done

service/controller/azurecluster/handler/azureclusteridentity/create.go
whites11: done

client/organization_factory.go
```diff
@@ -26,6 +26,7 @@ const (
 )
 
 type Interface interface {
+	GetCredentialSecret(ctx context.Context, objectMeta v1.ObjectMeta) (*v1alpha1.CredentialSecret, error)
```
fiunchinho, line 29: image

service/controller/azurecluster/handler/azureclusteridentity/create.go
```diff
@@ -0,0 +1,255 @@
...
+	var err error
+	azureCluster, err := key.ToAzureCluster(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Retrieve the legacy secret related to the organization this AzureCluster belongs to.
+	var legacySecret corev1.Secret
+	{
+		credentialSecret, err := r.azureClientsFactory.GetCredentialSecret(ctx, azureCluster.ObjectMeta)
```
fiunchinho, line 39: how much stuff would break if we'd change `GetCredentialSecret` to return a `Secret`? it feels weird to _get credential secret then find that secret_.

fiunchinho:
how is the life cycle of `AzureIdentities` resources? Can they be actually updated? If you want to change the client id or tenant id, shouldn't you create a new `AzureIdentity`?

service/controller/azurecluster/handler/azureclusteridentity/create.go
whites11: should be doable I guess. Or we can create a new function in the interface just for this considering the whole authentication thing will change soonish when we'll work on https://github.com/giantswarm/giantswarm/issues/15602

whites11:
> how is the life cycle of `AzureIdentities` resources? Can they be actually updated? If you want to change the client id or tenant id, shouldn't you create a new `AzureIdentity`?

I decided to ignore these kind of problems and rely on admission controllers to check that. WDYT?

fiunchinho:
> > how is the life cycle of `AzureIdentities` resources? Can they be actually updated? If you want to change the client id or tenant id, shouldn't you create a new `AzureIdentity`?
> 
> I decided to ignore these kind of problems and rely on admission controllers to check that. WDYT?

I asked because if we think about `AzureIdentity` as immutable, then we don't need to _keep them up to date_, which would make this simpler.

whites11:
> > > how is the life cycle of `AzureIdentities` resources? Can they be actually updated? If you want to change the client id or tenant id, shouldn't you create a new `AzureIdentity`?
> > 
> > 
> > I decided to ignore these kind of problems and rely on admission controllers to check that. WDYT?
> 
> I asked because if we think about `AzureIdentity` as immutable, then we don't need to _keep them up to date_, which would make this simpler.

It actually makes a lot of sense. @tuommaki ?

service/controller/azurecluster/handler/azureclusteridentity/create.go
whites11: I took a look at this. I agree with you this would be a nice improvement, but since it would require a lot of changes I was thinking about leaving it as it is right now, and do a refactoring when we'll change the whole thing for ticket #15602.
Does it work for you or is this a blocker in your opinion?

fiunchinho: Works for me.

tuommaki (approved):
Looks fine from my point of view. :+1:

tuommaki:
> > > > how is the life cycle of `AzureIdentities` resources? Can they be actually updated? If you want to change the client id or tenant id, shouldn't you create a new `AzureIdentity`?
> > > 
> > > 
> > > I decided to ignore these kind of problems and rely on admission controllers to check that. WDYT?
> > 
> > 
> > I asked because if we think about `AzureIdentity` as immutable, then we don't need to _keep them up to date_, which would make this simpler.
> 
> It actually makes a lot of sense. @tuommaki ?

Before reading through code I didn't even consider updating `AzureClusterIdentity` CRs from our credentials, but then I realized that as long as we have these GS style clusters accompanied UX, we must support that - otherwise it becomes very error prone to manage these, e.g. when SP expires. Therefore I think that this update logic is good to be in-place. I don't see any downsides with that for now.

PR #1353, opened 2021-02-11
Reference Spark CR as boostrap Ref from the MachinePool CR

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15570

### DIFF
--- a/service/controller/azureconfig/handler/capzcrs/create.go
+++ b/service/controller/azureconfig/handler/capzcrs/create.go
@@ -2,6 +2,7 @@ package capzcrs
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"strconv"
 
@@ -175,6 +176,16 @@ func (r *Resource) mapAzureConfigToAzureCluster(ctx context.Context, cr provider
 			},
 			ResourceGroup: key.ClusterID(&cr),
 			NetworkSpec: capzv1alpha3.NetworkSpec{
+				APIServerLB: capzv1alpha3.LoadBalancerSpec{
+					Name: fmt.Sprintf("%s-%s-%s", cr.Name, "API", "PublicLoadBalancer"),
+					SKU:  "Standard",
+					Type: "Public",
+					FrontendIPs: []capzv1alpha3.FrontendIP{
+						{
+							Name: fmt.Sprintf("%s-%s-%s-%s", cr.Name, "API", "PublicLoadBalancer", "Frontend"),
+						},
+					},
+				},
 				Vnet: capzv1alpha3.VnetSpec{
 					CIDRBlocks:    []string{key.VnetCIDR(cr)},
 					Name:          key.VnetName(cr),
@@ -220,8 +231,9 @@ func (r *Resource) mapAzureConfigToAzureMachine(ctx context.Context, cr provider
 				},
 			},
 			OSDisk: capzv1alpha3.OSDisk{
-				OSType:     "Linux",
-				DiskSizeGB: int32(50),
+				OSType:      "Linux",
+				CachingType: "ReadWrite",
+				DiskSizeGB:  int32(50),
 				ManagedDisk: capzv1alpha3.ManagedDisk{
 					StorageAccountType: "Premium_LRS",
 				},
--- a/service/controller/azureconfig/handler/capzcrs/create.go
+++ b/service/controller/azureconfig/handler/capzcrs/create.go
@@ -2,7 +2,6 @@ package capzcrs
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 	"strconv"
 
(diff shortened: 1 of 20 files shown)


service/controller/azureconfig/handler/workermigration/create.go
```diff
@@ -379,6 +379,20 @@ func (r *Resource) ensureMachinePoolExists(ctx context.Context, cr providerv1alp
 		return expcapiv1alpha3.MachinePool{}, microerror.Mask(err)
 	}
 
+	var bootstrapCRRef *corev1.ObjectReference
+	{
+		s := runtime.NewScheme()
+		err := corev1alpha1.AddToScheme(s)
+		if err != nil {
+			panic(fmt.Sprintf("corev1alpha1.AddToScheme: %+v", err))
+		}
```
fiunchinho, line 388: is there a way to get the `Scheme` from the controller-runtime `client` instance that we have?

service/controller/azureconfig/handler/capzcrs/create.go
```diff
@@ -175,6 +176,16 @@ func (r *Resource) mapAzureConfigToAzureCluster(ctx context.Context, cr provider
 			},
 			ResourceGroup: key.ClusterID(&cr),
 			NetworkSpec: capzv1alpha3.NetworkSpec{
+				APIServerLB: capzv1alpha3.LoadBalancerSpec{
+					Name: fmt.Sprintf("%s-%s-%s", cr.Name, "API", "PublicLoadBalancer"),
```
fiunchinho, line 180: Not sure if they are used anywhere else, but what do you think about creating functions in the `key` package that return this string (and the `FrontendIPs` name too)?

fiunchinho (changes requested):
IIUC the handler that we are updating is the one creating the CAPI/CAPZ CR's from `AzureConfig`. But some clusters already went through that process before. Will they pick up the new field if they were already created by the handler? I believe there is a function that detects if the `AzureCluster` CR needs to be updated or not, which is not taking this field into consideration so it wouldn't be added to already existing CR's.

Maybe we can cover that scenario in the tests, because I believe we already have cases for updating CR's, like `case 5: Ensure that AzureCluster partial VNET gets updated`.

service/controller/azureconfig/handler/capzcrs/create.go
nprokopic: These are GS-specific requirements currently, would it be possible to have these `APIServerLB` fields in admission controller create defaulting? And then also have those values validated in admission controller, and block change/update of these values?
This would be in addition to what's done here, as the mapping is needed for existing CRs, but the mapping will maybe be gone soon-ish.

WDYT?

axbarsan:
> IIUC the handler that we are updating is the one creating the CAPI/CAPZ CR's from `AzureConfig`. But some clusters already went through that process before.

What I intended to target is clusters that are upgrading from a pre-nodepools version to a nodepools version. So the `AzureConfig` would be discarded, and `AzureCluster`/`AzureMachine` and other friends would be created. These new resources need to have the correct structure, to pass CAPI/CAPZ validation.

There is also the option to default on the admission controller level. But CAPI/CAPZ is already defaulting these fields to different values from ours. I wanted to make my life easier and create the resources with the right values from the start

fiunchinho:
> > IIUC the handler that we are updating is the one creating the CAPI/CAPZ CR's from `AzureConfig`. But some clusters already went through that process before.
> 
> What I intended to target is clusters that are upgrading from a pre-nodepools version to a nodepools version. So the `AzureConfig` would be discarded, and `AzureCluster`/`AzureMachine` and other friends would be created. These new resources need to have the correct structure, to pass CAPI/CAPZ validation.
> 
> There is also the option to default on the admission controller level. But CAPI/CAPZ is already defaulting these fields to different values from ours. I wanted to make my life easier and create the resources with the right values from the start

But we need all CRs to pass the validation when they are updated in the future. Why not update all of them?

axbarsan:
> > > IIUC the handler that we are updating is the one creating the CAPI/CAPZ CR's from `AzureConfig`. But some clusters already went through that process before.
> > 
> > 
> > What I intended to target is clusters that are upgrading from a pre-nodepools version to a nodepools version. So the `AzureConfig` would be discarded, and `AzureCluster`/`AzureMachine` and other friends would be created. These new resources need to have the correct structure, to pass CAPI/CAPZ validation.
> > There is also the option to default on the admission controller level. But CAPI/CAPZ is already defaulting these fields to different values from ours. I wanted to make my life easier and create the resources with the right values from the start
> 
> But we need all CRs to pass the validation when they are updated in the future. Why not update all of them?

Do you mean existing `AzureCluster`/etc.? I've already updated those manually 😅

axbarsan:
I updated the PR to just include the reference to the boostrap `Spark` CR, since all the other defaulting is now done in the admission controller.

service/controller/azureconfig/handler/workermigration/create.go
axbarsan: I haven't been able to get the scheme from the actual client, but I was able to at least share the same scheme between the 2 references.

tuommaki (approved):

fiunchinho (approved):

PR #1354, opened 2021-02-11
Ensure `AzureCluster` CR has the `SubscriptionID` field set.

whites11:

this PR adds the migration logic needed to populate the `SubscriptionID` field of the `AzureCluster` CR based on the legacy organization secret

### DIFF
--- a/service/controller/azurecluster/handler/azureclusteridentity/create.go
+++ b/service/controller/azurecluster/handler/azureclusteridentity/create.go
@@ -19,9 +19,10 @@ import (
 const (
 	secretDataFieldName = "clientSecret"
 
-	legacySecretClientIDFieldName     = "azure.azureoperator.clientid"
-	legacySecretClientSecretFieldName = "azure.azureoperator.clientsecret"
-	legacySecretTenantIDFieldName     = "azure.azureoperator.tenantid"
+	legacySecretClientIDFieldName       = "azure.azureoperator.clientid"
+	legacySecretClientSecretFieldName   = "azure.azureoperator.clientsecret"
+	legacySecretSubscriptionIDFieldName = "azure.azureoperator.subscriptionid"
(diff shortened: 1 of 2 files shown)


service/controller/azurecluster/handler/azureclusteridentity/create.go
```diff
@@ -121,7 +133,8 @@ func (r *Resource) ensureNewSecret(ctx context.Context, legacySecret corev1.Secr
 				Name:      newName,
 				Namespace: newNamespace,
 				Labels: map[string]string{
-					apiextensionslabels.ManagedBy: project.Name(),
+					apiextensionslabels.ManagedBy:    project.Name(),
+					apiextensionslabels.Organization: legacySecret.GetLabels()[apiextensionslabels.Organization],
```
whites11, line 137: I reckon the organization label can still be useful to us and it doesn't hurt

fiunchinho (approved):
What'd be the UX for this? I guess `kubectl-gs` can populate the subscription id field in `AzureCluster` depending on the `--organization`/`--owner` parameter?

service/controller/azurecluster/handler/azureclusteridentity/create.go
```diff
@@ -99,6 +100,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
+	// Check if the AzureCluster CR has a SubscriptionID set.
+	if azureCluster.Spec.SubscriptionID == "" {
```
fiunchinho, line 104: In next PR's when we will use this field, we'd need to cancel reconciliation until this field is set.

whites11:
> What'd be the UX for this? I guess `kubectl-gs` can populate the subscription id field in `AzureCluster` depending on the `--organization`/`--owner` parameter?

yeah or the admission controller could I guess

PR #1357, opened 2021-02-12
Capz auth refactoring

whites11:

this PR uses `AzureClusterIdentity` and related CRs to initialize the Azure API clients.

Current problems still unaddressed:

- `PartnerID` field is still not managed (where to store it?)
- At cluster creation, the first few loops fail because the `SubscriptionID` and `identityRef` are not set in the `AzureCluster`. It eventually settles but there's a good chance of being alerted for `ErrorRateTooHigh`.
- Not sure if the multi SP management is correct and not sure if we can implement it the right way with upstream types only.

Looking for early feedback and ideas


pkg/credential/credential.go
```diff
@@ -39,53 +36,56 @@ func NewK8SCredentialProvider(k8sclient k8sclient.Interface, gsTenantID string,
...
+	if err != nil {
+		return auth.ClientCredentialsConfig{}, "", "", microerror.Mask(err)
 	}
 
 	// The tenant cluster resources will belong to a subscription that belongs to a different Tenant ID than the one used for authentication.
-	k.logger.Debugf(ctx, "Azure subscription %#q belongs to the tenant ID %#q which is different than the Tenant ID %#q that owns the Service Principal. Using multi tenant authentication", subscriptionID, tenantID, k.gsTenantID)
-	credentials := auth.NewClientCredentialsConfig(clientID, clientSecret, k.gsTenantID)
-	credentials.AuxTenants = append(credentials.AuxTenants, tenantID)
+	credentials := auth.NewClientCredentialsConfig(clientID, clientSecret, tenantID)
+	if tenantID != k.gsTenantID {
```
whites11, line 85: This is the problematic bit. If the "SP tenant ID" (passed as a parameter to azure operator and defined in `installations` repo) and the tenant ID stored in the secret (let's call it BYOC tenant) are different, I always try to add one of the two as `AuxTenants`. I have zero idea what that means and if it can fail and why

tuommaki: AFAICS this PR in current setting doesn't yet decouple Azure client factory (`client` package) from credentials factory (`credentials` package). If that was done and they were integrated with the specific credential struct I proposed earlier, could you add this lost field in that one?

whites11: didn't do the decoupling yet. Plus I expanded my knowledge about the tenant issue and I'd like to brainstorm a solution for that.

PR #1365, opened 2021-02-16
Don't panic on unallocated subnet entries

tuommaki:
Don't panic if AzureCluster CR has a subnet without defined CIDBlocks.
While this is not necessarily strictly valid in all use cases, the code
should still handle the situation gracefully.


tuommaki:
I'm not 100% sure if these cover all problems I noticed today, but at least these will bring us one step further.

whites11 (approved):
Seems legit

PR #1367, opened 2021-02-18
avoid saving the azuremachine CR too often

whites11:
This handler was calling `Update` on the `AzureMachine` CR at every reconciliation loop, regardless of any changes being made to the CR.
This triggered a loop in the handler that was running in an infinite sequence.
This PR check for changes in the CR and avoid saving if there isn't any

### DIFF
--- a/service/controller/azuremachine/handler/azuremachinemetadata/create.go
+++ b/service/controller/azuremachine/handler/azuremachinemetadata/create.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/giantswarm/microerror"
-	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/giantswarm/azure-operator/v5/pkg/helpers"
 	"github.com/giantswarm/azure-operator/v5/service/controller/key"
@@ -23,15 +22,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, cr interface{}) error {
 	// annotation set correctly in both places.
 	err = helpers.InitAzureMachineAnnotations(ctx, r.ctrlClient, r.logger, &azureMachine)
 	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	err = r.ctrlClient.Update(ctx, &azureMachine)
-	if errors.IsConflict(err) {
-		r.logger.Debugf(ctx, "conflict trying to save object in k8s API concurrently", "stack", microerror.JSON(microerror.Mask(err)))
-		r.logger.Debugf(ctx, "cancelling resource")
-		return nil
-	} else if err != nil {
 		return microerror.Mask(err)
 	}
 
--- a/service/controller/azuremachine/handler/azuremachinemetadata/create.go
+++ b/service/controller/azuremachine/handler/azuremachinemetadata/create.go
@@ -3,7 +3,9 @@ package azuremachinemetadata
 import (
 	"context"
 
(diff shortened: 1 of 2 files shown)


nprokopic:
Are you sure this is triggering new reconciliation loops? We were checking a case like this before, and it should not happen, as update should basically be a noop if there were no changes.

If it is triggering new reconciliation loops without CR being actually changed, it might be a bug in operatorkit cc @giantswarm/sig-operator

nprokopic:
> If it is triggering new reconciliation loops without CR being actually changed, it might be a bug in operatorkit cc @giantswarm/sig-operator

I might be wrong about operatorkit bug, but we did have few discussions about the topic and it came down to these updates without actual changes not having any effect.

whites11:
> Are you sure this is triggering new reconciliation loops? We were checking a case like this before, and it should not happen, as update should basically be a noop if there were no changes.
> 
> If it is triggering new reconciliation loops without CR being actually changed, it might be a bug in operatorkit cc @giantswarm/sig-operator

100% sure I can provide CRs and logs

nprokopic:
There might be some issue in the func called few lines above then:

```
err = helpers.InitAzureMachineAnnotations(ctx, r.ctrlClient, r.logger, &azureMachine)
```

https://github.com/giantswarm/azure-operator/blob/master/pkg/helpers/azuremachine.go#L16

whites11:
> There might be some issue in the func called few lines above then:
> 
> ```
> err = helpers.InitAzureMachineAnnotations(ctx, r.ctrlClient, r.logger, &azureMachine)
> ```
> 
> https://github.com/giantswarm/azure-operator/blob/master/pkg/helpers/azuremachine.go#L16

BTW I copied what you did in another place where you use the same function

https://github.com/giantswarm/azure-operator/blob/master/service/controller/azurecluster/handler/azureclusterupgrade/resource.go#L69

nprokopic:
Touche :) Forgot now why I did that, it wasn't for avoiding reconciliation loops though.

My only concern here is that there might be an issue/bug somewhere in code that is causing infinite reconciliation loops (maybe caused by me in that func :grimacing:), and this might hide it temporarily.

nprokopic:
> and this might hide it temporarily.

Like I possibly did with that other "changed" check.

whites11:
> > and this might hide it temporarily.
> 
> Like I possibly did with that other "changed" check.

I'm not sure right now but I think that's the only piece of code where we actually edit the Azure machine cr. If that's true it has to be it causing the reconciliation madness

tuommaki (approved):
Nice :+1:

service/controller/azuremachine/handler/azuremachinemetadata/create.go
```diff
@@ -26,13 +30,18 @@ func (r *Resource) EnsureCreated(ctx context.Context, cr interface{}) error {
...
-		return nil
-	} else if err != nil {
-		return microerror.Mask(err)
+	_, annotationIsSet := azureMachine.Annotations[annotation.LastDeployedReleaseVersion]
+	changed = !annotationWasSet && annotationIsSet
+
+	if changed {
+		err = r.ctrlClient.Update(ctx, &azureMachine)
+		if errors.IsConflict(err) {
+			r.logger.Debugf(ctx, "conflict trying to save object in k8s API concurrently", "stack", microerror.JSON(microerror.Mask(err)))
```
fiunchinho, line 39: https://github.com/giantswarm/azure-operator/pull/1322

image

PR #1369, opened 2021-02-19
Avoid overriding the NP size when the scaling is changed by autoscaler.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15845

The `NodePool` handler compares the desired ARM deployment with the latest ARM deployment supplied to check if there are any changes that require the deployment to be updated.
One of the fields that is compared is the `scaling` field, that contains the min, max and current number of replicas for the NP.

When we first submit the deployment (new node pool) the "current" field is set to the same value as the "min" field.
From that moment on, if "min" != "max" (i.e. autoscaling is enabled) the "current" field is not reliable any more, because the autoscaler can change it any time.

That means that everytime the cluster autoscaler changed the size of a NP, azure operator considered the node pool deployment to be out of date and reapplied the deployment, basically fighting against with the cluster autoscaler.

This PR ignores the `current` field and only considers the `min` and `max` to decide if the deployment needs to be applied or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Avoid blocking the whole `AzureConfig` handler on cluster creation because we can't update the `StorageClasses`.
+- Avoid overriding the NP size when the scaling is changed by autoscaler. 
 
 ## [5.4.0] - 2021-02-05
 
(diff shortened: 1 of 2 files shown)


fiunchinho:
When autoscaler is not enabled, we use the `min` as `current` anyway, so this PR won't change the behavior there, right?

fiunchinho (approved):

whites11:
> When autoscaler is not enabled, we use the `min` as `current` anyway, so this PR won't change the behavior there, right?

the first creation of the deployment is identical.
if we change the min AND max size, the change is picked up.
I don't see why it shouldn't work, yes

PR #1370, opened 2021-02-21
Ensure node pools min size is applied immediately when changed.

whites11:
Because of how the cluster autoscaler works (see https://kubernetes.slack.com/archives/C09R1LV8S/p1613748423162500) there is an unexpected (in my opinion) behaviour in our current implementation of autoscaling.

Example:

- create a cluster with one node pool with min: 3 max: 10
- cluster comes up with 3 workers in the node pool
- scale the node pool to min: 5 max: 10

At this point, one would expect the node pool to scale to 5 workers. That's not going to happen until the cluster autoscaler detects some unschedulable pods.

this PR suggests an easy change in the azure operator, that ensures the current size of the node pool VMSS is always respecting the min size set for autoscaling when the min or max size are changed. 
In my honest opinion this is a much better behaviour from the customer point of view.

No strong opinions on this, if you don't like it I'm ok closing this PR.


service/controller/azuremachinepool/handler/nodepool/deployment.go
```diff
@@ -61,7 +61,9 @@ func (r Resource) getDesiredDeployment(ctx context.Context, storageAccountsClien
 	if key.NodePoolMinReplicas(machinePool) != key.NodePoolMaxReplicas(machinePool) {
 		// Autoscaler is enabled. Will need to use the current number of replicas from the VMSS if it exists.
 		if !vmss.IsHTTPStatus(404) {
-			currentReplicas = int32(*vmss.Sku.Capacity)
+			if int32(*vmss.Sku.Capacity) > currentReplicas {
```
tuommaki, line 64:
```suggestion
			// Update VMSS desired number of nodes in case existing desired
			// minimum number of nodes is bigger than current desired number of
			// nodes by cluster auto-scaler. It doesn't automatically increase
			// the number of nodes to minimum level so it must be triggered via
			// API here.
			if int32(*vmss.Sku.Capacity) > currentReplicas {
```

tuommaki (approved):
Devil's always in the details. I'm fine with this, but feel that additional comment is necessary in the code as otherwise it's not obvious at first look.

fiunchinho (approved):

T-Kukawka (approved):

PR #1379, opened 2021-02-23
Fix upgrade

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15884

this PR addresses 2 problems:

1. upgrading node pools with zero replicas didn't ever finish
2. the upgrade conditions didn't work reliably for autoscaling node pools

### DIFF
--- a/service/controller/azuremachinepool/handler/nodepool/create_scale_workers.go
+++ b/service/controller/azuremachinepool/handler/nodepool/create_scale_workers.go
@@ -104,8 +104,7 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
 	r.Logger.Debugf(ctx, "The desired number of workers is: %d", desiredWorkerCount)
 
 	if desiredWorkerCount == 0 {
-		// The node pool is empty, the upgrade process is not finished.
-
+		// The node pool is empty, the upgrade process can stop here.
 		r.Logger.Debugf(ctx, "The size of the Node Pool is 0: upgrade is complete")
 		return DeploymentUninitialized, nil
 	}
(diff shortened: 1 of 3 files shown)


service/controller/machinepool/handler/machinepoolupgrade/lastdeployedreleaseversion.go
```diff
@@ -136,10 +136,10 @@ func allNodePoolNodesUpToDate(ctx context.Context, tenantClusterClient client.Cl
 
 	// azure-admission-controller ensures that machinePool.Spec.Replicas is
 	// always set
-	requiredReplicas := *machinePool.Spec.Replicas
+	minRequiredReplicas := key.NodePoolMinReplicas(machinePool)
```
nprokopic, line 139: Can cluster-autoscaler be completely not enabled, so that these annotations are missing (or whatever is the behavior when it's generally disabled, not only during upgrades)? In that case it would fall back to `machinePool.Spec.Replicas`.

whites11: ensured by the azure admission controller https://github.com/giantswarm/azure-admission-controller/blob/e0695e1d24a783c9dab971efff8d4490fa55c29e/pkg/machinepool/autoscaling.go#L24

nprokopic (approved):

service/controller/azuremachinepool/handler/nodepool/create_scale_workers.go
```diff
@@ -103,6 +103,13 @@ func (r *Resource) scaleUpWorkerVMSSTransition(ctx context.Context, obj interfac
 	desiredWorkerCount := int64(len(oldInstances) * 2)
 	r.Logger.Debugf(ctx, "The desired number of workers is: %d", desiredWorkerCount)
 
+	if desiredWorkerCount == 0 {
+		// The node pool is empty, the upgrade process is not finished.
```
fiunchinho, line 107: Comment seems to disagree with logged line.

whites11: not == now. HOW DON'T YOU GET THAT :P

whites11: fixed

fiunchinho (approved):

PR #1382, opened 2021-02-25
Added new factory for azure client supporting both the WC and the MC

whites11:

This PR only adds the new interfaces and implementation for the new factories needed to capz authentication.

The factories are not used anywhere in this PR.


azureclient/credentialprovider/cli_flags.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/giantswarm/micrologger"
+	"sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type CLIFlagsCredentialProviderConfig struct {
+	CtrlClient ctrl.Client
+	Logger     micrologger.Logger
+
+	ManagementClusterClientID       string
```
whites11, line 19: I am not sure if, instead of passing values here, I should pass the `flag` object and do the parsing of cli arguments inside this module. WDYT?

azureclient/credentialprovider/error.go
```diff
@@ -0,0 +1,67 @@
...
+
+func IsApplicationNotFoundInADError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	{
+		dErr, ok := err.(autorest.DetailedError)
+		if ok {
+			if dErr.PackageType == "azure.multiTenantSPTAuthorizer" && dErr.Method == "WithAuthorization" && strings.HasPrefix(dErr.Message, "Failed to refresh one or more Tokens for request to") {
```
whites11, line 60: If you have better ideas to catch this error I'm all ears

azureclient/basicfactory/spec.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage"
+
+	"github.com/giantswarm/azure-operator/v5/azureclient/credentialprovider"
+)
+
+// The basicfactory.Interface interface defines an interface to return an Azure API Client given the
+// authentication data provided via a AzureClientCredentialsConfig object.
+// Implementations of this interface are not meant to be run standalone, but to be used by a credentialawarefactory.Interface.
+type Interface interface {
```
tuommaki, line 19: What if the name of this package was `basic` and this interface was `Factory` - then it would be `basic.Factory` under `azureclient`. WDYT?

azureclient/credentialprovider/cli_flags.go
```diff
@@ -0,0 +1,133 @@
...
+	ManagementClusterClientSecret   string
+	ManagementClusterSubscriptionID string
+	ManagementClusterTenantID       string
+}
+
+type CLIFlagsCredentialProvider struct {
```
tuommaki, line 28: Is this implementation actually tied to CLI flags in any way?

Could this be `ManagementClusterCredentialProvider` ?

azureclient/credentialprovider/error.go
tuommaki: Do those errors possibly have any kind of error code to match? If not, I guess this is best we can do, unless SDK has some kind of matcher somewhere in the API (I think there are at least some).

azureclient/credentialsawarefactory/spec.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-11-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage"
+	"github.com/giantswarm/apiextensions/v3/pkg/apis/provider/v1alpha1"
+)
+
+// The credentialsawarefactory.Interface interface defines an interface for retrieving an Azure API Client
+// Based on the Giantswarm Cluster ID. Based on the implementation, it can return Azure API Clients for the
+// Management cluster of the Workload Cluster.
+type Interface interface {
```
tuommaki, line 17: Like with `basic.Factory`, what if this package was `credentialsaware` and interface `Factory`?

tuommaki (commented):
In general the code looks fine, but maybe the naming could be slightly adjusted?

azureclient/basicfactory/spec.go
whites11: what about the default implementation? Now it's called BasicFactory how do you suggest to name it?

azureclient/credentialprovider/cli_flags.go
whites11: renamed

azureclient/credentialsawarefactory/spec.go
whites11: same as earlier, how would you name the implementation?

azureclient/credentialprovider/error.go
whites11: not really. the http status code is 0.
I believe this is a parsing error happening upstream.

azureclient/basicfactory/spec.go
tuommaki: Up to you. When there's an interface that is expected to be used, I'd just create some `New()` function which constructs it and the implementation can be anything. `basicFactory` or `factory` maybe?

whites11: ok makes sense. Ensured I use the interface as the return method and renamed to `factory`

tuommaki (approved):

PR #1409, opened 2021-04-06
Upgrade operatorkit & clean up old v2 references

tuommaki:
Latest OperatorKit `v4.3.1` received an update to drop use of self-link
which was deprecated in k8s 1.20.

Old references to v2 of OperatorKit were also cleaned up in this change.


flag/service/service.go
```diff
@@ -1,7 +1,7 @@
 package service
 
 import (
-	"github.com/giantswarm/operatorkit/v2/pkg/flag/service/kubernetes"
+	"github.com/giantswarm/operatorkit/v4/pkg/flag/service/kubernetes"
```
tuommaki, line 4: For some reason `v2` was still used in few places. I updated these.

whites11 (approved):

fiunchinho (approved):

PR #1413, opened 2021-04-15
return error if credential is using single tenant SP

whites11:
This PR makes azure operator return an error if the customer is using the "old style" service principal.
It also adds all organization's tenant IDs as aux tenants in the management cluster's azure clients. 
This is needed by vnet peering call that needs access to both subscriptions at the same time


fiunchinho (commented):
remember the changelog. We need to check all current secrets are properly labeled.

whites11:
> remember the changelog. We need to check all current secrets are properly labeled.

yes we need to

tuommaki (approved):
Seems fine if it works for you

PR #1426, opened 2021-04-27
Safequick

whites11:
This PR aims at making upgrade of large node pools safer, by scaling up the node pools 5 nodes at a time during the double-the-nodepool-size- phase.
It makes upgrade procedure longer but it should be safer 429-wise

### DIFF
--- a/pkg/handler/nodes/scalestrategy/safequick.go
+++ b/pkg/handler/nodes/scalestrategy/safequick.go
@@ -1,0 +1,21 @@
+package scalestrategy
+
+type SafeQuick struct {
+}
+
+const (
+	safeThreshold = 5
+)
+
+func (i SafeQuick) GetNodeCount(currentCount int64, desiredCount int64) int64 {
+	// Cluster size decreased or unchanged.
+	if currentCount >= desiredCount {
+		return desiredCount
+	}
+
+	if desiredCount-currentCount > safeThreshold {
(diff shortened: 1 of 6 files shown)


pkg/handler/nodes/scalestrategy/safequick.go
```diff
@@ -0,0 +1,21 @@
+package scalestrategy
+
+type SafeQuick struct {
```
tuommaki, line 3: `scalestrategy.SafeQuick` doesn't express what it is, what it does or how it works. Could it be maybe `scalestrategy.Staircase` or `scalestrategy.Steps` or `scalestrategy.Stepper`?

tuommaki (approved):
LGTM when tested. Minor note on strategy naming.

pkg/handler/nodes/scalestrategy/safequick.go
whites11: good point, renamed

fiunchinho:
What did I miss? did we hit 429 again?

whites11:
> What did I miss? did we hit 429 again?

during GK production upgrade. Their node pool is 21 nodes and when we created 21 nodes at the same time we got 429.
It recovered quickly but it generated an outage because controller manager wasn't working and thus routes were not in place and the pods scheduled in new nodes couldn't reach some azure services

PR #1428, opened 2021-05-03
Only check k8s nodes to determine if upgrade finished

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/16966


whites11 (approved):
LGTMIIWFY

service/controller/azureconfig/handler/masters/create_master_instances_upgrading.go
```diff
@@ -128,6 +138,27 @@ func (r *Resource) masterInstancesUpgradingTransition(ctx context.Context, obj i
...
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	mastersVMSS, err := virtualMachineScaleSetsClient.Get(ctx, key.ClusterID(azureConfig), key.MasterVMSSName(*azureConfig))
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	azureOperatorVersionTag, ok := mastersVMSS.Tags["gs-azure-operator.giantswarm.io-version"]
```
whites11, line 154: Don't we have the label in a const anywhere?

fiunchinho: I looked for it but we haven't because we use it directly in the json template, not in Go code.

whites11: alright, LGTM

whites11 (approved):

PR #1430, opened 2021-05-03
When deleting a node pool, also delete the VMSS role assignment.

whites11:
When deleting a node pool, we did not delete the role assignment created to use managed identity.
That meant it was impossible to re-create a node pool with the same name (because role assignment creation used to fail).

This PR aims at solving this issue by deleting the role assignment when deleting a node pool

### DIFF
--- a/service/controller/azuremachinepool/handler/nodepool/delete.go
+++ b/service/controller/azuremachinepool/handler/nodepool/delete.go
@@ -142,12 +142,30 @@ func (r *Resource) deleteARMDeployment(ctx context.Context, azureMachinePool *ca
 
 // deleteVMSS deletes the VMSS from Azure.
 func (r *Resource) deleteVMSS(ctx context.Context, azureMachinePool *capzexpv1alpha3.AzureMachinePool, resourceGroupName, vmssName string) error {
-	r.Logger.LogCtx(ctx, "message", "Deleting machine pool VMSS")
-
 	virtualMachineScaleSetsClient, err := r.ClientFactory.GetVirtualMachineScaleSetsClient(ctx, azureMachinePool.ObjectMeta)
 	if err != nil {
 		return microerror.Mask(err)
 	}
+
+	// Delete role assignment related to this VMSS.
+	{
+		vmss, err := virtualMachineScaleSetsClient.Get(ctx, resourceGroupName, vmssName)
+		if IsNotFound(err) {
+			r.Logger.LogCtx(ctx, "message", "Machine pool VMSS was already deleted")
+			return nil
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if vmss.Identity.PrincipalID != nil {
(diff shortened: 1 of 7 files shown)


service/controller/azuremachinepool/handler/nodepool/delete.go
```diff
@@ -142,13 +142,31 @@ func (r *Resource) deleteARMDeployment(ctx context.Context, azureMachinePool *ca
...
 // deleteVMSS deletes the VMSS from Azure.
 func (r *Resource) deleteVMSS(ctx context.Context, azureMachinePool *capzexpv1alpha3.AzureMachinePool, resourceGroupName, vmssName string) error {
-	r.Logger.LogCtx(ctx, "message", "Deleting machine pool VMSS")
-
 	virtualMachineScaleSetsClient, err := r.ClientFactory.GetVirtualMachineScaleSetsClient(ctx, azureMachinePool.ObjectMeta)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
+	// Delete role assignment related to this VMSS.
```
fiunchinho, line 150: Are we allowed to remove the VMSS first and then the role assignment? It feels safer but not sure if it's possible.

whites11: It is. Role assignment has a reference to the vmss and not the opposite so it's fine to delete the role assignment first

service/controller/azuremachinepool/handler/nodepool/delete.go
```diff
@@ -161,3 +179,39 @@ func (r *Resource) deleteVMSS(ctx context.Context, azureMachinePool *capzexpv1al
...
+	}
+
+	scope := fmt.Sprintf("/subscriptions/%s/resourcegroups/%s", roleAssignmentsClient.SubscriptionID, key.ClusterID(azureMachinePool))
+
+	results, err := roleAssignmentsClient.ListForScope(ctx, scope, fmt.Sprintf("principalId eq '%s'", principalID))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.Logger.Debugf(ctx, "Found %d role assignments to delete", len(results.Values()))
```
tuommaki, line 199: This deletes all role assignments found from a given resource group. How does this work when there are three NPs in a cluster and user deletes one? Are those remaining ones re-created automatically or not needed at all?

whites11: Not true. `ListForScope` is called with a filter by principalID and the principal is unique by vmss.
Thus I'm deleting only one assignment

fiunchinho (approved):

service/controller/azuremachinepool/handler/nodepool/delete.go
tuommaki: Whoa!

Totally missed that.

PR #1432, opened 2021-05-10
Spot upgrades

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/17132

The goal of this PR is to make upgrades of clusters containing spot instances both more reliable and more graceful.
Problems of the previous approach:

1. If there is no capacity for spot instances, upgrade can be stuck for long time (until there is capacity)
2. spot instances nodes don't get drained before deletion.

Problem number 1 is solved by skipping the `WaitForWorkersToBecomeReady` step for spot instances NPs.
Problem number 2 is solved by using the [Simulate Eviction](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesetvms/simulateeviction) API call to delete spot instances. This way the `azure-scheduled-events` app receives a `Preempt` notification and at least tries to drain the nodes.

Unluckily, fix number 2 required me to bump the azure sdk because this is a recently added API call and with that came a few (simple) changes in various places in the code, making the PR bigger.

I also decided to merge 2 states of the upgrade process: `TerminateOldWorkerInstances` and `WaitForOldWorkersToBeGone`. The two steps were almost identical so it didn't make sense to keep them separate.


fiunchinho:
I'd say not draining spot instances before deletion is not a big deal, as they are spot instances.

whites11:
You convinced me Jose I will simplify this PR

PR #1433, opened 2021-05-11
Spot instances upgrades improvement

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/17132

The goal of this PR is to make upgrades of clusters containing spot instances more reliable.
Using the previous approach, if there is no capacity for spot instances, upgrade can be stuck for long time (until there is capacity)

To solve the problem, in this PR I am skipping the `WaitForWorkersToBecomeReady` step for spot instances NPs.

I also decided to merge 2 states of the upgrade process: `TerminateOldWorkerInstances` and `WaitForOldWorkersToBeGone`. The two steps were almost identical so it didn't make sense to keep them separate.


tuommaki (approved):
Seems good to me

fiunchinho (changes requested):
CHANGELOG and we are good to go

whites11:
There was a change log entry already

PR #1442, opened 2021-05-13
Check error using wrapped error

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/17180

In the `apimachinery` version that we currently depend on, `errors.Is*` functions use a switch/case to check error type. In release `v0.19.0` onwards it uses `errors.As`, which takes into account the wrapped error. Until we upgrade to this release, we need these changes to actually check using the wrapped error.


whites11 (approved):
Thanks a lot

PR #1448, opened 2021-05-13
Create AzureClusterIdentity on the cluster namespace

fiunchinho:
Currently azure-operator creates an `AzureClusterIdentity` (and a secret for it) when `AzureCluster` has no `AzureClusterIdentity` assigned. This happens when creating clusters nowadays, as `AzureClusterIdentity` is not required. In this PR we change the namespace where these objects are created, to use the cluster namespace instead of `giantswarm`.

When moving to `capz` Clusters will have `AzureClusterIdentity`. In this PR we make azure-operator to ignore `AzureClusterIdentity` that are not managed by azure-operator.

There are three relevant scenarios, and everything should keep working normally.
- Clusters already created and using a recent azure-operator version. These clusters will have a `AzureClusterIdentity` already created and assigned, in the `giantswarm` namespace. The azure-operator will keep these `AzureClusterIdentity` up to date.
- New Clusters created with the changes in this PR. The azure-operator will create the `AzureClusterIdentity` and keep them up to date on the cluster namespace.
- `capz` clusters where the `AzureCluster` and `AzureClusterIdentity` was created manually in the cluster namespace. In this case, the azure-operator will ignore the `AzureClusterIdentity`.


service/controller/azurecluster/handler/azureclusteridentity/create.go
```diff
@@ -87,6 +87,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			return microerror.Mask(err)
 		}
 
+		// If AzureClusterIdentity was not created by azure-operator, ignore it
+		managedBy := azureClusterIdentity.Labels[apiextensionslabels.ManagedBy]
+		if managedBy != project.Name() {
+			return nil
+		}
```
nprokopic, line 94: Will managed-by label get updated to CAPZ controller during migration to CAPZ?

fiunchinho: I don't think so, because they won't be managed by `capz`. For `capz` clusters, `AzureClusterIdentities` and `Secrets` need to be created when creating the cluster (either the customer creates them or our UIs creates them). I tried explaining the difference scenario on the PR description.

nprokopic: In that case, IIUC, we should at least remove azure-operator from managed-by label, right?

whites11 (approved):
LGTM

service/controller/azurecluster/handler/azureclusteridentity/create.go
fiunchinho: If a customer wants to reuse the `AzureClusterIdentity` and *we remove azure-operator* then yeah, the best thing to do would be to remove the label to avoid confusion.

tuommaki (commented):
Seems fine to me, but is there something in here that would require attention from Team Biscuit? I believe they own the automation that takes care of RBAC rules' management on MC and I wonder if `AzureClusterIdentity` would need to have more restricted rights than e.g. `AzureCluster` ?

tuommaki (approved):

PR #1469, opened 2021-06-15
Update list-releases.sh

T-Kukawka:
Not sure which version of it updated back then, but now should work :D


fiunchinho (changes requested):
make file executable and we are gooooood

fiunchinho (approved):

scripts/list-releases.sh
```diff
@@ -6,15 +6,12 @@ if [ "$1" = "" ]; then
...
-opsctl list installations > list-of-installations.txt
-gsctl list endpoints > list-of-endpoints.txt
 
 provider=$1
 active_release=$2
 for installation in $(opsctl list installations --provider=$provider --short); do
-		gsctl select endpoint "${inallation_data[0]}"
-		
-		if [ "$active_release" = "" ]
+		gsctl select endpoint "${installation[0]}"
```
whites11, line 13: why do you bother doing this? You can run `gsctl list releases -e <endpoint>` without selecting it

PR #1475, opened 2021-06-21
Avoid vmss upgrade stuck

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/17758

This PR adds some checks in the azure machine pool's `node` resource in order to recover from situations where the azure machine pool state machine is in an unknown state.

See https://github.com/giantswarm/giantswarm/issues/17763


tuommaki (approved):
LGTM

PR #1478, opened 2021-06-23
Always reconciliate node pool size

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/17793

currently if one changes the size of a fixed-size (non autoscaling) node pool via azure portal, azure operator ignores the operation.

For example.

NP1: min 3 max 3
I delete one replica via portal (NP1 has 2 replicas now)

At this point I'd expect azure operator to restore the correct size of 3, but it doesn't.

This PR addresses this use case and makes azure operator apply the deployment if the current number of replicas in the NP VMSS and the fixed size of the NP do not match.

### DIFF
--- a/service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
+++ b/service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
@@ -119,7 +119,11 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 	deploymentNeedsToBeSubmitted := currentDeployment.IsHTTPStatus(http.StatusNotFound)
 	nodesNeedToBeRolled := false
 	if !deploymentNeedsToBeSubmitted {
-		changes, err := template.Diff(currentDeployment, desiredDeployment)
+		currentReplicas := int32(-1)
+		if !vmss.IsHTTPStatus(404) {
+			currentReplicas = int32(*vmss.Sku.Capacity)
+		}
+		changes, err := template.Diff(currentDeployment, desiredDeployment, currentReplicas)
 		if err != nil {
 			return currentState, microerror.Mask(err)
 		}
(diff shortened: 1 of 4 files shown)


tuommaki (approved):
I think this change breaks the original design of `template` package in a way that might not be very maintainable in the long term, but I also do understand the problem at hand and given that this code will be eventually replaced by CAPZ, I give my stamp on this.

whites11:
> I think this change breaks the original design of `template` package in a way that might not be very maintainable in the long term, but I also do understand the problem at hand and given that this code will be eventually replaced by CAPZ, I give my stamp on this.

I completely agree Tuommy. The new parameter breaks the whole point, but what else can we do? The problem is that the arm deployment does not update when there are changes in the resources it created.
The alternative would be to compare the size in the state handler directly, but it is tricky to get the value of parameters from there. If you have better ideas I'm more than happy to review it

whites11:
@tuommaki how about now?

service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
```diff
@@ -124,6 +124,17 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
...
 
+		if !contains(changes, "scaling") {
+			desiredParameters, err := template.NewFromDeployment(desiredDeployment)
+			if err != nil {
+				return currentState, microerror.Mask(err)
+			}
+
+			if desiredParameters.Scaling.MinReplicas == desiredParameters.Scaling.MaxReplicas && desiredParameters.Scaling.CurrentReplicas != int32(*vmss.Sku.Capacity) {
+				changes = append(changes, "scaling")
+			}
```
tuommaki, line 135: This is actually very clever way to workaround the problem. Did you test that this works in practice?

tuommaki (approved):
> The problem is that the arm deployment does not update when there are changes in the resources it created.
The alternative would be to compare the size in the state handler directly, but it is tricky to get the value of parameters from there. If you have better ideas I'm more than happy to review it.

My initial idea for "proper" way to implement this is to create a new handler that is executed late in the reconciliation round. It would only ensure that the VMSS capacity matches the desired count and if there are any changes to ARM deployments, ideally this handler would be skipped.

That being said, this would take considerably more time to develop & test so therefore I'm all for your workaround.

service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
whites11: Yes I tested as many use cases I could think of

PR #1489, opened 2021-07-06
Fix udev rules to fix /boot automount 

ericgraf:
`/boot` automount currently fails because the udev rules are missing TAG=systemd on azure storage boot device.

This is fixed upstream with https://github.com/kinvolk/init/pull/41.  Until that change is merged and in the flatcar base image, this change is required to fix the automount issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix udev rules that caused `/boot` automount to fail
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


ericgraf:
Tested working using v15.0.0 as base release on gremlin.

whites11 (approved):
Seems legit.
May I ask what was the outcome of this bug? We never noticed it so far.

ericgraf:
@whites11 since we don't modify boot we recreate the entire instance we likely would not have noticed it. 

The only reason why we noticed it was because we started testing falco. Falco tries to monitor the boot directory for security risks.  
In the past sysdig also noticed this issue https://github.com/giantswarm/giantswarm/issues/15433

tuommaki (approved):
LGTMIIWFY

MarcelMue (approved):

PR #1492, opened 2021-07-13
Allow using an existing public IP for the NAT gateway of worker nodes.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/18048

This PR allows using an externally-managed, existing public IP address for the workers nat gateway.
This is useful for customers that want to have control of the public IP address for egress traffic of worker nodes.


service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
```diff
@@ -130,7 +130,10 @@ func (r *Resource) deploymentUninitializedTransition(ctx context.Context, obj in
 				return currentState, microerror.Mask(err)
 			}
 
-			if desiredParameters.Scaling.MinReplicas == desiredParameters.Scaling.MaxReplicas && desiredParameters.Scaling.CurrentReplicas != int32(*vmss.Sku.Capacity) {
+			if desiredParameters.Scaling.MinReplicas == desiredParameters.Scaling.MaxReplicas &&
```
whites11, line 133: Unrelated with this PR, but it panic'd during tests so I added a new check to avoid the panic

service/controller/key/key.go
```diff
@@ -55,8 +56,6 @@ const (
 
 	AnnotationEtcdDomain        = "giantswarm.io/etcd-domain"
 	AnnotationPrometheusCluster = "giantswarm.io/prometheus-cluster"
-	AnnotationNodePoolMinSize   = "cluster.k8s.io/cluster-api-autoscaler-node-group-min-size"
```
whites11, line 58: leftovers

service/controller/azuremachinepool/handler/nodepool/create_deployment_uninitialized.go
fiunchinho: weird

service/controller/azurecluster/handler/azureconfig/create.go
```diff
@@ -151,6 +152,9 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			}
 		}
 
+		// Ensure External IP address annotation is up to date.
+		presentAzureConfig.Annotations[localannotation.WorkersEgressExternalPublicIP] = mappedAzureConfig.Annotations[localannotation.WorkersEgressExternalPublicIP]
```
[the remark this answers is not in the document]
whites11: I prefer to have the annotation explicit. I don't want to copy over annotations I don't know about because, unlike labels, annotations most often have a direct effect on the cluster

fiunchinho (approved):
LGTM

PR #1528, opened 2021-09-13
Remove vpn gateway delete

whites11:
This PR deletes the code that checks vpn gateway is deleted from resource groups.
The deletion already happened when upgrading to 15.0.0 so this is not needed any more.


.circleci/config.yml
```diff
@@ -27,15 +29,31 @@ workflows:
             tags:
               only: /^v.*/
 
+      # build and push docker image to docker.
```
whites11, line 32: Added because of https://gigantic.slack.com/archives/C04TGHDEF/p1630672178006900

fiunchinho (approved):

PR #1555, opened 2021-11-05
Drain nodes

whites11:
This PR is about solving the node drain problem during an upgrade.
Once a node is deleting, it's not possible to detach disks from the node and that disrupted the upgrade process.

WIth this PR azure operator cordons and drains nodes before deleting them to overcome that problem.


service/controller/azuremachinepool/handler/nodepool/create.go
```diff
@@ -89,11 +101,40 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	// Get a list of all nodes, including worker nodes.
+	nodeList := v1.NodeList{}
+	err = wcClient.List(ctx, &nodeList)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	for _, node := range nodeList.Items {
+		if node.Labels["role"] == "master" || node.Labels["kubernetes.io/role"] == "master" {
+			// Check if node has the right azure operator version label
```
fiunchinho, line 127: so we no longer rely on `annotation.IsMasterUpgrading` being set?

whites11: It was not set anywhere.

fiunchinho (approved):

pkg/drainer/key.go
```diff
@@ -0,0 +1,37 @@
+package drainer
+
+import (
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func isCriticalPod(podName string) bool {
```
AverageMarcus, line 10: I'm unsure about this function. Don't really like having the critical pods hardcoded.

Is it possible to instead infer that they're critical from their priority class or similar?

AverageMarcus (approved):

pkg/drainer/key.go
whites11: This comes straight from node-operator https://github.com/giantswarm/node-operator/blob/36ad1d7b5ff3a5c9fea14246c1b5a1f80083e27c/service/controller/key/key.go#L25

I agree with you we could improve it, but I'm already out of time with this release I'd leave it like it is if you don't mind

AverageMarcus: Yeah, don't think it's a blocker. Just something to be mindful of I think :)

PR #1594, opened 2022-02-02
Exclude capi deps

fiunchinho:
Trying `renovate` feature to exclude dependencies from getting automatic updates. If this works I'll make a PR so this is applied to all giantswarm repositories. 

We want to control CAPI upgrades instead of receiving automatic PR's to upgrade.


fiunchinho:
Going with a ping for the test.

PR #1631, opened 2022-02-22
Roll nodes if cgroups change

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/845

This PR makes nodes in a node pool roll automatically when there is a switch between groups v1 and v2 or vice versa


AverageMarcus (approved):
LGTM

PR #1651, opened 2022-05-05
Set `AzureMachine.Status.Ready` according to AzureMachine's Ready condition

whites11:
Towards:  https://github.com/giantswarm/giantswarm/issues/21366
`kubectl get azuremachine` shows a `Ready` column that displays data from the `status.ready` field.
We never update that field, but rather rely on the `Ready` condition in the status.
this PR sets the `status.ready` field to the same value as the `Ready` condition


njuettner (approved):
lgtm

PR #1655, opened 2022-06-07
unfuck the nancy fuckup

whites11:
fix the nancy madness


fiunchinho (approved):
I'd add a note in the changelog regarding updating dependencies, just in case something goes wrong, we know this happened.

PR #1658, opened 2022-06-07
Localhost node name fix

whites11:
bumped k8scc to solve localhost node name issue


whites11:
just deps bump, going on with a ping

PR #1686, opened 2022-09-01
Ensure giantswarm user is always first

whites11:
Ensure giantswarm user is always created first in the VM to avoid cert-exporter failing to scrape because of lack of permissions.


whites11:
due to lack of reviewers I'll merge straight away

PR #1709, opened 2023-03-07
Bump healthz

whites11:
towards https://github.com/giantswarm/giantswarm/issues/26077 and https://gigantic.slack.com/archives/C02HLSDH3DZ/p1678110177988739


service/controller/cloudconfig/master_template.go
```diff
@@ -50,6 +50,7 @@ func (c CloudConfig) NewMasterTemplate(ctx context.Context, data IgnitionTemplat
 		params.CalicoPolicyOnly = true
 		params.DockerhubToken = c.dockerhubToken
 		params.DisableIngressControllerService = true
+		params.EnableCronJobTimeZone = true
```
whites11, line 53: became an optional feature in latest k8scc so we need to forcefully enable it

T-Kukawka (approved):

PR #1714, opened 2023-04-12
Remove logic to migrate experimental CAPI CRDs

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/2350

This hasn't been tested on a MC yet.

The logic to migrate depends on old CRDs that are long gone. Now that all clusters haven been already migrated to non-experimental CRDs, we can remove the dependency on the old CRDs.
After this is merged, we can upgrade dependencies like `apiextensions` or `capi`.


whites11 (approved):
I guess it's ok to merge without testing. We will test as part of the process for the next release, if any

